Issue #11, opened 2015-09-19
Linux: Javascript Error


battlemidget:
I am starting to look into what is needed to make this work flawlessly on Linux, right now the app works using the dev commands, one thing that is popping up is this error:

```
Uncaught Exception:
TypeError: Cannot read property 'setBadge' of undefined
    at updateBadge (/home/adam/Projects/caprine/index.js:22:10)
    at BrowserWindow.<anonymous> (/home/adam/Projects/caprine/index.js:52:3)
    at emitTwo (events.js:87:13)
    at BrowserWindow.emit (events.js:172:7)
```

I'm going to see if I can fix some of those todo items in the other bug as well.

Heres a screenshot for anyone curious what it looks like on Ubuntu running KDE

![image](https://cloud.githubusercontent.com/assets/51892/9977997/baac9cf8-5eea-11e5-8930-924e35790674.png)

sindresorhus:
Cool :)

Yup, seems I need to put [this](https://github.com/sindresorhus/caprine/blob/5f351dec8a8eb0648076c7dfe4cb209a68d904f5/index.js#L22) into an if statement. On OS X we use the dock badge to show a count of notifications? Is there anything similar for Linux?

battlemidget:
Just from my limited reading for electron I think https://github.com/atom/electron/blob/master/docs/api/tray.md may be what you want for cross platform tray indication. Still looking into it and will patch the code to see if I can make it work

3nit:
Same error (`Cannot read property 'setBadge' of undefined`) happens on Windows too.

PR #13, opened 2015-09-21
Add Linux support

tschuy:
Like you noted in #11, this essentially boils down to putting the ``app.dock`` into an if statement. (I don't think there's any Linux equivalent to the dock icon with a number on it, or the bouncing dock icon.)

I also rearranged the menu to be more cross-platformy, since there's no appname menu on most distributions of Linux. (Though there is in Gnome 3, the desktop manager I use, but I couldn't figure out how to add to it from Electron, and Atom doesn't use it either.)

There is a bug with desktop notifications crashing (https://github.com/atom/electron/issues/2294) that seems pretty awful. The issue isn't entirely clear which versions of what it presents itself on, so I added a note to the README about it. The notifications themselves work just fine, and as long as you don't click on them, the system won't crash.

I would be happy to make any changes you need before merging this. I could also test it on a non-Gnome environment.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ function createMainWindow() {
 		'show': false,
 		'width': 800,
 		'height': 600,
-		'icon': 'media/Icon.png',
+		'icon': path.join(__dirname, 'media', 'Icon.png'),
 		'min-width': 400,
 		'min-height': 200,
 		'title-bar-style': 'hidden-inset',
--- a/index.js
+++ b/index.js
@@ -31,6 +31,7 @@ function createMainWindow() {
 		'show': false,
 		'width': 800,
 		'height': 600,
+		'icon': 'media/Icon.png',
 		'min-width': 400,
 		'min-height': 200,
 		'title-bar-style': 'hidden-inset',
--- a/index.js
+++ b/index.js
@@ -13,6 +13,10 @@ require('crash-reporter').start();
 let mainWindow;
 
 function updateBadge(title) {
+	// don't try to update the dock icon if the system has no dock
+	if (!app.dock) {
+		return;
+	}
 	// ignore `Sindre messaged you` blinking
 	if (title.indexOf('Messenger') === -1) {
 		return;
--- a/index.js
+++ b/index.js
@@ -13,7 +13,6 @@ require('crash-reporter').start();
 let mainWindow;
 
 function updateBadge(title) {
-	// don't try to update the dock icon if the system has no dock
 	if (!app.dock) {
 		return;
 	}
--- a/menu.js
+++ b/menu.js
@@ -12,69 +12,6 @@ function sendAction(action) {
 
 const tpl = [
 	{
-		label: appName,
-		submenu: [
-			{
-				label: `About ${appName}`,
-				role: 'about'
-			},
-			{
-				type: 'separator'
-			},
-			{
-				label: 'Preferences...',
-				accelerator: 'Cmd+,',
-				click() {
-					sendAction('show-preferences');
-				}
-			},
-			{
-				type: 'separator'
-			},
-			{
-				label: 'Log Out',
-				click() {
-					sendAction('log-out');
-				}
-			},
-			{
-				type: 'separator'
-			},
-			{
-				label: 'Services',
-				role: 'services',
-				submenu: []
-			},
-			{
-				type: 'separator'
-			},
-			{
-				label: `Hide ${appName}`,
-				accelerator: 'Cmd+H',
-				role: 'hide'
-			},
-			{
-				label: 'Hide Others',
-				accelerator: 'Cmd+Shift+H',
-				role: 'hideothers'
-			},
-			{
-				label: 'Show All',
-				role: 'unhide'
-			},
-			{
-				type: 'separator'
-			},
-			{
-				label: `Quit ${appName}`,
-				accelerator: 'Cmd+Q',
-				click() {
-					app.quit();
-				}
-			}
-		]
-	},
-	{
 		label: 'File',
 		submenu: [
 			{
@@ -137,13 +74,6 @@ const tpl = [
 				label: 'Close',
 				accelerator: 'CmdOrCtrl+W',
 				role: 'close'
-			},
-			{
-				type: 'separator'
-			},
-			{
-				label: 'Bring All to Front',
-				role: 'front'
 			}
 		]
 	},
@@ -174,5 +104,85 @@ ${process.platform} ${process.arch} ${os.release()}`;
 		]
 	}
 ];
+
+const appMenu = [
+	{
+		label: `About ${appName}`,
+		role: 'about'
+	},
+	{
+		type: 'separator'
+	},
+	{
+		label: 'Preferences...',
+		accelerator: 'CmdOrCtrl+,',
+		click() {
+			sendAction('show-preferences');
+		}
+	},
+	{
+		type: 'separator'
+	},
+	{
+		label: 'Log Out',
+		click() {
+			sendAction('log-out');
+		}
+	}
+];
+
+if (process.platform === 'darwin') {
+	// add primary menu with app name
+	tpl.unshift({
+		label: appName,
+		submenu: appMenu
+	});
+
+	// add OS X-only hide options
+	tpl[0].submenu.push([{
+		type: 'separator'
+	},
+	{
+		label: `Hide ${appName}`,
+		accelerator: 'CmdOrCtrl+H',
+		role: 'hide'
+	},
+	{
+		label: 'Hide Others',
+		accelerator: 'CmdOrCtrl+Shift+H',
+		role: 'hideothers'
+	},
+	{
+		label: 'Show All',
+		role: 'unhide'
+	}]);
+
+	tpl[3].submenu.push(
+		{
+			type: 'separator'
+		},
+		{
+			label: 'Bring All to Front',
+			role: 'front'
+		}
+	);
+} else {
+	// add main menu items to File menu
+	tpl[0].submenu = tpl[0].submenu.concat(appMenu);
+}
+
+// add quit to either app menu or File
+tpl[0].submenu.push(
+	{
+		type: 'separator'
+	},
+	{
+		label: `Quit ${appName}`,
+		accelerator: 'CmdOrCtrl+Q',
+		click() {
+			app.quit();
+		}
+	}
+);
 
 module.exports = Menu.buildFromTemplate(tpl);
--- a/menu.js
+++ b/menu.js
@@ -10,179 +10,279 @@ function sendAction(action) {
 	BrowserWindow.getFocusedWindow().webContents.send(action);
 }
 
-const tpl = [
-	{
-		label: 'File',
-		submenu: [
-			{
-				label: 'New Conversation',
-				accelerator: 'CmdOrCtrl+N',
-				click() {
-					sendAction('new-conversation');
-				}
-			}
-		]
-	},
-	{
-		label: 'Edit',
-		submenu: [
-			{
-				label: 'Undo',
-				accelerator: 'CmdOrCtrl+Z',
-				role: 'undo'
-			},
-			{
-				label: 'Redo',
-				accelerator: 'Shift+CmdOrCtrl+Z',
-				role: 'redo'
-			},
-			{
-				type: 'separator'
-			},
-			{
-				label: 'Cut',
-				accelerator: 'CmdOrCtrl+X',
-				role: 'cut'
-			},
-			{
-				label: 'Copy',
-				accelerator: 'CmdOrCtrl+C',
-				role: 'copy'
-			},
-			{
-				label: 'Paste',
-				accelerator: 'CmdOrCtrl+V',
-				role: 'paste'
-			},
-			{
-				label: 'Select All',
-				accelerator: 'CmdOrCtrl+A',
-				role: 'selectall'
-			}
-		]
-	},
-	{
-		label: 'Window',
-		role: 'window',
-		submenu: [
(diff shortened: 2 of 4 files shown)


menu.js
```diff
@@ -175,4 +105,84 @@ ${process.platform} ${process.arch} ${os.release()}`;
...
+	},
+	{
+		label: `Quit ${appName}`,
+		accelerator: 'CmdOrCtrl+Q',
+		click() {
+			app.quit();
+		}
+	}
+);
+
```
sindresorhus: Since the menues are so different on OS X and Linux I think it would be better to just create a separate file with a Linux menu instead of juggling this mess. The `Help` menu is common and can be a `menu-common.js` file and required into each.

readme.md
```diff
@@ -5,10 +5,11 @@
 <br>
 [![](media/screenshot.png)](https://github.com/sindresorhus/caprine/releases/latest)
 
+*Requires OS X 10.8 or newer and Linux. Windows support planned.*
```
sindresorhus:
```
*Requires OS X 10.8+ or Linux. Windows support planned.*
```

readme.md
```diff
@@ -5,10 +5,11 @@
...
 
-## Install
+## Installing on OS X
 
-*Requires OS X 10.8 or newer. Linux and Windows support planned.*
+*To run on Linux, clone the source and follow the ``dev`` instructions below.*
```
sindresorhus: Why so manual? We can package for Linux using `electron-packager` as described in the Linux issue.

readme.md
```diff
@@ -34,6 +35,11 @@ Desktop notifications can be turned on in Preferences.
 
 <div align="center"><img src="media/screenshot-notification.png" width="358"></div>
 
+NOTE: There is a <a href="https://github.com/atom/electron/issues/2294">known bug</a>
```
sindresorhus: Use Markdown link syntax

readme.md
```diff
@@ -34,6 +35,11 @@ Desktop notifications can be turned on in Preferences.
...
+with Electron's handling of desktop notifications on systems running Gnome 3
+that may cause Caprine to crash if notifications are clicked. Until this bug is
+resolved, do not click on notifications if they cause your the app to crash on
+your system.
```
sindresorhus: Don't hard-wrap

index.js
```diff
@@ -13,6 +13,10 @@ require('crash-reporter').start();
 let mainWindow;
 
 function updateBadge(title) {
+	// don't try to update the dock icon if the system has no dock
```
sindresorhus: I don't think this needs a comment. The code is self-explanatory.

menu.js
tschuy: That's reasonable. It started out not too bad, and then as I added more and more to it, it got crazier and crazier :stuck_out_tongue:

sindresorhus:
Make sure the Linux menu is consistent with other Linux apps.

menu.js
```diff
@@ -175,4 +105,84 @@ ${process.platform} ${process.arch} ${os.release()}`;
...
+const appMenu = [
+	{
+		label: `About ${appName}`,
+		role: 'about'
+	},
+	{
+		type: 'separator'
+	},
+	{
+		label: 'Preferences...',
```
sindresorhus: Re https://github.com/sindresorhus/caprine/pull/13#issuecomment-141874421

On Linux it's only `Preferences`, not `Preferences...`.

tschuy:
Unfortunately, ``About`` goes under help on Linux, so I can't actually factor out the help menu. :cry: 

From what I managed to get by messing around with ``electron-packager`` the best I'll be able to do is make an executable that can then be zipped up, it doesn't have a way to create a single runnable file. I'll make the script throw the build into a zip.

tschuy:
what would you prefer the behavior of ``npm run build`` be? it's constrained by running ``/bin/sh``, not ``/bin/bash``, so I can't do something fancy like ``if [[ "$OSTYPE" == "linux-gnu" ]]; then npm run build-linux; else npm run build-osx; fi``. I could write a separate ``build.sh`` script that compiles your current system, or I could just make it run OS X by default, or both?

sindresorhus:
`npm run build` should build for both platforms. `electron-packager` can handle this fine. See its docs.

tschuy:
When trying to build them together, I run into the following problems:

* ``sign`` doesn't exist on Linux, but ``electron-packager`` still tries to use the ``codesign`` executable, so it crashes
* it doesn't appear to want to put the executable into an archive, even when ``-asar`` is set
* for whatever reason, instead of setting the app icon at compile time, it's set in the source code as the ``icon`` value passed to ``BrowserWindow``. However, ``media/`` isn't included in the build, so there's no static file to reference.

I think two separate build commands might be necessary for that top reason. For the bottom two, I'm sure it's just me being silly and missing something in their documentation. (I still have no idea what ``asar`` is supposed to look like when it's successfully outputted.)

sindresorhus:
> sign doesn't exist on Linux, but electron-packager still tries to use the codesign executable, so it crashes

sindresorhus:
> sign doesn't exist on Linux, but electron-packager still tries to use the codesign executable, so it crashes

Ugh, `electron-packager` is so buggy. Wish we had something better...

I'm ok with just being able to build from OS X. I'm the only one that creates builds anyways. Having one build massively simplifies it, so I think it's worth the limitation.

> it doesn't appear to want to put the executable into an archive, even when -asar is set

Can you elaborate? It's the `zip` command that does this, not `electron-packager`.

> for whatever reason, instead of setting the app icon at compile time, it's set in the source code as the icon value passed to BrowserWindow. However, media/ isn't included in the build, so there's no static file to reference.

The media folder is ignored. I guess we can ignore everything in the media folder, except the icon.

tschuy:
> Can you elaborate?

from ``electron-packager``'s docs, it reads like ``asar`` should put everything into an archive: https://github.com/maxogden/electron-packager#usage

for now, I've just set the ``build-linux`` command to stick it in there itself.

What I could do is have ``build-linux`` and ``build-osx`` commands, and just have ``build`` be ``npm run build-osx && npm run build-linux``. This would still be one command, and it'd allow anyone that wants to run builds themselves (though you're right that people working on the project have no reason to build, only you) to just run the build they want. I don't know if this would break some caching mechanism that ``electron-packager`` has, though.

I tried removing ``media/`` from the build command for Linux and it's not included in the ``dist/Caprine-linux-x64`` folder, so I'm going to look more into it.

tschuy:
ok, I have a solution that more or less works, as far as I can tell. It should allow you to create a Linux zip and an OS X zip by running ``build``, and allow others to create zips for either system by running ``build-osx`` or ``build-linux``. The Linux build has a functioning icon, and I included a ``.desktop`` template in the README for anyone who wants to create a shortcut. Since this isn't using a proper install mechanism (which is to say, a deb or rpm), installation is a little clunky.

menu.js
```diff
@@ -10,169 +10,279 @@ function sendAction(action) {
...
+						const body = `
+		**Please succinctly describe your issue and steps to reproduce it.**
 
--
+		-
 
-${app.getName()} ${app.getVersion()}
-${process.platform} ${process.arch} ${os.release()}`;
+		${app.getName()} ${app.getVersion()}
+		${process.platform} ${process.arch} ${os.release()}`;
```
sindresorhus: Don't indent this. Template literals are whitespace sensitive.

readme.md
```diff
@@ -5,12 +5,11 @@
 <br>
 [![](media/screenshot.png)](https://github.com/sindresorhus/caprine/releases/latest)
 
+*Requires OS X 10.8+ and Linux. Windows support planned.*
 
-## Install
+## Installing
```
sindresorhus: Keep it `Install`

menu.js
```diff
@@ -10,169 +10,279 @@ function sendAction(action) {
...
-			},
-			{
-				label: 'Report Issue',
-				click() {
-					const body = `
-**Please succinctly describe your issue and steps to reproduce it.**
+			]
+		}
+	];
+} else {
```
sindresorhus:
```
} else if (process.platform === 'linux') {
```

menu.js
```diff
@@ -10,169 +10,279 @@ function sendAction(action) {
...
+					}
+				},
+				{
+					label: 'Log Out',
+					click() {
+						sendAction('log-out');
+					}
+				},
+				{
+					label: `Quit ${appName}`,
```
sindresorhus: AFAIK, linux apps don't have this.

menu.js
```diff
@@ -10,169 +10,279 @@ function sendAction(action) {
...
+				{
+					label: 'Paste',
+					accelerator: 'CmdOrCtrl+V',
+					role: 'paste'
+				},
+				{
+					label: 'Select All',
+					accelerator: 'CmdOrCtrl+A',
+					role: 'selectall'
+				},
```
sindresorhus: Linux apps doesn't have all of these either. Look at which one it does have.

menu.js
```diff
@@ -10,169 +10,279 @@ function sendAction(action) {
...
+			role: 'help',
+			submenu: [
+				{
+					label: 'Website',
+					click() {
+						shell.openExternal('https://github.com/sindresorhus/caprine');
+					}
+				},
+				{
+					label: 'Report Issue',
```
sindresorhus: These two submenu items shouldn't be duplicated. Put them in an external array and concat them into each of the menues.

package.json
```diff
@@ -17,7 +17,9 @@
   "scripts": {
     "test": "xo",
     "start": "electron .",
-    "build": "electron-packager . $npm_package_productName --overwrite --out=dist --ignore='^/dist$' --ignore='^/media$' --prune --asar --platform=darwin --arch=x64 --icon=media/Icon.icns --app-bundle-id=com.sindresorhus.caprine --sign='Developer ID Application: Sindre Sorhus (YG56YK5RN5)' --app-version=$npm_package_version --version=$npm_package_electronVersion && cd dist/Caprine-darwin-x64 && zip -ryXq9 ../Caprine-${npm_package_version}.zip Caprine.app"
+    "build": "npm run build-osx && npm run build-linux",
+    "build-osx": "electron-packager . $npm_package_productName --overwrite --out=dist --ignore='^/dist$' --ignore='^/media$' --prune --platform=darwin,linux --arch=x64 --icon=media/Icon.icns --app-bundle-id=com.sindresorhus.caprine --sign='Developer ID Application: Sindre Sorhus (YG56YK5RN5)' --app-version=$npm_package_version --version=$npm_package_electronVersion && cd dist/Caprine-darwin-x64 && zip -ryXq9 ../Caprine-${npm_package_version}.zip Caprine.app",
```
sindresorhus: You need to remove `linux` here.

menu.js
```diff
@@ -10,169 +10,279 @@ function sendAction(action) {
...
-				type: 'separator'
-			},
-			{
-				label: 'Preferences...',
-				accelerator: 'Cmd+,',
-				click() {
-					sendAction('show-preferences');
+let tpl;
+
+if (process.platform === 'darwin') {
```
sindresorhus: I'd rather have the two menues in two different variables and do the conditional at the bottom with just those variables.

menu.js
```diff
@@ -10,169 +10,279 @@ function sendAction(action) {
...
-${process.platform} ${process.arch} ${os.release()}`;
+		${app.getName()} ${app.getVersion()}
+		${process.platform} ${process.arch} ${os.release()}`;
 
-					shell.openExternal(`https://github.com/sindresorhus/caprine/issues/new?body=${encodeURIComponent(body)}`);
+						shell.openExternal(`https://github.com/sindresorhus/caprine/issues/new?body=${encodeURIComponent(body)}`);
+					}
+				},
+				{
+					label: `About ${appName}`,
```
sindresorhus: On Ubuntu it's just `About`.

Look at the existing Linux menues to make sure everything is consistent.

sindresorhus:
Is there any way to have the app installable on Linux with something like `Homebrew Cask` on OS X?

If https://github.com/loopline-systems/electron-builder supported Linux we could use that. Maybe worth opening a feature request for that?

battlemidget:
We could look at how atom builds their deb's/rpm's and mimic that. Would at least support the major distributions

tschuy:
I've addressed your feedback about the menus and fixed the accidental inclusion of linux in the build-osx command. 

For menu consistency, I'm not sure there's a really good standard compared to what there is on OS X. I tried to go off of Gedit. Most Gnome apps use the menu at the top of the screen (that I've not seen any other app use, and don't think there's support for in electron), but gedit has a more traditional File / Edit / Help etc.

You mentioned that the edit menu shouldn't have some of those items, but gedit has all of them. I've left them there for now; which ones were you thinking of?

About packaging -- I don't think there's a good way to distribute this unless we make it build an RPM and a DEB. I'll make a feature request on https://github.com/loopline-systems/electron-builder, and revisit this later if they add support.

tschuy:
Also, I'd be more than happy to squash these commits if you want so that ``address feedback`` isn't in the git log.

tschuy:
issue for adding linux support to the builder: https://github.com/loopline-systems/electron-builder/issues/7

battlemidget:
> You mentioned that the edit menu shouldn't have some of those items, but gedit has all of them. I've left them there for now; which ones were you thinking of?

I think it just depends on the application, things like gedit, kate, mousepad, etc make use of those edit menu items where as others don't. They probably are useful for any type of text editing but I'm not sure this app needs menu items for it. Take for example Telegram on the desktop for Linux, it doesn't contain any of those menu items (File, Edit) etc.

sindresorhus:
@tschuy Landed! Really appriciate you working on this @tschuy :)

Issue #6, opened 2015-09-17
cmd+w functionality


jedireza:
Could we make the cmd+w functionality on the Mac not close the app but just close/hide the window?

sindresorhus:
I know that's the OS X way, but doesn't make a lot of sense when the app is a single window app. Why do you need this?

jedireza:
I've been closing the app instead of just hiding the window. I guess it's muscle memory. The Gitter and Slack apps, although single windows also work this way.

sindresorhus:
If you intend to hide it, you should just hide it and not close it then: <kbd>Cmd</kbd> <kbd>H</kbd>

Caprine works like any other single-window apps on OS X: Notes, Reminder, Photos, Contacts, App Store, Photo Booth, FaceTime, Calculator, Dictionary, etc. All of those close the app if you close the window.

timche:
+1

In my opinion, Caprine should stay active. It's a chatting application and needs the option to stay active in the background. I'm closing my app windows with `cmd + w` in any other app like Slack or Goofy/ChitChat.

sindresorhus:
@timche Other apps doing it wrong is not a good argument. Cmd+H is the standard way of hiding a window. It's kinda amazing that you would rather maintain a fork than learn how to do it correctly...

timche:
Can you show me where this rule is written down that this is a wrong approach?

It's just how the majority of apps are handling cmd + w and they are not closing the app at all ... I don't need apps like Notes, Contacts, App Store running frequently in the background so it's okay they close at cmd + w, but when I'm using a chat app, I expect it's not quitting the whole app with cmd + w. I'm not the only one with this expectation.

sindresorhus:
> Can you show me where this rule is written down that this is a wrong approach?

What the built-in apps do is the rule in my view. If you can find anything written backing your argument I would be happy to reconsider.

But again, why not just always use Cmd+H and get consistent behaviour instead of Cmd+W which behaves sporadically.

timche:
> If you can find anything written backing your argument I would be happy to reconsider.

I can't, it's just my personal experience with other apps I'm using daily.

> But again, why not just always use Cmd+H and get consistent behaviour instead of Cmd+W which behaves sporadically.

Because I'm already get used to cmd + w, which behaves actually consistent except Caprine.

[...]

PR #32, opened 2015-11-19
Persist window state and hide window on Cmd+W

phamdaniel:
Remembers the windows position and size for when reopening the app. 
On quit, saves the current window state in JSON.
Restores from file on open.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-native-reassign no-undef */
 'use strict';
 const ipc = require('electron').ipcRenderer;
 
@@ -25,7 +26,7 @@ const NativeNotification = Notification;
 Notification = function (title, options) {
 	const notification = new NativeNotification(title, options);
 	notification.addEventListener('click', () => {
-		ipc.send('notificationClicked');
+		ipc.send('notification-click');
 	});
 
 	return notification;
--- a/browser.js
+++ b/browser.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-native-reassign no-undef */
 'use strict';
 const ipc = require('electron').ipcRenderer;
 
@@ -20,6 +19,7 @@ ipc.on('log-out', () => {
 	document.querySelector('._54nq._2i-c._150g._558b._2n_z li:last-child a').click();
 });
 
+/* eslint-disable no-native-reassign, no-undef */
 // Extend and replace the native notifications.
 const NativeNotification = Notification;
 
@@ -34,3 +34,4 @@ Notification = function (title, options) {
 Notification.prototype = NativeNotification.prototype;
 Notification.permission = NativeNotification.permission;
 Notification.requestPermission = NativeNotification.requestPermission.bind(Notification);
+/* eslint-enable no-native-reassign, no-undef */
--- a/browser.js
+++ b/browser.js
@@ -18,3 +18,18 @@ ipc.on('log-out', () => {
 
 	document.querySelector('._54nq._2i-c._150g._558b._2n_z li:last-child a').click();
 });
+
+// Extend and replace the native notifications.
+const NativeNotification = Notification;
+
+Notification = function (title, options) {
+	const notification = new NativeNotification(title, options);
+	notification.addEventListener('click', () => {
+		ipc.send('notificationClicked');
+	});
+
+	return notification;
+};
+Notification.prototype = NativeNotification.prototype;
+Notification.permission = NativeNotification.permission;
+Notification.requestPermission = NativeNotification.requestPermission.bind(Notification);
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ require('electron-dl')();
 electron.crashReporter.start();
 
 let mainWindow;
+let isQuitting = false;
 
 function updateBadge(title) {
 	if (!app.dock) {
@@ -57,15 +58,20 @@ function createMainWindow() {
 	});
 
 	win.loadURL('https://www.messenger.com/login/');
-	win.on('close', () => {
-		if (!win.isFullScreen()) {
-			const bounds = win.getBounds();
-			storage.set('lastWindowState', {
-				x: bounds.x,
-				y: bounds.y,
-				width: bounds.width,
-				height: bounds.height
-			});
+	win.on('close', e => {
+		if (isQuitting) {
+			if (!win.isFullScreen()) {
+				const bounds = win.getBounds();
+				storage.set('lastWindowState', {
+					x: bounds.x,
+					y: bounds.y,
+					width: bounds.width,
+					height: bounds.height
+				});
+			}
+		} else {
+			e.preventDefault();
+			win.hide();
 		}
 	});
 	win.on('closed', app.quit);
@@ -90,4 +96,12 @@ app.on('ready', () => {
 		e.preventDefault();
 		electron.shell.openExternal(url);
 	});
+});
+
+app.on('activate', () => {
+	mainWindow.show();
+});
+
+app.on('before-quit', () => {
+	isQuitting = true;
 });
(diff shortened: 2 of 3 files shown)


phamdaniel:
Also added in the code to address #6
I added it here as having it in another pull request would have caused merge conflicts.
It can be separated if need be.

sindresorhus:
Nice! Generally looks good. Noticed one thing though. When I close the window (and it's hidden), and I get a new message notification, click the notification, the app is focused, but the window is not shown. If the app window is hidden with Cmd+H instead of closed, it works fine.

storage.js
```diff
@@ -0,0 +1,29 @@
...
+	fs.writeFileSync(dataFilePath, JSON.stringify(data));
+};
+
+exports.get = function (key) {
+	const data = readData();
+	let value = null;
+	if (key in data) {
+		value = data[key];
+	}
+	return value;
```
sindresorhus:
```js
return readData()[key];
```

index.js
```diff
@@ -27,11 +29,20 @@ function updateBadge(title) {
 }
 
 function createMainWindow() {
+	let lastWindowState = storage.get('lastWindowState');
+	if (!lastWindowState) {
+		lastWindowState = {
+			width: 800,
+			height: 600
+		};
+	}
```
sindresorhus:
```js
const lastWindowState = storage.get('lastWindowState') || {width: 800, height: 600};
```

index.js
```diff
@@ -47,6 +58,22 @@ function createMainWindow() {
...
+					width: bounds.width,
+					height: bounds.height
+				});
+			}
+		} else {
+			e.preventDefault();
+			win.hide();
+		}
+	});
 	win.on('closed', app.quit);
```
sindresorhus: This should be removed now that closing the window no longer quits the app.

index.js
```diff
@@ -47,6 +58,22 @@ function createMainWindow() {
 	});
 
 	win.loadURL('https://www.messenger.com/login/');
+	win.on('close', e => {
+		if (isQuitting) {
+			if (!win.isFullScreen()) {
+				const bounds = win.getBounds();
+				storage.set('lastWindowState', {
```
sindresorhus: Why not just save `bounds` directly?

index.js
```diff
@@ -47,6 +58,22 @@ function createMainWindow() {
 	});
 
 	win.loadURL('https://www.messenger.com/login/');
+	win.on('close', e => {
+		if (isQuitting) {
+			if (!win.isFullScreen()) {
```
sindresorhus: I think it would be cleaner to do this in `before-quit`.

storage.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+const app = require('app');
+const fs = require('fs');
+const path = require('path');
+const dataFilePath = path.join(app.getPath('userData'), 'data.json');
+
+function readData() {
+	if (!fs.existsSync(dataFilePath)) {
```
sindresorhus: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback

> fs.exists() should not be used to check if a file exists before calling fs.open(). Doing so introduces a race condition since other processes may change the file's state between the two calls. Instead, user code should call fs.open() directly and handle the error raised if the file is non-existent.

storage.js
```diff
@@ -0,0 +1,29 @@
...
+		return {};
+	}
+
+	return JSON.parse(fs.readFileSync(dataFilePath, 'utf-8'));
```
sindresorhus: `utf-8` => `utf8`

phamdaniel:
Thanks for the feedback, hopefully I covered everything.

The reason why Cmd+H inherently worked was that "hidden" windows would show as soon as they have focus (from Cmd+Tab and the likes).
So now, in order to show "closed" windows we have to tell the app that the notification was clicked.
Like Electron notifications don't emit events, I extended the native code to trigger one on click. But now it causes the build to fail...

browser.js
```diff
@@ -18,3 +18,18 @@ ipc.on('log-out', () => {
 
 	document.querySelector('._54nq._2i-c._150g._558b._2n_z li:last-child a').click();
 });
+
```
sindresorhus:
```js
/* eslint-disable no-native-reassign no-undef */
```

at the top of this file should fix the build failure ;)

browser.js
```diff
@@ -18,3 +18,18 @@ ipc.on('log-out', () => {
...
+const NativeNotification = Notification;
+
+Notification = function (title, options) {
+	const notification = new NativeNotification(title, options);
+	notification.addEventListener('click', () => {
+		ipc.send('notificationClicked');
+	});
+
+	return notification;
+};
```
sindresorhus: Why not just use https://developer.mozilla.org/en-US/docs/Web/API/Notification/onclick instead of decorating `Notification`?

browser.js
```diff
@@ -18,3 +18,18 @@ ipc.on('log-out', () => {
...
+Notification.prototype = NativeNotification.prototype;
+Notification.permission = NativeNotification.permission;
+Notification.requestPermission = NativeNotification.requestPermission.bind(Notification);
```
sindresorhus: I don't think you need to manually set `permission` and `requestPermission` here. Inheriting the prototype should give you those methods implicitly.

browser.js
```diff
@@ -18,3 +18,18 @@ ipc.on('log-out', () => {
...
 	document.querySelector('._54nq._2i-c._150g._558b._2n_z li:last-child a').click();
 });
+
+// Extend and replace the native notifications.
+const NativeNotification = Notification;
+
+Notification = function (title, options) {
+	const notification = new NativeNotification(title, options);
+	notification.addEventListener('click', () => {
+		ipc.send('notificationClicked');
```
sindresorhus:
```js
ipc.send('notification-click');
```

browser.js
phamdaniel: I wasn't able to use it as it wouldn't trigger on click.

sindresorhus: @phamdaniel Ok, would you mind opening an Electron issue about that?

https://github.com/atom/electron/issues

browser.js
phamdaniel: For some reason if you don't set it manually it won't allow the decorated notification to show up.

browser.js
phamdaniel: sure thing.

browser.js
sindresorhus: Alright

index.js
```diff
@@ -2,8 +2,10 @@
 const path = require('path');
 const fs = require('fs');
 const electron = require('electron');
+const ipc = require('electron').ipcMain;
 const app = electron.app;
 const appMenu = require('./menu');
+const storage = require('./storage');
```
etiktin: You might want to consider using [configstore](https://www.npmjs.com/package/configstore) for this stuff.

sindresorhus: @etiktin I made `configstore`, but I think the current solution is better in this case as `electron` has a more comprehensive access to data directory paths on various OSes.

phamdaniel:
Figured it out and it builds now.

sindresorhus:
Thank you! :) Very nice work on this @phamdaniel.

Zirro:
@sindresorhus I came across this pull request while working on another Electron app, and I am wondering whether it would make sense to move the persisting window state-feature into its own module. I certainly think that there are more Electron apps which would benefit from it. What do you think?

etiktin:
@Zirro check out [electron-window-state](https://www.npmjs.com/package/electron-window-state).

Zirro:
@etiktin Ah, thank you.

browser.js
sindresorhus: @phamdaniel Where you able to make the `.onclick` thing work? According to the referenced issue you were, but I just tried using the following and couldn't make it work:

```js
Notification.onlick = () => {
	ipc.send('notification-click');
};
```

phamdaniel: No it didn't work and upon further digging it wasn't due to an error in Electron.
One of the main reason why that code doesn't work is because, in Facebook's code, the onclick function is being redefined when a new notification is created.

PR #36, opened 2015-12-05
Add tray for non-darwin platforms

dogancelik:
Because Windows or other platforms [doesn't have a Dock feature](https://github.com/atom/electron/blob/master/docs/api/app.md), you can't show or quit Caprine after closing main window. It stays hidden in background forever. So I added a simple Tray for non Darwin platforms.

(Used EditorConfig)

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ const ipc = require('electron').ipcMain;
 const app = electron.app;
 const appMenu = require('./menu');
 const storage = require('./storage');
+const createTray = require('./tray');
 
 require('electron-debug')();
 require('electron-dl')();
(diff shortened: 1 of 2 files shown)


dogancelik:
I tried to fix xo test but it gives errors for call/apply because Reflect is undefined

dogancelik:
I removed anonymous function and put Darwin check in the main function, this should fix the errors.

dogancelik:
I squashed the commits, please review @sindresorhus

sindresorhus:
Generally looks good, but I'm busy this week, so will test on Windows and Linux next week and do a release then.

sindresorhus:
Thank you! :)

https://github.com/sindresorhus/caprine/releases/tag/0.5.0

Issue #37, opened 2015-12-17
Crashes on startup on Linux


rwhogg:
As of version 0.5.0, Caprine crashes on Linux after starting up. It looks like the issue is related to the new tray icon.

When you start up, it displays a pop-up window displaying the following message:
Uncaught Exception:
TypeError: Error processing argument -1.
    at TypeError (native)
    at module.exports.win (/home/bob/software/caprine/resources/app/tray.js:11:15)
    at EventEmitter.<anonymous> (/home/bob/software/caprine/resources/app/index.js:76:2)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7)

As I was writing this issue, I discovered the solution. Will post a patch in a few minutes.

sindresorhus:
Ugh, the only time I don't test on Linux, because my Ubuntu VM is borked, this happens. Thanks for reporting :)

PR #38, opened 2015-12-17
Fix icon path on Linux

rwhogg:
This fixes a crash (#37) that makes Caprine version 0.5.0 unusable on Linux.

@dogancelik, if you could test this to make sure it still works on Windows, that would be greatly appreciated.

### DIFF
--- a/tray.js
+++ b/tray.js
@@ -1,7 +1,8 @@
 const path = require('path');
 const electron = require('electron');
 const app = electron.app;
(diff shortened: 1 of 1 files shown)


dogancelik:
Yes, it works well, thanks for the mention :smiley:

sindresorhus:
Thank you @rwhogg :)

https://github.com/sindresorhus/caprine/releases/tag/v0.5.1

PR #39, opened 2016-01-06
Add quit option to app menu and open app from tray upon left click

danielhusar:
Since app by default go to tray I think its nice to have quit option in file dropdown.
Also left click on tray does nothing and most users probably expect app to open.

### DIFF
--- a/menu.js
+++ b/menu.js
@@ -189,6 +189,15 @@ const linuxTpl = [
 				click() {
 					sendAction('log-out');
 				}
+			},
+			{
+				type: 'separator'
+			},
+			{
+				label: 'Quit',
+				click() {
+					app.quit();
+				}
 			}
 		]
 	},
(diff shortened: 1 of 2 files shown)


menu.js
```diff
@@ -189,6 +189,15 @@ const linuxTpl = [
 				click() {
 					sendAction('log-out');
 				}
+			},
+			{
+				type: 'separator'
+			},
+			{
+				label: `Quit ${appName}`,
```
sindresorhus: Should just be `Quit`.

kevva:
:+1:

menu.js
danielhusar: Sure but you have it same in OSX :) (altrough its not under File dropdown)
https://github.com/danielhusar/caprine/blob/improvements/menu.js#L75

sindresorhus: The convention differ between OSes. On Linux (Ubuntu) it's mostly `Quit`.

kevva: It's `Quit ${appName}` on OS X. But maybe the naming convention differs?

danielhusar: :+1:

kevva:
Nice work, thanks!

PR #42, opened 2016-01-07
Add Linux installer script

### LINKED ISSUE #15: Linux installer
Looks like we can use [`fpm`](https://github.com/jordansissel/fpm) to easily create a `deb` package.

Example: https://github.com/Aluxian/Facebook-Messenger-Desktop/blob/29b493f443ce2dc385ed62437a5c9013e4e45408/gulpfile.coffee#L107
## 

@tschuy Interested?

kevva:
Creates a working `.deb` file as far as I can see. Do we need some `post-install` or `post-remove` scripts or something (probably for a shortcut)?

Fixes #15.

cc @tschuy @forivall


forivall:
you can setup an install script like https://github.com/sudo-systems/re-kodi/blob/master/install_scripts/linux_install_ia32.sh to install the .desktop files, however, i don't really like the approach they use, since that doesn't register the location in the package management system. It looks like https://github.com/Automattic/wp-desktop/blob/0eead4c6d50507f633f00d35e77c4b7004ef35cf/resource/build-scripts/package-linux.js does this properly.

Instead, the package should have a file at `/usr/share/applications/caprine.desktop` (and include the icon in `/usr/share/icons/hicolor` or `/usr/share/pixmaps`)

Instead, the post-install & post-remove scripts should just have `update-desktop-database`. (not sure if that's just an arch thing)

The main problem that it has right now, is that this currently installs the files into the root directory, `/`, creating a '/locales', '/resources', etc. Either you need to set up the directories appropriately and install caprine under `/opt` or `/usr/share`.

Sorry for the rush comment, i'm in a rush.

sindresorhus:
@kevva :arrow_up:

sindresorhus:
Closing for lack of activity.

PR #44, opened 2016-01-30
Add shortcuts for finding and selecting conversations

phamdaniel:
Focus on search bar: Cmd or Ctrl + F

Specific to OSX:
* Next convo: Cmd + Shift + ]
* previous convo: Cmd + Shift + [

On Windows and Linux:
* Next : Ctrl + Tab
* Previous: Ctrl + Shift + Tab

Didn't figure out how to trigger the loading of more conversations when at the bottom of the list.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,6 +1,6 @@
 'use strict';
 const ipc = require('electron').ipcRenderer;
-const listSelector = 'ul[aria-label="Conversation List"] > li';
+const listSelector = 'div[aria-label="Conversations"] > ul > li';
 
 ipc.on('show-preferences', () => {
 	// create the menu for the below
@@ -22,7 +22,7 @@ ipc.on('log-out', () => {
 
 ipc.on('find', () => {
 	document.querySelector('._58al').focus();
-})
+});
 
 ipc.on('next-conversation', () => {
 	const index = getSelectedIndex() + 1;
@@ -40,7 +40,7 @@ function getSelectedIndex() {
 	const selected = document.querySelector('._5l-3._1ht1._1ht2');
 	const list = selected.parentNode;
 
-	return Array.prototype.indexOf.call(list.children, selected);
+	return Array.from(list.children).indexOf(selected);
 }
 
 /* eslint-disable no-native-reassign, no-undef */
--- a/browser.js
+++ b/browser.js
@@ -1,5 +1,6 @@
 'use strict';
 const ipc = require('electron').ipcRenderer;
+const listSelector = 'ul[aria-label="Conversation List"] > li';
 
 ipc.on('show-preferences', () => {
 	// create the menu for the below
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Next convo: Cmd + Shift + ]
previous convo: Cmd + Shift + [

Why these on OS X?

I tried the Messages.app and it shows Ctrl+Tab in the menu, but both Ctrl+Tab and Cmd+} works.

browser.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 const ipc = require('electron').ipcRenderer;
+const listSelector = 'ul[aria-label="Conversation List"] > li';
```
sindresorhus: This won't work with non-English languages.

browser.js
```diff
@@ -19,6 +20,29 @@ ipc.on('log-out', () => {
...
+	if (index >= 0) {
+		document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
+	}
+});
+
+function getSelectedIndex() {
+	const selected = document.querySelector('._5l-3._1ht1._1ht2');
+	const list = selected.parentNode;
+
+	return Array.prototype.indexOf.call(list.children, selected);
```
sindresorhus: Just use `Array.from()` to convert a node list to an array.

sindresorhus: You could also use `[].includes()`

menu.js
```diff
@@ -148,6 +148,40 @@ const darwinTpl = [
 				type: 'separator'
 			},
 			{
+				label: 'Select Next Conversation',
+				accelerator: (() => {
+					if (process.platform === 'darwin') {
+						return 'Cmd+Shift+}';
+					}
+					return 'Ctrl+Tab';
```
sindresorhus: This could just be a ternary.

phamdaniel:
Hmmmm
I put those as I thought those were the OSX defaults... (they're the only one that work in Terminal)
We can switch to Ctrl + Tab if you think that's better

phamdaniel:
Actually it's more universal to have it be Ctrl Tab as the [ and ] can be in some really weird positions on non English keyboards.

browser.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 const ipc = require('electron').ipcRenderer;
+const listSelector = 'div[aria-label="Conversations"] > ul > li';
```
sindresorhus: Still won't work on non-English Messenger.

sindresorhus:
Would be useful if it would go back to 1, if you press Ctrl+Tab on the last item.

phamdaniel:
This time it should do it.
Didn't notice that the translations were identical for the test language last time...

sindresorhus:
Excellent. Thanks for the PR and quick turnaround :)

PR #47, opened 2016-01-31
Add Dark Mode

oskarkrawczyk:
Since I'm a huge fan of night modes figured I'll add one to Caprine.

You can toggle it from the Electron/Caprine menu item:

screen shot 2016-01-31 at 08 55 52

Quick preview of how it looks:

screen shot 2016-01-31 at 08 55 46

Obviously the sheer amount of random-looking class names is a bit scary, hope FB doesn't change them too often.

### DIFF
--- a/browser.css
+++ b/browser.css
@@ -170,7 +170,8 @@ html.nightMode ._1n-e {
 
 /* messages list: text header above the messages */
 html.nightMode ._17w2,
-html.nightMode ._ih3 ._3oh- {
+html.nightMode ._ih3 ._3oh-,
+html.nightMode ._3oh- {
 	color: rgba(255, 255, 255, 0.7);
 }
 
--- a/browser.css
+++ b/browser.css
@@ -76,3 +76,157 @@ html {
 		display: none !important;
 	}
 }
+
+a, button, *[role="button"] {
+	outline: none !important;
+}
+
+html.nightMode body {
+	color: rgba(255, 255, 255, 0.7);
+}
+
+/* main container? */
+html.nightMode ._4sp8 {
+	background: #192633 !important;
+}
+
+/* main content */
+html.nightMode ._1q5- {
+	background: #192633 !important;
+	border-left: 1px solid rgba(255, 255, 255, 0.05) !important;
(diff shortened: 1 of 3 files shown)


sindresorhus:
This is really awesome! :)

Would you be up for committing to maintaining the night mode? E.g. if the selectors needs updating. I don't think I will have time for that. Would be happy to add you as a maintainer if so.

What do you think about adding a keyboard shortcut for activating night mode?

I've opened a [request](https://github.com/atom/electron/issues/4300) to expose the OS X Dark Mode info in Electron. That way we could activate night mode automatically when OS X is in Dark Mode.

oskarkrawczyk:
Sure, I'll be sticking to using Caprine until FB releases their native IM (which is to believed they're working on). So I have no problem maintaining the Night Mode.

An additional shortcut sounds cool. My suggestion would be `CmdOrCtrl+d` :-)

Also, I'm thinking on renaming Night Mode to Dark Mode, just to have consistency with OS X, what do you think?

sindresorhus:
:+1: to both questions.

oskarkrawczyk:
@sindresorhus Added the keyboard shortcut, fixed a few outstanding color issues, and I think we're ready to push this out the doors!

sindresorhus:
> Also, I'm thinking on renaming Night Mode to Dark Mode, just to have consistency with OS X, what do you think?

Could you do this too?

oskarkrawczyk:
Ah, damn, completely forgot about this.

oskarkrawczyk:
@sindresorhus Done

sindresorhus:
Awesome! The dark mode is really nice. I'll be using it as default :)

https://github.com/sindresorhus/caprine/releases/tag/v1.0.0

amilajack:
The cask version should also be updated:

https://github.com/caskroom/homebrew-cask/blob/master/Casks/caprine.rb

sindresorhus:
@vinkla Would you be able to update it?

vinkla:
@sindresorhus sure, I'll update it.

Issue #53, opened 2016-02-17
Change conversation shortcuts on OS X


oskarkrawczyk:
This is funny but I opened up Atom, and wanted to implement conversation switching just to find out it's already there, only under key binding that I would never expected.

I'm kind of used to the native tab switching keys on OS X: `CMD + Shift + [` and `CMD + Shift + ]`.

I'm sure there's a reason why Caprine is currently using different ones, but perhaps it would make a good idea to change them for a more native experience?

sindresorhus:
I think Ctrl+Tab is more common.? And it's easier to type.

Both shortcuts work in Chrome, Messages.app, Safari, etc, though.

oskarkrawczyk:
Interesting, I've never heard of this shortcut, actually asked a bunch of friends, they also use `CMD+Shift+[]` – which also work everywhere.

And since that's the case, perhaps we could have both as well?

What are your thoughts?

sindresorhus:
> I've never heard of this shortcut, actually asked a bunch of friends, they also use CMD+Shift+[] – which also work everywhere.

I had never heard of `CMD+Shift+[]` until just now either.

> And since that's the case, perhaps we could have both as well?

Sure.

oskarkrawczyk:
It shall be done

PR #54, opened 2016-02-17
Additional conversation switchers on OS X

oskarkrawczyk:
As it turns out it's not all that easy to add multiple shortcuts, actually it doesn't seem to be possible at all to add multiple accelerators.

Which is why the only fallback would be the oldschool way, not perfect but it works.

Issue #53


sindresorhus:
Looks good. Thanks :)

PR #57, opened 2016-02-25
Fix some styling in dark mode

lukyth:
- Make a timestamp of unread message has different color from a read one
- Add hover effect when choosing people for new conversation
- Fix people list stacking border
- Fix for new class name (Seems like Facebook has changed some of them)


sindresorhus:
🎉:)

Issue #58, opened 2016-02-26
Tray icon on KDE


kidager:
The tray icon on Linux (KDE) is a blank white rectangle with no title, also clicking on it does nothing, I have to right click and choose either show or hide or exit.

sindresorhus:
I don't use Linux and I only have a Ubuntu to test with.

Maybe @rwhogg or @dogancelik knows.

rwhogg:
I apologize, but I mostly use GNOME and don't have KDE available to test with at the moment.

sindresorhus:
@kidager Can you share a screenshot of how it looks?

sindresorhus:
We're really just using the Electron API here, and it works fine on Ubuntu, so I'd say it's a bug in Electron. Would you mind opening an issue about it here https://github.com/atom/electron/issues/new with a screenshot, KDE version and your above description of the problem and reference this ticket.

dogancelik:
I booted up a virtual fresh Kubuntu installation and opened Caprine, I don't get any tray icons, no errors either. Might be an Electron issue.

kidager:
The white square is for Messenger and the icon next to it is for [WhatsDesktop](https://github.com/mawie81/whatsdesktop) (which is a similar app based on Electron also)

So this does not seem to be related to electron.

![screenshot](https://cloud.githubusercontent.com/assets/3314980/13373751/6723bf4e-dd70-11e5-811e-b1a059e8539d.png)

I'm using Debian 8.3 x64 with KDE Plasma 4.14.2

@dogancelik I guess Kubuntu is using Plasma 5 that why we have had different results

kidager:
It seems to be an icon size problem, I have reduced the size of the icon and I got the following results:
![screenshot004](https://cloud.githubusercontent.com/assets/3314980/13373948/48708248-dd76-11e5-969b-2d2319960109.png)
![screenshot005](https://cloud.githubusercontent.com/assets/3314980/13373949/4d7b1e24-dd76-11e5-8965-da834c1b387e.png)

patefoniq:
I also have problem with icon but a little different. Everything shown on screenshot.

![Screenshot1_20230305_111757](https://user-images.githubusercontent.com/11136688/222954552-43a8ce0c-8cc1-4449-a390-004a416df5ac.png)

![Screenshot1_20230305_111624](https://user-images.githubusercontent.com/11136688/222954551-7f71f0fd-0cf8-467d-8552-6731574b043b.png)

PR #59, opened 2016-02-27
Changed icons size for Linux KDE (Plasma 4) #58

kidager:
This should solve the icon problem for KDE Plasma 4.

As for the icon clicks and icons not appearing in Plasma 5, it seems to be a problem with electron since it affects many applications.


sindresorhus:
Thanks :)

kidager:
Anytime :smile:

PR #62, opened 2016-03-09
Only allow a single instance of the app

### LINKED ISSUE #45: Multiple instances on Linux
When app is hidden into tray area and I click icon on taskbar - new instance is starting instead of showing existing one. Also there is no option to disable tray icon (which would be workaround).

phamdaniel:
This will probably fix #45 but I don't have Linux to test.
Tested on OS X and it prevents multi instancing via the command line.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,6 +13,19 @@ require('electron-dl')();
 let mainWindow;
 let isQuitting = false;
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Looks good. Will merge when I can get on a faster internet and test it out. Probably not until monday.

PR #65, opened 2016-03-11
Allow app to automatically enter Dark-mode

### LINKED ISSUE #48: Enable Caprine dark mode when OS X dark mode is enabled
Depends on https://github.com/atom/electron/issues/4300.

phamdaniel:
New version of Electron was released :tada: 

Made it so that the auto switching is optional.
I'm not great at naming things... so I could use a bit of help there.

Closes #48

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -32,11 +32,9 @@ ipc.on('previous-conversation', previousConversation);
 
 ipc.on('dark-mode', toggleDarkMode);
 
-ipc.on('link-theme', () => {
-	if (process.platform === 'darwin' && storage.get('linkedTheme')) {
-		storage.set('darkMode', app.isDarkMode());
-		setDarkMode();
-	}
+app.on('platform-theme-changed', () => {
+	storage.set('darkMode', app.isDarkMode());
+	setDarkMode();
 });
 
 function nextConversation() {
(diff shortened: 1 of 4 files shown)


sindresorhus:
Don't need an option for this. It should just happen.

z-ax:
@sindresorhus, I think Dark layout for FB messenger should be optional (and maybe automatically switch only on first launch), as far as it changes colors, etc and lots of people just want it standard way as it is on the website.

But it personally like it dark :)

sindresorhus:
If people choose dark mode in OS X, I don't see why they wouldn't want it dark in the app. They can still manually turn it back to light if they prefer, but I think automatic is a good default.

z-ax:
With option to switch - yes, according to system light/dark mode by default is good.

phamdaniel:
I'm not too sure I understood what the final verdict is, so here's how it is right now:

* Default behaviour is to sync the theme with OS X dark mode.
* Checkbox to disable the auto switching.

sindresorhus:
It should follow dark mode, but still have the current ability to manually toggle.

phamdaniel:
Manual toggle is always on. But now, should we keep the ability to disable the auto switching?

sindresorhus:
> But now, should we keep the ability to disable the auto switching?

No

browser.js
```diff
@@ -31,6 +32,13 @@ ipc.on('previous-conversation', previousConversation);
 
 ipc.on('dark-mode', toggleDarkMode);
 
+ipc.on('link-theme', () => {
```
sindresorhus: Just listen to the `platform-theme-changed` event here.

phamdaniel:
Alright, I think I got it this time.

sindresorhus:
Thank you! This is awesome :)

sindresorhus:
Adding you to the repo since you've done some good PRs :)

See: http://openopensource.org

z-ax:
So, now Dark mode is enabled every time I launch app, even if switched to normal before quit. Not so good, I'd say :)

PR #80, opened 2016-04-23
Set attach point for sheets to under the toolbar

phamdaniel:
Tiny UI improvement for OS X.

Before:
screen shot 2016-04-23 at 9 33 59 am

After:
screen shot 2016-04-23 at 9 31 32 am


sindresorhus:
Good idea! :)

PR #81, opened 2016-04-23
Implement ability to resize text

### LINKED ISSUE #73: Zoom
Hello,

I am using Caprine with Ubuntu 15.04. I think it would be nice to have the ability to zoom, as in scale the text/images. It would make reading the text much easier.

phamdaniel:
Fixes #73 

This is implemented in the same way as iMessages, in that instead of zooming in and out on the whole interface this will only affect the conversation's text size.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -4,6 +4,7 @@ const ipc = electron.ipcRenderer;
 const app = electron.remote.app;
 const storage = electron.remote.require('./storage');
 const listSelector = 'div[role="navigation"] > ul > li';
+const conversationSelector = '._4u-c._1wfr > ._5f0v.uiScrollableArea';
 
 ipc.on('show-preferences', () => {
 	// create the menu for the below
@@ -38,6 +39,24 @@ app.on('platform-theme-changed', () => {
 	setDarkMode();
 });
 
+ipc.on('zoom-reset', () => {
+	setZoom(1.0);
+});
+
+ipc.on('zoom-in', () => {
+	const zoomFactor = storage.get('zoomFactor') + 0.1;
+	if (zoomFactor <= 1.4) {
+		setZoom(zoomFactor);
+	}
+});
+
+ipc.on('zoom-out', () => {
+	const zoomFactor = storage.get('zoomFactor') - 0.1;
+	if (zoomFactor >= 0.7) {
+		setZoom(zoomFactor);
+	}
+});
+
 function nextConversation() {
 	const index = getNextIndex(true);
 	document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
@@ -67,6 +86,12 @@ function getNextIndex(next) {
 	return (index % list.length + list.length) % list.length;
 }
 
+function setZoom(zoomFactor) {
+	const node = document.getElementById('zoomFactor');
+	node.innerHTML = `${conversationSelector} {zoom: ${zoomFactor} !important}`;
+	storage.set('zoomFactor', zoomFactor);
+}
+
 // link the theme if it was changed while the app was closed
 if (process.platform === 'darwin') {
 	storage.set('darkMode', app.isDarkMode());
@@ -74,6 +99,18 @@ if (process.platform === 'darwin') {
 
 // activate Dark Mode if it was set before quitting
 setDarkMode();
+
+// Inject a global style node to maintain zoom factor after conversation change.
+// Also set the zoom factor if it was set before quitting.
+document.addEventListener('DOMContentLoaded', () => {
+	const zoomFactor = storage.get('zoomFactor') || 1.0;
+	const style = document.createElement('style');
+	style.type = 'text/css';
+	style.id = 'zoomFactor';
+
+	document.body.appendChild(style);
+	setZoom(zoomFactor);
+});
 
 // it's not possible to add multiple accelerators
 // so need to do this the oldschool way
--- a/browser.js
+++ b/browser.js
@@ -45,6 +45,7 @@ ipc.on('zoom-reset', () => {
 
 ipc.on('zoom-in', () => {
 	const zoomFactor = storage.get('zoomFactor') + 0.1;
+
 	if (zoomFactor <= 1.4) {
 		setZoom(zoomFactor);
 	}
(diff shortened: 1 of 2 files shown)


browser.js
```diff
@@ -38,6 +39,24 @@ app.on('platform-theme-changed', () => {
 	setDarkMode();
 });
 
+ipc.on('zoom-reset', () => {
+	setZoom(1.0);
+});
+
+ipc.on('zoom-in', () => {
+	const zoomFactor = storage.get('zoomFactor') + 0.1;
+	if (zoomFactor <= 1.4) {
```
kevva: Add extra line break between line 47 and 48.

browser.js
```diff
@@ -38,6 +39,24 @@ app.on('platform-theme-changed', () => {
...
+		setZoom(zoomFactor);
+	}
+});
+
+ipc.on('zoom-out', () => {
+	const zoomFactor = storage.get('zoomFactor') - 0.1;
+	if (zoomFactor >= 0.7) {
```
kevva: Same as above.

phamdaniel:
Added in the line breaks

menu.js
```diff
@@ -138,6 +138,32 @@ const darwinTpl = [
...
+			{
+				label: 'Make Text Normal Size',
+				accelerator: 'CmdOrCtrl+0',
+				click() {
+					sendAction('zoom-reset');
+				}
+			},
+			{
+				label: 'Make Text Bigger',
+				accelerator: 'CmdOrCtrl+=',
```
sindresorhus: `=` => `+`

menu.js
```diff
@@ -138,6 +138,32 @@ const darwinTpl = [
 		]
 	},
 	{
+		label: 'View',
+		submenu: [
```
sindresorhus: This sub-menu is the same for all OSes, so just put it in a variable as reuse it.

menu.js
```diff
@@ -138,6 +138,32 @@ const darwinTpl = [
...
+			{
+				label: 'Make Text Normal Size',
+				accelerator: 'CmdOrCtrl+0',
+				click() {
+					sendAction('zoom-reset');
+				}
+			},
+			{
+				label: 'Make Text Bigger',
```
sindresorhus: > Increase Text Size

menu.js
```diff
@@ -138,6 +138,32 @@ const darwinTpl = [
...
+				accelerator: 'CmdOrCtrl+=',
+				click() {
+					sendAction('zoom-in');
+				}
+			},
+			{
+				label: 'Make Text Smaller',
```
sindresorhus: > Decrease Text Size

menu.js
```diff
@@ -138,6 +138,32 @@ const darwinTpl = [
 		]
 	},
 	{
+		label: 'View',
+		submenu: [
+			{
+				label: 'Make Text Normal Size',
```
sindresorhus: > Reset Text Size

browser.js
```diff
@@ -75,6 +102,18 @@ if (process.platform === 'darwin') {
 // activate Dark Mode if it was set before quitting
 setDarkMode();
 
+// Inject a global style node to maintain zoom factor after conversation change.
+// Also set the zoom factor if it was set before quitting.
+document.addEventListener('DOMContentLoaded', () => {
+	const zoomFactor = storage.get('zoomFactor') || 1.0;
+	const style = document.createElement('style');
+	style.type = 'text/css';
```
sindresorhus: This is the default, so no point in setting it.

browser.js
```diff
@@ -38,6 +39,26 @@ app.on('platform-theme-changed', () => {
...
+
+	if (zoomFactor <= 1.4) {
+		setZoom(zoomFactor);
+	}
+});
+
+ipc.on('zoom-out', () => {
+	const zoomFactor = storage.get('zoomFactor') - 0.1;
+
+	if (zoomFactor >= 0.7) {
```
sindresorhus: 0.8

browser.js
```diff
@@ -38,6 +39,26 @@ app.on('platform-theme-changed', () => {
...
 });
 
+ipc.on('zoom-reset', () => {
+	setZoom(1.0);
+});
+
+ipc.on('zoom-in', () => {
+	const zoomFactor = storage.get('zoomFactor') + 0.1;
+
+	if (zoomFactor <= 1.4) {
```
sindresorhus: 1.6

sindresorhus:
Nice! Thanks for working on this @phamdaniel :)

browser.js
```diff
@@ -67,6 +88,12 @@ function getNextIndex(next) {
 	return (index % list.length + list.length) % list.length;
 }
 
+function setZoom(zoomFactor) {
+	const node = document.getElementById('zoomFactor');
+	node.innerHTML = `${conversationSelector} {zoom: ${zoomFactor} !important}`;
```
sindresorhus: `node.textContent` instead?

phamdaniel:
It's always a pleasure getting feedback from you 😄

menu.js
```diff
@@ -138,6 +138,9 @@ const darwinTpl = [
 		]
 	},
 	{
+		label: 'View'
```
sindresorhus: I would just add the submenu here instead of at the bottom, for clarity:

```js
submenu: viewSubmenu
```

phamdaniel:
Done! Though I had to move the the submenu declarations to the top of the file to get access to them.

menu.js
```diff
@@ -16,6 +16,54 @@ function sendAction(action) {
...
+	{
+		label: 'Reset Text Size',
+		accelerator: 'CmdOrCtrl+0',
+		click() {
+			sendAction('zoom-reset');
+		}
+	},
+	{
+		label: 'Increase Text Size',
+		accelerator: 'CmdOrCtrl+Plus',
```
sindresorhus: This shows up as

screen shot 2016-04-25 at 03 18 37

While it should be:

screen shot 2016-04-25 at 03 18 58


Notice the shortcut.

sindresorhus:
Zooming no longer works.

phamdaniel:
The shortcut problem is a bug in Electron. See https://github.com/electron/electron/issues/1507
As for the zooming not working, I'm not too sure... It's still working on my end.

sindresorhus:
> The shortcut problem is a bug in Electron. See electron/electron#1507

Ah, I see. I went with `Cmd+=`. Then at least it works as intended when using the shortcuts.

> As for the zooming, I'm not too sure... It's still working on my end.

My mistake. I managed to use an older Electron version from another branch...

sindresorhus:
Thank you! :) 🍻

PR #82, opened 2016-04-24
Set unread badge for tray

### LINKED ISSUE #75: Add unread message indicator to tray
Hello,

I have tested by having friends message me when the window is up, and minimized to the tray, in either case the libnotify style notifications do not come up.

phamdaniel:
Fixes #75 

This has the same behaviour as the badge on OS X

How it looks on Windows:
tray-image

I don't have Linux, so that remains untested.


sindresorhus:
Woot! :)

You need to update the regex in the `--ignore` option in the build script in package.json. Currently it throws in a built app as the new icon is not included.

sindresorhus:
Works fine now.

sindresorhus:
tumblr_ms960vdlii1qdlh1io1_400

PR #87, opened 2016-04-30
Bump Electron to 0.37.8 and fix deprecated methods

### LINKED ISSUE #83: Fix the shortcuts shown in the text zoom menu items
From https://github.com/sindresorhus/caprine/pull/81#discussion_r60851232.

It's an Electron bug: https://github.com/electron/electron/issues/1507

danhp:
Bump to 0.37.8 to fix #83 
Also use https://github.com/danhp/electron-osx-appearance as `app.isDarkMode()` and `platform-theme-changed` event got deprecated. Ref https://github.com/electron/electron/pull/5282

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,8 +1,9 @@
 'use strict';
 const electron = require('electron');
 const ipc = electron.ipcRenderer;
-const app = electron.remote.app;
 const storage = electron.remote.require('./storage');
+const osxAppearance = require('electron-osx-appearance');
+
 const listSelector = 'div[role="navigation"] > ul > li';
 const conversationSelector = '._4u-c._1wfr > ._5f0v.uiScrollableArea';
 const selectedConversationSelector = '._5l-3._1ht1._1ht2';
@@ -57,10 +58,12 @@ ipc.on('archive-conversation', () => {
 
 ipc.on('dark-mode', toggleDarkMode);
 
-app.on('platform-theme-changed', () => {
-	storage.set('darkMode', app.isDarkMode());
-	setDarkMode();
-});
+if (process.platform === 'darwin') {
+	osxAppearance.onDarkModeChanged(() => {
+		storage.set('darkMode', osxAppearance.isDarkMode());
+		setDarkMode();
+	});
+}
 
 ipc.on('zoom-reset', () => {
 	setZoom(1.0);
@@ -173,7 +176,7 @@ function openDeleteModal() {
 
 // link the theme if it was changed while the app was closed
 if (process.platform === 'darwin') {
-	storage.set('darkMode', app.isDarkMode());
+	storage.set('darkMode', osxAppearance.isDarkMode());
 }
 
 // activate Dark Mode if it was set before quitting
--- a/menu.js
+++ b/menu.js
@@ -26,7 +26,7 @@ const viewSubmenu = [
 	},
 	{
 		label: 'Increase Text Size',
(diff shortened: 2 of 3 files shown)


sindresorhus:
I think you misunderstood. It was only moved: https://github.com/electron/electron/blob/master/docs/api/system-preferences.md#systempreferencesisdarkmode-os-x

danhp:
They were moved but the code turned out quite long so I wrote a wrapper for it.

sindresorhus:
`systemPreferences.isDarkMode()` is still there.

browser.js
```diff
@@ -57,10 +58,12 @@ ipc.on('archive-conversation', () => {
 
 ipc.on('dark-mode', toggleDarkMode);
 
-app.on('platform-theme-changed', () => {
-	storage.set('darkMode', app.isDarkMode());
-	setDarkMode();
-});
+if (process.platform === 'darwin') {
+	osxAppearance.onDarkModeChanged(() => {
```
sindresorhus: `prefs.subscribeNotification('AppleInterfaceThemeChangedNotification', () => {`

danhp:
Woops, I'll put that back.

sindresorhus:
Actually, it's nicer with your module. Let's use it.

sindresorhus:
But update it to use the native `.isDarkMode()` method.

danhp:
Sounds good, I'll update the module.

sindresorhus:
Why limit the module to just appearance. I bet there are lots of useful defaults you can get with `systemPreferences.getUserDefault` other than appearance stuff.

sindresorhus:
And don't forget to add the module to https://github.com/sindresorhus/awesome-electron in two weeks ;)

danhp:
I wanted to keep the module as focused as possible. But yeah, you can do a lot with `getUserDefaults`, the only problem is that you need to know the name and the exact type associated with it.

sindresorhus:
> the only problem is that you need to know the name and the exact type associated with it.

That's why a wrapper module would be so useful. Don't really care if it's part of the current one or a new one. Just a thought if you're looking for project ideas ;)

menu.js
```diff
@@ -26,7 +26,7 @@ const viewSubmenu = [
 	},
 	{
 		label: 'Increase Text Size',
-		accelerator: 'CmdOrCtrl+=',
+		accelerator: 'CmdOrCtrl+Plus',
```
sindresorhus: This requires using `Cmd Shift +` now. ?

danhp: Weird... I get 
image

sindresorhus: I get that too, but it only works when I use the above shortcut. Not what's shown. Try.

danhp: My plus is only accessible through shift so that's the only command I can input...

sindresorhus: Ah, you're right. Just weird that minus is placed on the keyboard so that it doesn't need Shift, while Plus does. Not the best design...

> Unless you mean we have both Cmd+= and Cmd+Plus be mapped to the same action?

That would be favorable. See and vote for: https://github.com/electron/electron/issues/5256

sindresorhus:
Thanks :)

danhp:
http://giphy.com/gifs/shaun-the-sheep-movie-not-my-gif-2016-oscar-nominations-tIeCLkB8geYtW

danhp:
No Problem

Issue #89, opened 2016-05-04
Dark mode issues


sindresorhus:
Some more dark mode issues.

Clicking `Forward` when viewing an image in a chat.

![screen shot 2016-05-04 at 20 41 43](https://cloud.githubusercontent.com/assets/170270/15015926/c07d751a-1238-11e6-87cf-9bdaedb3b661.png)

-

// @danhp

PR #91, opened 2016-05-04
Fix messages forwarding modal

danhp:
Ref #89 
I wasn't able to reproduce the error so this is a hopeful guess I targeted the right css.

But I did fix the transparency issue
Before:
screen shot 2016-05-04 at 11 55 12 am

After:
screen shot 2016-05-04 at 11 53 21 am


sindresorhus:
Thanks :)

PR #99, opened 2016-05-16
Add option to auto hide the menu bar

danhp:
This is blatantly stolen from Slack's Windows app.

Before:
dirty
 
With the option turned on:
clean

It follows native behavior in that if hidden, hitting `Alt` will show the menu

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,6 +56,7 @@ function createMainWindow() {
 		minWidth: 400,
 		minHeight: 200,
 		titleBarStyle: 'hidden-inset',
+		autoHideMenuBar: true,
 		webPreferences: {
 			// fails without this because of CommonJS script detection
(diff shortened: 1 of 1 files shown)


sindresorhus:
Why does it need an option? From what I can remember, Windows does this by default.

danhp:
Slack has it as an option. Probably to always show the menu by default.
Personally think it should just hide it by default though.

sindresorhus:
> Personally think it should just hide it by default though.

👍 Let's do it then :)

danhp:
And it's up :smile:

sindresorhus:
@danhp Cool. Can you do the same commit on Anatine? Just commit directly ;)

danhp:
Sure can

PR #102, opened 2016-05-21
Fix window dragging

### LINKED ISSUE #101: Unable to move window on Mac OS X 10.11.4
Just upgraded from 0.2.1 to 1.4.0 (really need some kind of update notifications) and I'm no longer able to drag the window around to reposition it.  Previously I was able to drag the window title bar to move the app to my 2nd monitor, but the current version ignores mouse drags in the title bar area.  I can kind of move it around the main monitor by resizing it, but I can't get it onto the 2nd monitor this way as OS X 10.11 doesn't allow windows to span multiple monitors.  I'm using Mac OS X 10.11.4, Caprine v1.4.0.

danhp:
Close #101

### DIFF
--- a/browser.css
+++ b/browser.css
@@ -2,6 +2,12 @@ html {
 	overflow: hidden;
 }
 
+/* bind the toolbar as the window's draggable region */
+._36ic._5l-3,
+._5742 {
+	-webkit-app-region: drag;
+}
+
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks :)

Ugh, missed that change in the Electron release notes...

PR #119, opened 2016-08-10
Edit tray and maximize

pierreneter:
When we click on tray icon, if App still open at taskbar, focus app, else show.
Close app when app was maximized, next open app will maximize.

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,9 +47,13 @@ function updateBadge(title) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
> When we click on tray icon, if App still open at taskbar, focus app, else show.

Happy to accept a PR doing just this, but there are just too many changes here for too little gain, so I'm gonna pass on the maximize stuff.

PR #120, opened 2016-08-12
Bounce the dock icon when receiving a new message

### LINKED ISSUE #111: [OS X] Add an option for a Dock bounce animation
Some apps are able to bounce the Dock icon when a notification appears, see:
dock-bounce
It'd be cool if Caprine was able to do it as well.

For me it's a notification that slightly catches my attention but not that much as banners so it's easier for me to skip it if I don't have time for it at the moment.

lukyth:
This pull request would make the dock icon continuously bounce when receiving a new message until Caprine window is focus.

This would resolve @mgol 's #111 

However, I think someone might want to turn this off. So may be this should be an option?

And is it more appropriate to bounce only once rather than bounce continuously? I'm still not sure about this.

I'd love to hear your thoughts! 😃

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,15 +29,13 @@ if (isAlreadyRunning) {
 	app.quit();
 }
 
-function updateBadge(title) {
-	// ignore `Sindre messaged you` blinking
-	if (title.indexOf('Messenger') === -1) {
-		return;
-	}
-
+function getMessageCount(title) {
 	let messageCount = (/\(([0-9]+)\)/).exec(title);
 	messageCount = messageCount ? Number(messageCount[1]) : 0;
+	return messageCount;
+}
 
+function updateBadge(messageCount) {
 	if (process.platform === 'darwin' || process.platform === 'linux') {
 		app.setBadgeCount(messageCount);
 	}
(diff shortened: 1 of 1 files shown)


mgol:
I'd prefer it to bounce once so that it draws my attention but so that I
can ignore it at the same time without further actions needed.

That seems to be an area of custom preferences, though so maybe it should
be settable in settings?


-- 
Michał Gołębiowski

sindresorhus:
Why do you need it to bounce? You already get a notification on new messages.

mgol:
I don't want a banner notification, it's too distracting and it covers part
of the screen which is problematic for me if a particular person sends me
many messages at the moment. It also includes a message preview which I
often don't want.

A bounce is visible but it's gone immediately; nothing obstructs my screen
and it's easier to ignore if I decide so at the moment.


-- 
Michał Gołębiowski

sindresorhus:
Ok, sure, but should be an option and only one bounce.

lukyth:
@sindresorhus @mgol I'll working on it.

sindresorhus:
ping :)

sindresorhus:
Closing for lack of response.

daveseco7:
I can take a look at this if it's ok with you guys.
Regarding the option, I was thinking of adding a new menu entry. Something similar to [this].(http://imgur.com/a/oRvrX)

PR #126, opened 2016-09-13
Add CLI flag to start app minimized

### LINKED ISSUE #64: Start minimized on Linux
#52 So it is not possible, or easily possible, to start Caprine minimized to the tray in Linux.  This should be a option in settings/preferences, or at least available in the command line so it is easier to accomplish.

mckennajones:
Fixes #64 
Added a support for a simple "-m" flag that will cause the app to start up minimized. 
I only tested this on Arch Linux. It should work for all linux distributions, and in theory on MacOS. 

Since there are no other command line options I wasn't sure if we should have a help/usage message or not

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,10 +109,9 @@ app.on('ready', () => {
 		page.insertCSS(fs.readFileSync(path.join(__dirname, 'browser.css'), 'utf8'));
 		page.insertCSS(fs.readFileSync(path.join(__dirname, 'dark-mode.css'), 'utf8'));
 
-		if (process.argv.indexOf("-m") > -1){
+		if (process.argv.indexOf('-m') > -1) {
 			mainWindow.minimize();
-		}
-		else {
+		} else {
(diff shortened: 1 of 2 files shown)


mckennajones:
Sorry that really only should have been one commit :)

index.js
```diff
@@ -108,7 +108,12 @@ app.on('ready', () => {
 	page.on('dom-ready', () => {
 		page.insertCSS(fs.readFileSync(path.join(__dirname, 'browser.css'), 'utf8'));
 		page.insertCSS(fs.readFileSync(path.join(__dirname, 'dark-mode.css'), 'utf8'));
-		mainWindow.show();
+
+		if (process.argv.indexOf('-m') > -1) {
```
sindresorhus: Use https://github.com/substack/minimist

And a long flag instead: `--minimized`

sindresorhus:
Needs to be documented in the readme.

Issue #127, opened 2016-09-19
use cmd/ctrl + number hotkeys for switching channels instead of archive/other


zebapy:
Feels far more common to use `cmd+1` (through 9) hotkeys to switch between chats (another popular desktop FB messenger app does this) over using the hotkeys to archive or delete channels. This is also how browser hotkeys work for switching between tabs. It'd be nice if Caprine had this functionality. 

edit: Slack also mimics this when switching teams.

sindresorhus:
Sure, if someone does a good pull request and comes up with some good shortcuts for those existing actions :)

zebapy:
@sindresorhus I have some of it started. What would you recommend we switch the Mute, Delete, Archive hotkeys to?

zebapy:
@sindresorhus have you noticed these have stopped working?

arielserafini:
"Select Next Conversation" and "Select Previous Conversation" have also stopped working.

zebapy:
Did this stop working yet again?

PR #130, opened 2016-09-20
Cmd hotkey thread switch

zebapy:
Closes https://github.com/sindresorhus/caprine/issues/127

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -95,6 +95,11 @@ function previousConversation() {
 	document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
 }
 
+function jumpToConversation(index) {
+  index = (index >= 1 ? index - 1 : index);
+  document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
+}
+
 // returns the index of the selected conversation
 // if no conversation is selected, returns null.
 function getIndex() {
@@ -187,13 +192,17 @@ document.addEventListener('DOMContentLoaded', () => {
 // it's not possible to add multiple accelerators
 // so need to do this the oldschool way
 document.addEventListener('keydown', event => {
-	if (process.platform === 'darwin' && event.metaKey && event.shiftKey) {
-		if (event.keyCode === 221/* ] */) {
+	if (process.platform === 'darwin' && event.metaKey) {
+		if (event.shiftKey && event.keyCode === 221/* ] */) {
 			nextConversation();
 		}
 
-		if (event.keyCode === 219/* [ */) {
+		if (event.shiftKey && event.keyCode === 219/* [ */) {
 			previousConversation();
 		}
+
+    if (event.keyCode >= 49 /* 1 */ && event.keyCode <= 57 /* 9 */) {
+      jumpToConversation(parseInt(event.key));
+    }
 	}
 });
--- a/browser.js
+++ b/browser.js
@@ -95,8 +95,8 @@ function previousConversation() {
 	document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
 }
 
-function jumpToConversation(index) {
-	index = (index >= 1 ? index - 1 : index);
+function jumpToConversation(key) {
+	const index = (key >= 1 ? key - 1 : key);
 	document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
 }
 
--- a/browser.js
+++ b/browser.js
@@ -96,7 +96,7 @@ function previousConversation() {
 }
 
 function jumpToConversation(key) {
-	const index = (key >= 1 ? key - 1 : key);
+	const index = key === 0 ? key : key -1;
 	document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
 }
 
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -34,6 +38,19 @@ Desktop notifications can be turned on in Preferences.
...
+|---|---|
+|Cmd + n|New conversation|
+|Cmd + f|Search conversations|
+|Cmd + d|Toggle Dark mode|
+|Cmd + 1 through Cmd + 9|Jump to conversation|
+|Cmd + shift + m|Mute conversation|
+|Cmd + shift + a|Archive conversation|
+|Cmd + shift + d|Delete conversation|
+|Cmd + shift + [|Previous conversation|
+|Cmd + shift + ]|Next conversation|
```
sindresorhus: I would put the description first. And align the `|` in the source too.

And use `<kbd>Cmd</kbd>` etc, for the keys.

browser.js
```diff
@@ -187,13 +192,17 @@ document.addEventListener('DOMContentLoaded', () => {
...
+		if (event.shiftKey && event.keyCode === 221 /* ] */ ) {
 			nextConversation();
 		}
 
-		if (event.keyCode === 219/* [ */) {
+		if (event.shiftKey && event.keyCode === 219 /* [ */ ) {
 			previousConversation();
 		}
+
+		if (event.keyCode >= 49 /* 1 */ && event.keyCode <= 57 /* 9 */ ) {
```
sindresorhus: Use the new `.key` property instead and you can define the name of the keys instead of code: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

readme.md
sindresorhus: shift => Shift

browser.js
```diff
@@ -187,13 +192,17 @@ document.addEventListener('DOMContentLoaded', () => {
...
+			jumpToConversation(parseInt(event.key));
```
sindresorhus: `parseInt(event.key, 10)`

browser.js
```diff
@@ -95,6 +95,11 @@ function previousConversation() {
 	document.querySelectorAll(listSelector)[index].firstChild.firstChild.click();
 }
 
+function jumpToConversation(index) {
+	index = (index >= 1 ? index - 1 : index);
```
sindresorhus: The code would read clearer like this:

```js
const index = key === 0 ? key : key - 1;
```

browser.js
zebapy: Is it really worth doing that here? You'd have to parse the keyCode then do a comparison. This already uses `keyCode` above. The main reason I pass `event.key` to the jumpToConversation func is so we don't need a map of keycodes -> number to check.

readme.md
zebapy: All set I think. Not sure about the "align | in the source" meaning.

browser.js
sindresorhus: @zebapy Seems simpler and more readable to me:

```js
const num = parseInt(event.key, 10);

if (num >= 1 && num <= 9) {
  jumpToConversation(num);
}
```

browser.js
sindresorhus: Actually, why even check for `0`? We don't pass it down anyways.

readme.md
```diff
@@ -40,17 +40,17 @@ Desktop notifications can be turned on in Preferences.
...
+--------------------- | -----------------------
+New conversation      | <kbd>Cmd</kbd> + <kbd>n</kbd>
+Search conversations  | <kbd>Cmd</kbd> + <kbd>f</kbd>
+Toggle Dark mode      | <kbd>Cmd</kbd> + <kbd>d</kbd>
+Jump to conversation  | <kbd>Cmd</kbd> + <kbd>1</kbd> through <kbd>Cmd</kbd> + <kbd>9</kbd>
+Mute conversation     | <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>m</kbd>
+Archive conversation  | <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>a</kbd>
+Delete conversation   | <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>d</kbd>
+Previous conversation | <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>[</kbd>
+Next conversation     | <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>]</kbd>
```
sindresorhus: You don't really need the `+` I think. The `kbd` tags makes it clear.

zebapy: Ok I'll remove that. How do you feel about the `through` in the jump command?

sindresorhus: I'm good with that.

sindresorhus:
And https://github.com/sindresorhus/caprine/pull/130#discussion_r79662413

sindresorhus:
Looks great! Thank you :)

zebapy:
@sindresorhus awesome! Thanks for the good review pointers. I don't contribute to OSS much. I also just committed some code style changes to fix the travis failed build I think. Want me to PR that? 

Also running the `npm test` doesn't bring up the errors that travis did. running xo locally seemed to only show warnings whereas travis showed errors.

sindresorhus:
Your contribution was very good. I encourage you to continue and you'll soon get hooked on open source ;)

PR #133, opened 2016-09-26
Update package.json to include dependency for minimist

sidspencer:
building the app currently succeeds, but fails to run with an error message about needing minimist@^0.0.8. It just needed it to be listed in package.json as a dependency. Very small fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "electron-config": "^0.2.1",
     "electron-context-menu": "^0.6.0",
     "electron-debug": "^1.0.0",
-    "electron-dl": "^1.0.0"
+    "electron-dl": "^1.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -34,7 +34,8 @@
     "electron-config": "^0.2.1",
     "electron-context-menu": "^0.6.0",
     "electron-debug": "^1.0.0",
-    "electron-dl": "^1.0.0"
+    "electron-dl": "^1.0.0",
+    "minimist": "^0.0.8"
```
kevva: Why not add the latest version?

sidspencer: I only used 0.0.8 since that is the version the xo library called for. I have no issues with using a newer version -- sounds good to me.

PR #136, opened 2016-10-29
Vibrant window for dark theme

oskarkrawczyk:
screen shot 2016-10-29 at 18 14 21

If something's not working run:

```
npm run rebuild
```

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -185,7 +185,7 @@ document.addEventListener('DOMContentLoaded', () => {
 	// prevent flash of white on startup when in dark mode
 	// TODO: find a CSS only solution
 	if (config.get('darkMode')) {
-		document.documentElement.style.backgroundColor = '#192633';
+		document.documentElement.style.backgroundColor = 'rgba(0,0,0,0)';
 	}
 });
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "caprine",
   "productName": "Caprine",
-  "version": "1.6.0",
+  "version": "1.6.1",
   "description": "Unofficial Facebook Messenger app",
   "license": "MIT",
   "repository": "sindresorhus/caprine",
--- a/package.json
+++ b/package.json
@@ -34,12 +34,14 @@
     "electron-config": "^0.2.1",
     "electron-context-menu": "^0.6.0",
     "electron-debug": "^1.0.0",
-    "electron-dl": "^1.0.0"
+    "electron-dl": "^1.0.0",
+    "electron-vibrancy": "^0.1.0"
   },
   "devDependencies": {
     "electron": "^1.4.0",
     "electron-packager": "^8.0.0",
-    "xo": "*"
+    "xo": "*",
+    "electron-vibrancy": "^0.1.0"
   },
   "xo": {
     "esnext": true,
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "caprine",
   "productName": "Caprine",
-  "version": "1.6.1",
+  "version": "1.6.0",
   "description": "Unofficial Facebook Messenger app",
   "license": "MIT",
   "repository": "sindresorhus/caprine",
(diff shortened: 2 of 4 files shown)


browser.js
```diff
@@ -185,7 +185,7 @@ document.addEventListener('DOMContentLoaded', () => {
 	// prevent flash of white on startup when in dark mode
 	// TODO: find a CSS only solution
 	if (config.get('darkMode')) {
-		document.documentElement.style.backgroundColor = '#192633';
+		document.documentElement.style.backgroundColor = 'rgba(0,0,0,0)';
```
sindresorhus: `rgba(0,0,0,0)` => `rgba(0, 0, 0, 0)`

package.json
```diff
@@ -34,12 +34,14 @@
...
-    "electron-dl": "^1.0.0"
+    "electron-dl": "^1.0.0",
+    "electron-vibrancy": "^0.1.0"
   },
   "devDependencies": {
     "electron": "^1.4.0",
     "electron-packager": "^8.0.0",
-    "xo": "*"
+    "xo": "*",
+    "electron-vibrancy": "^0.1.0"
```
sindresorhus: It shouldn't be here.

oskarkrawczyk:
> Why only for the dark theme?

No reason, looks good on the dark one.

> Most apps using vibrancy only does it for the sidebar. See the Notes.app for example.

Most do, indeed. I like vibrancy so I use it for the whole window.

> The vibrancy effect is very buggy. It keeps glitching when I hover over different elements.

Interesting, I'm not getting this.

Let me know if you're willing to pull this in if I fix the mentioned things. Otherwise go ahead and close.

sindresorhus:
> Interesting, I'm not getting this.

Hmm, seems to have gone away. Maybe because I bumped Electron.

> Let me know if you're willing to pull this in if I fix the mentioned things. Otherwise go ahead and close.

Initial impression, I like it. Gonna run for a little while and see how I feel about it then.

-

I think the navbar should be made vibrant too.

sindresorhus:
Would appreciate it if people looking here could share their opinion on this.

## Before

screen shot 2016-11-04 at 23 54 48



## After
screen shot 2016-11-04 at 23 54 19

oskarkrawczyk:
> Hmm, seems to have gone away. Maybe because I bumped Electron.

I've noticed there's some artifacting when playing inline video – most likely due the fact that it's hardware accelerated.

> I think the navbar should be made vibrant too.

Actually, the nav-bar is semi-transparent, but the interface is built in a way that scrollable content doesn't go under it. And even if it did, I've tried turning on Chromium experimental features and adding `backdrop-filter:blur` but got a lot of artifacting (again, most likely because it's triggering GPU to do the drawing)  

> Initial impression, I like it. Gonna run for a little while and see how I feel about it then.

Sure. I've been running it for some time before creating a PR, just to see if the full-window vibrancy won't get annoying but it didn't. But maybe it's because I'm a sucker for all that blurs!

oskarkrawczyk:
Also, I think it should be possible to have Vibrancy an option, just like dark mode.

sindresorhus:
I tweeted about getting some opinions on this: https://twitter.com/sindresorhus/status/794584600664150017

sindresorhus:
> Also, I think it should be possible to have Vibrancy an option, just like dark mode.

I prefer the least amount of option. So not gonna do an option unless vibrancy is very unpopular or split opinions about it.

ashtonwar:
Does it still look pretty when the background isn't a desktop wallpaper? ie with another window / application behind it.

oskarkrawczyk:
@ashtonwar
screen shot 2016-11-04 at 18 23 56

kimroen:
I enjoy this look - I would use it.

terkelg:
ike itl I – the application feels more lightweight, but it's really just a matter of taste.

ivancuric:
Maybe just for the sidebar... Not sure. I'm not a fan of transparent interfaces. I feel the same way about transparent terminal windows.

ZakariaRidouh:
@ivancuric I guess there should be a simple option giving the choice?

danhp:
I have a preference for keeping the vibrancy contained to the sidebar so that it matches the rest of the system. Also https://github.com/electron/electron/pull/7898 has landed and is released as of `v1.4.7` so I'd use that instead of the module.

PR #140, opened 2016-11-15
Add support for showing/hiding the sidebar

orta:
Hey there, I wanted to be able to both focus on working on a bot, and to be able to easily take [screenshots](https://github.com/artsy/Mitosis/issues/8) of my work, so I added the ability to toggle the sidebar. 

It's your call whether this feature is worth it for the general public, but I thought it would be at least worth presenting as an option for you.

wide mode, no sidebar:
screen shot 2016-11-15 at 10 08 55

constrained width, invisible avatars
screen shot 2016-11-15 at 10 11 44

( I couldn't find a way to override the injected stylesheet's `left: "74px !important"` - didn't matter too much to me for my purposes )

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -85,6 +85,18 @@ ipc.on('zoom-out', () => {
 	}
 });
 
+ipc.on('toggle-sidebar', () => {
+	const root = document.querySelectorAll('[data-reactroot]')[0];
+
+	// Hiding the sidebar completely removes it on wide windows, but leaves it
+	// empty on smaller windows
+	const sidebar = root.firstElementChild.firstElementChild;
+	const display = sidebar.style.display;
+	const shouldHide = (display === '');
+	const newDisplay = shouldHide ? 'none' : '';
+	sidebar.style.display = newDisplay;
+});
+
 function nextConversation() {
 	const index = getNextIndex(true);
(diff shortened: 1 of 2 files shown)


sindresorhus:
Yeah, sure if you fix this issue when the windows is small.

browser.js
```diff
@@ -85,6 +85,18 @@ ipc.on('zoom-out', () => {
 	}
 });
 
+ipc.on('toggle-sidebar', () => {
+	const root = document.querySelectorAll('[data-reactroot]')[0];
```
sindresorhus: If you just need the first element, you can use `querySelector`.

menu.js
```diff
@@ -21,6 +21,16 @@ function sendAction(action) {
 
 const viewSubmenu = [
 	{
+		label: 'Toggle Sidebar Avatars',
```
sindresorhus: > Toggle Sidebar

menu.js
```diff
@@ -21,6 +21,16 @@ function sendAction(action) {
...
+		accelerator: 'CmdOrCtrl+Shift+H',
```
sindresorhus: Why H? Wouldn't S for sidebar be better?

Issue #138, opened 2016-11-09
Better highlighting for current tab


adrianvoica:
First, great job!
Second, please make a better highlighting system for the current tab (current open user in the left pane).
I ended up sending some messages to a different user, since I couldn't quite visually distinguish between the currently selected and its neighboring tabs.

![image](https://cloud.githubusercontent.com/assets/3178286/20136765/468df2b2-a67f-11e6-84c0-6c3a97198f15.png)

Cheers!

sindresorhus:
👍 Pull request welcome :)

PR #144, opened 2016-11-29
Better highlighting for current tab

tristanlepachyderme:
#138

### DIFF
--- a/browser.css
+++ b/browser.css
@@ -68,6 +68,11 @@ html {
 	margin-left: 65px !important;
 }
 
+/* better highlightment for the current thread (left side bar)*/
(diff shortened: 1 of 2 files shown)


sindresorhus:
In light mode there's no difference. I only see your change for a moment when I click the tab. In dark mode it works correctly, but when I click the tab of the open conversation, it changes slightly. Also it's the wrong blue color.

sindresorhus:
Closing for lack of response.

ZachGawlik:
@sindresorhus What blue color would you like to see? For light mode, the only issue was not including `!important`

PR #149, opened 2016-12-14
add a menu option for Reload Window

steelgr:
I tried to add an event handler that refreshes the page when computer wakes up from sleep but there are issues when computer is no longer connected to internet. This is a not so elegant solution that helps issues like opening the app while not connected to internet which caused the window to be unresponsible even when computer eventually connects.

### DIFF
--- a/menu.js
+++ b/menu.js
@@ -42,13 +42,13 @@ const viewSubmenu = [
 		}
 	},
     {
-        label: 'Reload Window',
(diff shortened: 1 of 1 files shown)


steelgr:
Will reopen when tests are passed

sindresorhus:
There's never a need to open a new PR. You can always fix the existing PR. Creating duplicate PRs just creates useless noise in the issue tracker...

PR #150, opened 2016-12-14
Add a menu item to reload the window

steelgr:
I tried to add an event handler that refreshes the page when computer wakes up from sleep but there are issues when computer is no longer connected to internet. This is a not so elegant solution that helps issues like opening the app while not connected to internet which caused the window to be unresponsible even when computer eventually connects.
(Replaced spaces with tabs and rebased my repo so commits stay clear)

### DIFF
--- a/menu.js
+++ b/menu.js
@@ -41,6 +41,14 @@ const viewSubmenu = [
 			sendAction('zoom-out');
 		}
 	}
+	{
+		label: 'Reload Window',
+		accelerator: 'CmdOrCtrl+R',
+		click() {
+			const win = BrowserWindow.getAllWindows()[0];
+			win.reload();
+		}
+	}
 ];
 
 const helpSubmenu = [
--- a/menu.js
+++ b/menu.js
@@ -40,7 +40,7 @@ const viewSubmenu = [
 		click() {
 			sendAction('zoom-out');
 		}
-	}
+	},
 	{
 		label: 'Reload Window',
 		accelerator: 'CmdOrCtrl+R',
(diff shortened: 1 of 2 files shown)


steelgr:
First time working with npm project,didn't know that "npm test" command existed. Sorry

menu.js
```diff
@@ -40,6 +40,14 @@ const viewSubmenu = [
 		click() {
 			sendAction('zoom-out');
 		}
+	},
+	{
+		label: 'Reload Window',
+		accelerator: 'CmdOrCtrl+R',
+		click() {
```
sindresorhus: The window is passed as an argument here, so you can use it instead.

menu.js
```diff
@@ -40,6 +40,14 @@ const viewSubmenu = [
 		click() {
 			sendAction('zoom-out');
 		}
+	},
```
sindresorhus: There should be a separator here.

sindresorhus:
@steelgr Also see: https://github.com/sindresorhus/caprine/pull/150/files#r92523622

steelgr:
How can I fix this? I didn't edit this file

sindresorhus:
No, I mean the inline comment.

menu.js
steelgr: Where is it passed as an argument? The use of "BrowserWindow.getAllWindows()[0]" is not correct?

steelgr:
I fixed the Travis issue with the missing variables and I merged with the version 1.8.0
ec26b700de6b6ea492027dc3334544ea7892aefe
a46b4813a1b6f6cc637e0d0c3977ffc00ac62951

sindresorhus:
I already implemented it another way: https://github.com/sindresorhus/caprine/commit/7472fc0d95372a882168b58f206632d77c539415 Didn't want the menu item.

PR #153, opened 2016-12-26
Remove redundant scrollbar in message entry

### LINKED ISSUE #490: Redundant scrollbar in message entry (Windows)
A redundant scrollbar appears if the window size is small enough that the right sidebar is hidden and the left sidebar is not toggled to hidden.

not good

ivancuric:
Remove unnecessary overflow-scroll on input field on Windows.

Fixes #490


sindresorhus:
Unless this is fixing a bug Caprine is introducing, I'm not really interested in having lots of fixing for tiny issues caused by messenger.com.

ivancuric:
It's a messenger.com issue.

PR #154, opened 2017-01-03
Disable forced greyscale font antialiasing

ivancuric:
The white font on colored backgrounds ends up being too thin on non-retina displays. Adding <code>&#8209;webkit&#8209;font&#8209;smoothing: initial;</code> fixes this by reverting it to the default behaviour.

Blink on retina displays will use greyscale rendering either way — this change does not affect it. This rule has no effect on Windows systems.

### DIFF
--- a/browser.css
+++ b/browser.css
@@ -3,7 +3,6 @@ html {
 }
 
 /* Bind the toolbar as the window's draggable region */
-
 ._36ic._5l-3,
 ._5742 {
 	-webkit-app-region: drag;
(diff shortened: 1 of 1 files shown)


sindresorhus:
Can you remove the unrelated changes?

sindresorhus:
Still incorrect. Please look at the diff tab.

sindresorhus:
Still incorrect. Just look at the diff...

sindresorhus:
Actually, I don't really care about non-Retina screens and this has been too much back and forth. Gonna pass on this, sorry.

PR #156, opened 2017-01-06
Fix conversation CSS selector

zebapy:
fixes https://github.com/sindresorhus/caprine/issues/127 again


sindresorhus:
This would only work for English locale. Can construct a selector without using the ARIA label?

PR #160, opened 2017-01-19
Add option to reload the content until problems with waking from slee…

sander85:
…p are fixed

Signed-off-by: Sander Lepik <sander@lepik.eu>

### DIFF
--- a/menu.js
+++ b/menu.js
@@ -121,6 +121,13 @@ const darwinTpl = [
 				}
 			},
 			{
+				label: 'Reload',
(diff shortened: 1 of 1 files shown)


sindresorhus:
You can already reload with Cmd/Ctrl+R in the latest version. Not interested in a menu option for it.

sander85:
OK. Didn't know that. But thanks for the option.

thagorastos:
Could you make the keyboard shortcut keyboard-layout-independent? When I don't write English, Cmd+R does not work.

PR #163, opened 2017-02-03
Add vibrancy effect to window on macOS

gerhardberger:
This PR adds a vibrancy effect to the app window.

I missed this for some time in the app so I added it and thought I'd share it maybe others would like it. This differs from the other similar PR (#136) that it doesn't use a separate module, but Electron's API and it works with normal and dark mode:

screen shot 2017-02-03 at 4 04 23 pm
screen shot 2017-02-03 at 4 04 41 pm

---

Closes #136


sindresorhus:
Sorry about the late reply. This is excellent. Thank you so much! :)

PR #169, opened 2017-03-07
Dock bounce animation

daveseco7:
Implements what was discussed in #120  #111 . It only bounces once and the animation has a duration of 1 second (`electron informal bounce`). The Default behaviour is bouncing turned **off** and it can be changed on the menu option or with the keyboard accelerator: **CMD + L**
This feature **can only be used in MacOS** since it relies on `app.dock.bounce()` method, which is only available for the `darwin` system, as far as I know.


daveseco7:
Should I look what's up with appveyor failing? It seems that is trying to locate an `appveyor.yml` file holding the build configs.

sindresorhus:
I don't see the point in having a keyboard shortcut for this. Either you want dock bouncing or you don't.

sindresorhus:
> Should I look what's up with appveyor failing?

No, it's unrelated and fixed on master.

index.js
```diff
@@ -14,13 +14,13 @@ require('electron-context-menu')();
...
 let mainWindow;
 let isQuitting = false;
+let prevMessageCount = 0;
 
 const isAlreadyRunning = app.makeSingleInstance(() => {
 	if (mainWindow) {
 		if (mainWindow.isMinimized()) {
 			mainWindow.restore();
 		}
-
```
sindresorhus: Don't do unrelated changes.

menu.js
```diff
@@ -114,6 +114,15 @@ const darwinTpl = [
 				}
 			},
 			{
+				label: 'Dock Bounce',
+				accelerator: 'Cmd+L',
+				type: 'checkbox',
+				checked: false,
+				click() {
+					sendAction('toggle-bounce');
```
sindresorhus: No point in using IPC for this. You can just set the config right here.

sindresorhus:
Can you also fix the merge conflict?

daveseco7:
The build is failing in Travis as a result of a lint error. I already tried to correct this [here](https://github.com/sindresorhus/caprine/pull/169/#pullrequestreview-26447271) but I revert it as you asked.

PR #172, opened 2017-03-12
Support retina display

dangh:
correct dpr cookie to screen scale factor

---

Fixes #172


sindresorhus:
Yay! Thank you so much for fixing this @dangh. The blurry icons has been annoying me for so long.

I'm about to sleep now, but will review and merge tomorrow morning :)

sindresorhus:
Thank you! :)

PR #174, opened 2017-03-22
Fix #173

### LINKED ISSUE #173: Messages have an ugly white border using dark mode.
This is probably something that Facebook changed on their side, but I didn't see this on my Ubuntu machine so I figured I'd create an issue. On Windows, message bubbles suddenly have a white border around them when using dark mode where they didn't before. It's present in both version 1.7 and 1.8.

image

Edit: switched back to Ubuntu and the borders are now there too.

dudeofawesome:
Facebook updated Messenger to add a white border around messages. This makes that border transparent in dark mode.

It also fixes a rendering issue with the header image in the readme


dudeofawesome:
I'm not sure why it's trying to merge in those old commits from my last PR. They don't show up in the diff though.

sindresorhus:
Thanks :)

PR #175, opened 2017-03-24
Update readme.md to fix broken link on screenshot.

hugomd:
I've noticed this has happened across quite a few repositories, unfortunately 😢

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,8 +2,9 @@
 
 > Unofficial Facebook Messenger app
 
-<br>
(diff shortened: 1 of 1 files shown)


sindresorhus:
The correct fix is to use HTML for the link too. It's using `` for a reason.

hugomd:
@sindresorhus I think GitHub changed something which broke it — I can change it to  if you'd like?

sindresorhus:
Yes

hugomd:
@sindresorhus Done 🎉

sindresorhus:
Thanks :)

PR #179, opened 2017-04-06
Improve highlight for selected conversation

### LINKED ISSUE #138: Better highlighting for current tab

vinnyang:
Fixes #138 

- Used the blue from message bubbles with different opacity in
dark/light modes.
- Slight tweaked the color of the latest message subtitle, so it would be easier to read in light mode.

### DIFF
--- a/browser.css
+++ b/browser.css
@@ -126,7 +126,7 @@ a,
 ._1ht1._1ht2 {
 	background-color: rgba(0, 132, 255, 0.3) !important;
 }
-/* Contact list: latest message, under conact name*/
+/* Contact list: latest message, under contact name */
 ._1ht1 ._1htf {
 	color: rgba(0, 0, 0, 0.7);
 }
--- a/browser.css
+++ b/browser.css
@@ -127,7 +127,3 @@ a,
 	background-color: rgba(0, 132, 255, 0.3) !important;
 
 }
-/* Contact list: selected person, latest message */
-._1ht1._1ht2 ._1htf {
-	color: rgba(0, 0, 0, 0.4);
-}
--- a/browser.css
+++ b/browser.css
@@ -124,7 +124,7 @@ a,
 
 /* Contact list: person (selected) */
 ._1ht1._1ht2 {
-	background-color: rgba(0, 132, 255, 0.8) !important;
+	background-color: rgba(0, 132, 255, 0.3) !important;
 
 }
 /* Contact list: selected person, latest message */
--- a/browser.css
+++ b/browser.css
@@ -125,5 +125,8 @@ a,
 /* Contact list: person (selected) */
 ._1ht1._1ht2 {
 	background-color: rgba(0, 132, 255, 0.3) !important;
-
+}
+/* Contact list: latest message, under conact name*/
+._1ht1 ._1htf {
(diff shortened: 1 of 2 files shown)


browser.css
```diff
@@ -121,3 +121,13 @@ a,
...
+
+/* Contact list: person (selected) */
+._1ht1._1ht2 {
+	background-color: rgba(0, 132, 255, 0.8) !important;
+
+}
+/* Contact list: selected person, latest message */
+._1ht1._1ht2 ._1htf {
+	color: rgba(0, 0, 0, 0.4);
+}
```
sindresorhus: This doesn't actually make it any more readable in either light or dark mode.

browser.css
```diff
@@ -121,3 +121,13 @@ a,
 ._4eby {
 	border-radius: 2px !important;
 }
+
+/* Contact list: person (selected) */
+._1ht1._1ht2 {
+	background-color: rgba(0, 132, 255, 0.8) !important;
```
sindresorhus: `0.8` is too strong of a color. I would go with `0.3`.

dark-mode.css
```diff
@@ -138,7 +138,7 @@ html.dark-mode ._1ht1 {
 
 /* Contact list: person (selected) */
 html.dark-mode ._1ht1._1ht2 {
-	background: rgba(255, 255, 255, 0.03) !important;
+	background-color: rgba(0, 132, 255, 0.6) !important;
```
sindresorhus: I think you instead should match the macOS dark mode highlight color, which is a gradient that starts at #1871DC and ends at #1870DB.

browser.css
vinnyang: @sindresorhus what color would you suggest for this?

dark-mode.css
vinnyang: Sure, I will change that.

browser.css
vinnyang: @sindresorhus I reverted this change, that latest message bit seems pretty readable in both modes with the new background colors you suggested.

sindresorhus:
Looks pretty good now, but I think the https://github.com/sindresorhus/caprine/pull/179#discussion_r111705760 text needs to be a bit darker in light mode, as it's not very legible.

In dark mode, the date screen shot 2017-04-17 at 16 04 53 when selected
 needs to be a bit lighter.

vinnyang:
@sindresorhus updated accordingly, please check again when you have a chance, thanks!

dark-mode.css
```diff
@@ -138,7 +138,7 @@ html.dark-mode ._1ht1 {
 
 /* Contact list: person (selected) */
 html.dark-mode ._1ht1._1ht2 {
-	background: rgba(255, 255, 255, 0.03) !important;
+	background: linear-gradient(-45deg, #1871DC, #1870DB) !important
```
sindresorhus: Should be `linear-gradient(180deg, #1871DC, #1870DB) !important;`

sindresorhus:
Looks great. Thank you! :)

Issue #178, opened 2017-04-05
Caprine 2.0.0 window can't be resized or maximized on Windows 10


mgol:
Caprine 2.0.0 window can't be resized or maximized on Windows 10 while it works fine on macOS. I think it worked in 1.8.0.

The window on Windows 10 is also quite large, on a full HD display it takes most of the screen so when you take that all into account, positioning the window so that you can see everything is quite hard.

zackschuster:
Can confirm the resizing issue. On my 2560x1600 display it fills (roughly) 1/4 of the screen.

BlueHatbRit:
Roughly 1/4 of the screen on my 1920x1080 as well with the same issue.

BlueHatbRit:
I've submitted a fix for this a PR 180

PR #180, opened 2017-04-08
Force the Window to be resizable for all platforms

BlueHatbRit:
This implements #178.

On Windows 10 the window couldn't be resized, I added `win.setResizable(true)` into the window creation function. Sadly there doesn't seem to be an option we can pass into the constructor so this had to be done after.

As this issue has only be found on Windows 10, it could be wrapped in a platform check but as I understand, we want to ensure it can be resized on all platforms anyway.

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,6 +110,8 @@ function createMainWindow() {
 		}
 	});
 
+	win.setResizable(true);
+
 	if (process.platform === 'darwin') {
 		win.setSheetOffset(40);
 	}


danhp:
From the [electron docs](https://electron.atom.io/docs/api/browser-window/#new-browserwindowoptions) the resizable option should already default to `true`. It would be great if you could open an issue over at https://github.com/electron/electron as this is more of a hack.

BlueHatbRit:
I had a look at creating a small project to recreate it to post it as an issue on electron but it seems I can't recreate it on a bare electron app. I'll dig a bit further and see if I can find the issue within Caprine instead.

obernardovieira:
Does this solves the problem for you ? I have tried the same but It didn't worked.
I didn't reported the issue before because I want to test on linux. It worked just fine, but on windows I cant resize or maximize and I this is not a electron bug (I think) because I'm working on others electron based projects and it's working just fine.

BlueHatbRit:
@obernardovieira I created this patch on my Windows 10 machine and the problem disappears. As @danhp has said it's probably more of a hack but I'm struggling to pin down the piece of code which is causing the original issue. Suggestions are welcome!

veniversum:
Issue is caused by setting `BrowserWindow` option `transparent` to `true` in [index.js](https://github.com/sindresorhus/caprine/blob/master/index.js#L105). Forcing window to be resizable may cause [unforeseen issues](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md#limitations).

Seems to be a regression bug from @163. Maybe we can do a platform check during creation of window, and set the transparent flag for darwin only?

mgol:
> Forcing window to be resizable may cause unforeseen issues.

Interesting. The link says:

> Transparent windows are not resizable. Setting resizable to true may make a transparent window stop working on some platforms.

which suggests transparent windows are not resizable in general, regardless of the platform used.

veniversum:
@mgol I think this is a Windows specific issue, regarding resizing of transparent windows. See electron/electron/issues/6107

mgol:
I see, thanks for the link!

In that case I think it's far more important for the window to be resizable than to be transparent so I'd favor disabling the transparency on Windows.

BlueHatbRit:
Apologies for my delay in getting back to this. I think disabling transparency is the way to go as well given what's been said, I'll go ahead and revert the current commit and disable transparency on Windows instead. That way the change is along side this PR.

PR #182, opened 2017-04-10
Remove greyscale font antialiasing

ivancuric:
The white font weight on colored backgrounds is way too light on non-retina displays:
image

With subpixel aa:
image

For retina displays, the `initial` property defaults to `greyscale` on Blink anyway.


sindresorhus:
Thanks, but I don't want to maintain nitpick changes. The Facebook Messenger team will fix this if it's actually an issue for users.

PR #185, opened 2017-04-16
readme.md: add hyphen in homebrew-cask

vitorgalvao:
The correct name is with the hyphen. Source: I’m the longest active maintainer of Homebrew-Cask.

The website version is wrong, the Github page is the canonical one. It has an hyphen.


sindresorhus:
That looks weird. I would have expected either `Homebrew Cask` or `homebrew-cask`. Anyways, fixed in master.

PR #188, opened 2017-04-18
Fix the @name-popover in the chat input box in dark mode

### LINKED ISSUE #187: Wrong colors when using names
The dialog that pops up when using @ in groups has wrong colors in dark-mode.
I guess the background-color of the dialog is wrong.

Please see picture.

image

OlaAlsaker:
Found the class _13iv, and changed the background color.

Fixes #187


sindresorhus:
Thanks :)

I also noticed the box jumps a little bit when selecting different people in dark mode. Would be nice to fix that too.

PR #194, opened 2017-05-01
Fix disappearing shadow on macOS

### LINKED ISSUE #183: Window shadow disappearing when minimized and restored on macOS
When Caprine is minimized (either by button or by shortcut) and then restored to front, the shadow under the window disappears. I've also noticed this when toggling vibrancy effect, but I can't reproduce it in this scenario. Here's the screenshot attached.

screen shot 2017-04-12 at 10 11 33

How to reproduce:

- open Caprine
- minimize it
- click Caprine icon to restore it
- shadow should appear behind window, but it doesn't

danhp:
Transparent windows aren't needed to get vibrancy to work. Windows must simply not have any backgrounds to them.

Previous change:
image

Fixes #183

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,7 +102,6 @@ function createMainWindow() {
 		titleBarStyle: 'hidden-inset',
 		autoHideMenuBar: true,
 		darkTheme: isDarkMode, // GTK+3
-		transparent: process.platform === 'darwin',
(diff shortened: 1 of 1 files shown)


sindresorhus:
@danhp Thanks for looking into this :)

PR #197, opened 2017-05-01
Add unread count and taskbar overlay icon for Windows

veniversum:
We can make use of overlay icons to display unread thread count in Windows, similar to how badge counts work in macOS & Unity. 
image

Since tray icons are hidden by default, this also eliminates the need for the user to pin the tray icon.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -80,13 +80,8 @@ function renderOverlayIcon(messageCount) {
 	ctx.fill();
 	ctx.textAlign = 'center';
 	ctx.fillStyle = 'white';
-	if (messageCount > 99) {
-		ctx.font = '90px sans-serif';
-		ctx.fillText('99', 64, 96);
-	} else {
-		ctx.font = '90px sans-serif';
-		ctx.fillText(String(messageCount), 64, 96);
-	}
+	ctx.font = '90px sans-serif';
+	ctx.fillText(String(Math.min(99, messageCount)), 64, 96);
 	return canvas;
 }
 
--- a/browser.js
+++ b/browser.js
@@ -68,6 +68,31 @@ function setVibrancy() {
 	document.documentElement.style.backgroundColor = 'transparent';
 }
 
+function renderOverlayIcon(messageCount) {
+	const canvas = document.createElement('canvas');
+	canvas.height = 128;
+	canvas.width = 128;
+	canvas.style.letterSpacing = '-5px';
+	const ctx = canvas.getContext('2d');
+	ctx.fillStyle = '#f42020';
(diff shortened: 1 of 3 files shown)


veniversum:
On Windows 10, make sure that 'Show badges on taskbar buttons' is enabled in taskbar settings. Also badges are hidden if small taskbar buttons are used.

image

sindresorhus:
That did it. Then we should document that. Badges are very useful, but I doubt most know they'll have to manually activate it.

browser.js
```diff
@@ -68,6 +68,31 @@ function setVibrancy() {
...
+	ctx.textAlign = 'center';
+	ctx.fillStyle = 'white';
+	if (messageCount > 99) {
+		ctx.font = '96px sans-serif';
+		ctx.fillText('99', 64, 96);
+	} else if (messageCount > 9) {
+		ctx.font = '96px sans-serif';
+		ctx.fillText(String(messageCount), 64, 96);
+	} else {
+		ctx.font = '102px sans-serif';
```
sindresorhus: This is too large. It looks weird. Maybe try 90px. Same with the above.

veniversum:
I think badges on task icons are enabled by default (at least in Win10, not sure about previous versions). Should we still document it?

sindresorhus:
@veniversum I installed Windows 10 3 months ago and it was not enabled by default for me, so yes, I think we should. We could just link to a page with instructions. Either of these: https://www.google.com.sg/search?q=enable+taskbar+badge+windows+10&oq=enable+taskbar+badge+windows+10&aqs=chrome..69i57.6423j0j1&sourceid=chrome&ie=UTF-8

browser.js
```diff
@@ -68,6 +68,28 @@ function setVibrancy() {
...
+	ctx.fill();
+	ctx.textAlign = 'center';
+	ctx.fillStyle = 'white';
+	if (messageCount > 99) {
+		ctx.font = '90px sans-serif';
+		ctx.fillText('99', 64, 96);
+	} else {
+		ctx.font = '90px sans-serif';
+		ctx.fillText(String(messageCount), 64, 96);
+	}
```
danhp: I'd replace this if/else block with a ternary to avoid repeated code

```
ctx.font = '90px sans-serif';

const count = messageCount > 99 ? '99' : String(messageCount);
ctx.fillText(count, 64, 96);

Comment (sindresorhus) [inline browser.js]:
```js
ctx.font = '90px sans-serif';
ctx.fillText(String(Math.min(99, messageCount)), 64, 96);
```

sindresorhus:
✨

PR #199, opened 2017-05-06
Fix sending and receiving voice/video calls from within Caprine

### LINKED ISSUE #92: Can't respond to calls
On OS X 10.11.4 (15E65).

When there's a call, a notification appears, and when clicked, a new tab opens in the browser, the call is automatically rejected, and I am offered to start a call with that person.

veniversum:
This should resolve #92. 

Currently, Messenger's calls blocks the window from closing while call is active. I'm not sure if we should override that behavior. Also dark mode doesn't propagate to the call window yet.

### DIFF
--- a/index.js
+++ b/index.js
@@ -201,9 +201,11 @@ app.on('ready', () => {
 
 	webContents.on('new-window', (e, url, frameName, disposition, options) => {
 		e.preventDefault();
-		if (url === 'about:blank') {  // Voice/video call popup
-			options.show = true;
-			e.newGuest = new electron.BrowserWindow(options);
+		if (url === 'about:blank') {
+			if (frameName === 'Video Call') {  // Voice/video call popup
+				options.show = true;
+				e.newGuest = new electron.BrowserWindow(options);
+			}
 		} else {
 			electron.shell.openExternal(url);
 		}
--- a/index.js
+++ b/index.js
@@ -199,9 +199,14 @@ app.on('ready', () => {
 		}
 	});
 
-	webContents.on('new-window', (e, url) => {
+	webContents.on('new-window', (e, url, frameName, disposition, options) => {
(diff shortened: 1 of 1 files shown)


sindresorhus:
Yay! This issue has been bothering me forever. ✨

I tested it today and seems to work fine. The titlebar is not draggable though, because I add the dragging  in the injected CSS. You can just set it to be a normal window with `{titleBarStyle: 'default'}`.

---

I also noticed some errors in the console:

```
https://www.messenger.com/common/preload_info/?trace_id=FOXoUNmFS7m&viz_state=hidden 
Failed to load resource: the server responded with a status of 404 ()

adminTextOutgoingCall.png
Failed to load resource: the server responded with a status of 404 ()
```

Notice the broken image:

screen shot 2017-05-07 at 16 38 05

veniversum:
I'm getting the 404s on the web version of messenger too, so it seems to be an issue on Facebook's end. The image is only found on the `facebook.com` host but not the `messenger.com` one.

We can either:
1. Hide icon for now by injecting CSS
    ```
    ._2pih.img {
    	display: none;
    }
    ```
2. Fix image paths to point to `facebook.com` instead.

<hr>
I'm also getting some other errors in the main console when a call is active, but it doesn't seem to be affecting the functionality of calls.

```
[18012:0506/054036.692:ERROR:audio_send_stream.cc(90)] Failed to set up send codec state.
[18012:0506/054049.791:ERROR:rtpreceiver.cc(61)] AudioRtpReceiver::OnSetVolume: No audio channel exists.
```

sindresorhus:
> I'm getting the 404s on the web version of messenger too, so it seems to be an issue on Facebook's end.

Ah, then we can just ignore it and wait for Facebook to fix it.

sindresorhus:
happy

PR #207, opened 2017-05-15
README - drop unneeded line

ovr:
Hey!

Looks like that this line not needed now

Thank


sindresorhus:
Why? The vibrant theme only works on macOS.

ovr:
Oh, I see, didnt understand that it relates to this theme
Was thinking that project is working only for OSX

PR #211, opened 2017-05-16
Use CSS variables

### LINKED ISSUE #193: Unify all the colors used in darkmode.css using CSS variables
Less duplication.

https://developers.google.com/web/updates/2016/02/css-variables-why-should-you-care

jacklightbody:
Converting dark-mode.css to use css variables. Fixes #193 

Think I got all the colors that appear multiple times, but didn't bother to convert most of the ones that are only used once.


sindresorhus:
Great. Thank you! :)

PR #213, opened 2017-05-18
Reload on system resume to refresh conversations

### LINKED ISSUE #103: App stops updating on resume after suspend mode
Pretty rare case, not sure yet about 1.4 (just updated), but for 1.3 reproduces.

When close MBP, go to another network and open it there some time later (few hours or so) - in some cases app stops getting updates. And can't send messages in this state as well, errors returned.
When app relaunched, messages was trying to send aren't there, but works finely, with fresh history.

I understand it may be not related to the app, as a Facebook's Messenger front-end bug. But maybe it's reasonable to use the [power-monitor](https://github.com/electron/electron/blob/master/docs/api/power-monitor.md) as workaround to reload the web view on `resume`?

veniversum:
Fixes #103.

We'll try to reload the main window when coming out of suspend, but only when `messenger.com` is reachable in order to prevent blank windows. Caprine will keep retrying until window is able to be refreshed.

This fix relies on `power-monitor` module of electron, which itself has some open issues:
https://github.com/electron/electron/issues/1615
https://github.com/electron/electron/issues/8560
These may prevent this fix from working on some systems.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,8 +6,8 @@ const electronLocalShortcut = require('electron-localshortcut');
 const log = require('electron-log');
 const {autoUpdater} = require('electron-updater');
 const isDev = require('electron-is-dev');
-const isReachable = require('is-reachable');
-const promiseRetry = require('promise-retry');
+const isOnline = require('is-online');
+const pRetry = require('p-retry');
 const appMenu = require('./menu');
 const config = require('./config');
 const tray = require('./tray');
@@ -122,17 +122,19 @@ function tryReloadMainWindow() {
 		return;
 	}
 	isReloadingMainWindow = true;
-	promiseRetry(reload, {
+	pRetry(reload, {
 		forever: true,
 		maxTimeout: 300000
-	}).then(isReloadingMainWindow = false).catch();
+	}).then(() => {
+		isReloadingMainWindow = false;
+	}).catch(() => {});
 
-	function reload(retry) {
-		return isReachable('messenger.com').then(reachable => {
-			if (reachable) {
+	function reload() {
+		return isOnline().then(online => {
+			if (online) {
 				mainWindow.reload();
 			} else {
-				retry(new Error('messenger.com not reachable'));
+				throw new Error('no internet connection');
 			}
 		});
 	}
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -6,6 +6,8 @@ const electronLocalShortcut = require('electron-localshortcut');
 const log = require('electron-log');
 const {autoUpdater} = require('electron-updater');
 const isDev = require('electron-is-dev');
+const isReachable = require('is-reachable');
```
sindresorhus: Use [`is-online`](https://github.com/sindresorhus/is-online). `is-reachable` can fail for many other reasons, like firewalls, etc. We only really care about the user being online.

index.js
```diff
@@ -6,6 +6,8 @@ const electronLocalShortcut = require('electron-localshortcut');
...
+const promiseRetry = require('promise-retry');
```
sindresorhus: Can you use [`p-retry`](https://github.com/sindresorhus/p-retry) instead? I prefer how it retries until you tell it not to, instead of only retrying when you tell it to. It supports the same options as `promise-retry`.

index.js
```diff
@@ -113,6 +116,28 @@ function setUpPrivacyBlocking() {
...
+function tryReloadMainWindow() {
+	if (isReloadingMainWindow) {
+		// Already trying to reload!
+		return;
+	}
+	isReloadingMainWindow = true;
+	promiseRetry(reload, {
+		forever: true,
+		maxTimeout: 300000
+	}).then(isReloadingMainWindow = false).catch();
```
sindresorhus: `.then(isReloadingMainWindow = false)` I don't think you meant this. You probably meant:

```js
.then(() => {
  isReloadingMainWindow = false;
})
```

Your way is just setting the value right away.

sindresorhus: And the `.catch()` doesn't actually catch. You need to do `.catch(() => {})`.

veniversum:
There are still some kinks in reloading on resume, and it's the user interaction isn't very smooth. There are side effects such as window gaining focus on reload, and unsent chats being lost on reload.

Closing this until we can find a better solution.

PR #219, opened 2017-06-01
Specify nodejs version to ^6 in package.json

PeterDaveHello:
As the nodejs version defined in .travis.yml, specify the version in package.json


sindresorhus:
Caprine is using Electron, not Node.js directly. This is also the wrong engine. I don't see the point in using `engines` here.

PeterDaveHello:
No need to keep it the same as the CI environment?

sindresorhus:
No

PR #221, opened 2017-06-01
Cache nodejs/npm dependencies in Travis CI

PeterDaveHello:
This helps speed up the CI builds.


sindresorhus:
No thanks. I think this is a bad idea, as we won't be able catch issues with sub-dependencies.

PR #226, opened 2017-06-08
Disable Facebook link tracking

### LINKED ISSUE #224: Block Facebook link tracking
It would be cool if Facebook link tracking was possible to disable. Currently e.g. clicking on a sent https://google.com really opens https://l.messenger.com/l.php?u=https%3A%2F%2Fgoogle.com%2F&h=A_VERY_LONG_TRACKING_ID

kevva:
Fixes #224.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict';
 const path = require('path');
 const fs = require('fs');
-const urlMod = require('url');
+const URL = require('url').URL;
 const electron = require('electron');
 const electronLocalShortcut = require('electron-localshortcut');
 const log = require('electron-log');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -222,6 +223,10 @@ app.on('ready', () => {
 				e.newGuest = new electron.BrowserWindow(options);
 			}
 		} else {
+			if (url.includes('://l.messenger.com/l.php')) {
```
sindresorhus: Can't you use `if (url.startWith('https://l.messenger.com/l.php')) {`? I'm pretty sure Facebook only uses HTTPS.

index.js
```diff
@@ -222,6 +223,10 @@ app.on('ready', () => {
...
+				url = urlMod.parse(url, true).query.u;
```
sindresorhus: Can you use the ne URL API: https://nodejs.org/api/url.html#url_the_whatwg_url_api

sindresorhus:
Can you add this feature to the readme?

index.js
kevva: Good idea. Fixed!

sindresorhus:
Thanks Kevin. As a bonus this also makes the links open much faster, since there's no longer a slow redirect.

kevva:
🎉

PR #227, opened 2017-06-12
Add flashing window on message on Windows

chabou:
Like bouncing dock icon on MacOs, it is a common feature on Windows to make window (and app icon in task bar) flashing.
This PR implements flashing when a new message is received and adds an option to disable it.
Default is true because it is default behaviour of all chat apps on Windows.

### DIFF
--- a/config.js
+++ b/config.js
@@ -12,7 +12,7 @@ module.exports = new Config({
 		},
 		alwaysOnTop: false,
 		bounceDockOnMessage: false,
-		flashWinOnMessage: true,
+		flashWindowOnMessage: true,
 		block: {
 			chatSeen: false,
 			typingIndicator: false
--- a/config.js
+++ b/config.js
@@ -12,6 +12,7 @@ module.exports = new Config({
 		},
 		alwaysOnTop: false,
 		bounceDockOnMessage: false,
+		flashWinOnMessage: true,
 		block: {
 			chatSeen: false,
 			typingIndicator: false
--- a/menu.js
+++ b/menu.js
@@ -372,6 +372,14 @@ const otherTpl = [
 			},
 			{
 				type: 'checkbox',
+				label: 'Flash Window on Message',
+				checked: config.get('flashWinOnMessage'),
+				click(item) {
+					config.set('flashWinOnMessage', item.checked);
+				}
+			},
+			{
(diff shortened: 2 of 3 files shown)


sindresorhus:
What other apps do this? Do they make it configurable?

chabou:
Slack (configurable): https://get.slack.help/hc/en-us/articles/201675007
Skype (not configurable I think)

Discord doesn't but users want it: https://feedback.discordapp.com/forums/326712-discord-dream-land/suggestions/10583004-taskbar-icon-blinking-only-without-notifications
Some users REALLY want it 😆 : 
> "Literally all other applications besides Discord support "window urgency" for this. 
It's what in Windows is known as "blinking task bar entry" (usually not the icon, but the application itself on the task bar - though both are possible). This is the standardized way to show there's something that needs your attention, and operating systems have keyboard shortcuts to jump to the windows that have it set.
Not having this is absurd, especially since it's a single line of code for every single operating system out there. Just implement it, there's no excuse. One line. Seriously."

menu.js
```diff
@@ -372,6 +372,14 @@ const otherTpl = [
 			},
 			{
 				type: 'checkbox',
+				label: 'Flash Window on Message',
+				checked: config.get('flashWinOnMessage'),
```
sindresorhus: Needs:

```js
visible: process.platform === 'win32'
```

So it won't show up on Linux.

config.js
```diff
@@ -12,6 +12,7 @@ module.exports = new Config({
 		},
 		alwaysOnTop: false,
 		bounceDockOnMessage: false,
+		flashWinOnMessage: true,
```
sindresorhus: Make it `flashWindowOnMessage`

sindresorhus:
Note: https://github.com/electron/electron/pull/7654/files#diff-294eccc02d01726fbdeb5f47f2aebc12R328

chabou:
I'm sorry that I didn't see this problem with Linux...

About your note: In the beginning, I turned off flashing when windows have focus. But it is not really required because when focusing, Messenger website update his title and messageCount became equal to 0. But you're right, this is something that Messenger could change. I added this security back.

sindresorhus:
Awesome. Thank you :)

PR #231, opened 2017-06-18
Add menu bar icon on macOS

### LINKED ISSUE #206: Add menu bar icon on macOS
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/caprine/issues/206)

It'd be cool to show the number of unread notifications and add the ability to remove the app from dock. Something like this: 
dock_icon_preview
menu_messenger_icon

The number may be too small in my example, but I think the main idea is there.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $2.00)

- [rororofff rororofff](https://issuehunt.io/u/rororofff) ($2.00)
### Submitted pull Requests
- [#804 Add menu bar mode for macOS](https://issuehunt.io/r/sindresorhus/caprine/pull/804)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/caprine/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
</details>

antonio-ramadas:
Fixes #206. Add menu bar icon and implement a global shortcut to toggle the window. The shortcut is Ctrl+M, where M stands for Messenger.


index.js
```diff
@@ -37,15 +38,19 @@ if (isAlreadyRunning) {
 	app.quit();
 }
 
-function updateBadge(title) {
+function getMessageCount(title) {
 	// ignore `Sindre messaged you` blinking
 	if (title.indexOf('Messenger') === -1) {
-		return;
+		return -1;
```
vadimdemedes: `-1` is usually used for indexes, so I'd return `null` instead.

tray.js
```diff
@@ -37,16 +36,57 @@ exports.create = win => {
...
+		tray.setImage(getDarwinIconPath(messageCount));
+	} else {
+		setBadge(messageCount);
+	}
+
+	updateToolTip(messageCount);
 };
 
-exports.setBadge = shouldDisplayUnread => {
+function setBadge(shouldDisplayUnread) {
```
vadimdemedes: Use arrow functions instead of "regular" ones.

index.js
```diff
@@ -176,7 +181,13 @@ function createMainWindow() {
 
 	win.on('page-title-updated', (e, title) => {
 		e.preventDefault();
-		updateBadge(title);
+
+		const messageCount = getMessageCount(title);
+		if (messageCount === -1) {
```
vadimdemedes: Shouldn't we update the badge anyway? E.g. in case message was just read and the unread count is zero now?

tray.js
```diff
@@ -4,27 +4,26 @@ const electron = require('electron');
...
+	let iconPath = '';
 
-	const toggleWin = () => {
-		if (win.isVisible()) {
-			win.hide();
-		} else {
-			win.show();
-		}
-	};
+	if (process.platform === 'darwin') {
```
vadimdemedes: This could be extracted into a `getIconPath()` function, that'd do platform check inside.

tray.js
vadimdemedes: `shouldDisplayUnread` => `hasUnreadMessages`.

tray.js
```diff
@@ -37,16 +36,57 @@ exports.create = win => {
...
+	updateToolTip(messageCount);
 };
 
-exports.setBadge = shouldDisplayUnread => {
+function setBadge(shouldDisplayUnread) {
 	if (process.platform === 'darwin' || !tray) {
 		return;
 	}
 
 	const icon = shouldDisplayUnread ? 'IconTrayUnread.png' : 'IconTray.png';
```
vadimdemedes: Isn't `getDarwinIconPath()` (potentially `getIconPath()` as mentioned above) responsible for determining which icon to display?

tray.js
```diff
@@ -37,16 +36,57 @@ exports.create = win => {
...
+		return;
+	}
+
+	lastCounter = counter;
+
+	let tip = app.getName();
+
+	if (counter > 0) {
+		const msg = (counter === 1 ? 'message' : 'messages');
+		tip = tip.concat(' - ', counter, ' unread ', msg);
```
vadimdemedes: =>

```js
tip += ` - ${counter} unread {msg}`;
```

tray.js
```diff
@@ -37,16 +36,57 @@ exports.create = win => {
...
+	}
+
+	tray.setToolTip(tip);
+}
+
+function getDarwinIconPath(counter) {
```
vadimdemedes: Since actual number of unread messages (`counter`) isn't needed here, I'd rename it to `hasUnreadMessages` and call `getDarwinIconPath()` like so: `getDarwinIconPath(messageCount > 0)`.

vadimdemedes:
Would be useful to be able to hide menubar icon altogether.

vadimdemedes:
Thanks for working on this, @antonio-ramadas!

index.js
```diff
@@ -54,8 +59,8 @@ function updateBadge(title) {
 		}
 	}
 
-	if (process.platform === 'linux' || process.platform === 'win32') {
-		tray.setBadge(messageCount);
+	if (['linux', 'win32', 'darwin'].indexOf(process.platform) !== -1) {
```
sindresorhus: This check is now moot as we only support those systems.

sindresorhus:
- Clicking the menubar icon many times doesn't always toggle the window. Sometimes it does nothing.
- The white menubar icon should not have a border.
- It should be possible to hide the menubar icon. Should it even be on by default? I'm not sure. Probably best to have a checkbox menuitem in the "Caprine" menu.

index.js
antonio-ramadas: That makes sense. Done!

index.js
antonio-ramadas: Well, no. This condition is for when the window title is blinking. It is unnecessary to update when it is blinking. The window title is different when received a new message. It is only required to update in this last case.

tray.js
antonio-ramadas: You're right. Done!

tray.js
antonio-ramadas: Seems better to me also. Done!

tray.js
antonio-ramadas: The previous refactoring I did now does this. So, done!

tray.js
antonio-ramadas: Complete miss by me. Done!

tray.js
antonio-ramadas: Yes, passing a number may trigger future bugs. Done!

index.js
antonio-ramadas: Yes, did not occur to me at the time. Done!

antonio-ramadas:
> - Clicking the menubar icon many times doesn't always toggle the window. Sometimes it does nothing.

Yes, I experience the same when I was trying to mess around. However, this only happens when the user toggles too quickly. A normal use works as suppose. This is now the responsibility of a npm module, but I can back it up to our own implementation if you want.

> - The white menubar icon should not have a border.

Ok! Will do when I've some more extra time.

> - It should be possible to hide the menubar icon. Should it even be on by default? I'm not sure. Probably best to have a checkbox menuitem in the "Caprine" menu.

By default, I don't think so. I can hide it and have the checkbox. Thinking about it, if the menubar icon is being shown, then the app may be hidden from the dock. With the global shortcut, when the menubar is enabled (and the dock hidden), the window may appear above all others in the current workspace. I've an app which does this and it is greatly convenient. What do you say?

> - When right-clicking the menubar icon it shows a "Toggle" menu item and "Quit. The "Toggle" menuitem is moot on macOS as left-clicking the icon already toggles the window.

I think this is the same for the other platforms. So, it's probably better to take it out of all platforms.

sindresorhus:
> but I can back it up to our own implementation if you want.

👍

sindresorhus:
> Thinking about it, if the menubar icon is being shown, then the app may be hidden from the dock. With the global shortcut, when the menubar is enabled (and the dock hidden), the window may appear above all others in the current workspace. I've an app which does this and it is greatly convenient. What do you say?

Agreed. I think this is the usual behavior for "menubar mode" in GUI apps.

antonio-ramadas:
- Added the double-click event to the menubar icon to be able to toggle more effectively.
- Implemented our own toggle window implementation.
- Wrapped up the tray function into a _continuity mode_ which only works on macOS. This mode registers the global shortcut, places the window on top of everything, fades when lost focus, the dock icon disappears and the menubar icon (tray) appears. Disabling it returns to _normal mode_.

Finally, @sindresorhus I believe I didn't understand what you said about the borders about the white icon. When I removed the thicker black line, it only appeared the lightning bolt with no background. Is this what you want?

antonio-ramadas:
Also, I experienced random crashes when disabling the _continuity mode_. Sometimes the app crashed on the first time, but there were times I had to _play_ for some time switching between modes until I crashed it. I've spent some time tracing it, but I'm not finding the bug. Also, Electron exits with code 0 and shows:
`Received signal 11 SEGV_MAPERR 000000001940`

Any help is appreciated!

sindresorhus:
If I choose "Disable Continuity" while the Caprine window is hidden, it should be shown again.

---

Clicking the Caprine menu bar icon fast stops it from being toggled.

---

> Finally, [@sindresorhus](https://github.com/sindresorhus) I believe I didn't understand what you said about the borders about the white icon. When I removed the thicker black line, it only appeared the lightning bolt with no background. Is this what you want?

Notice the difference in the Caprine icon and another one:

screen shot 2017-07-23 at 15 18 49
screen shot 2017-07-23 at 15 19 27

The black outer border should not be there.

And the lighting icon should be transparent, so it would be blue when highlighted.

// @mirshko

---

Can you add this feature to the https://github.com/sindresorhus/caprine#highlights section and https://github.com/sindresorhus/caprine#features

And add the keyboard shortcut to https://github.com/sindresorhus/caprine#keyboard-shortcuts

---

Why is it named "Continuity"?

mirshko:
@sindresorhus this is just a white version of the unread black logo. Is that what you want?

[unread_1.svg.zip](https://github.com/sindresorhus/caprine/files/1170216/unread_1.svg.zip)

sindresorhus:
The current version (shown above) is using a filled icon with border. It should either be the border without fill or filled without a border, like previewed in https://github.com/sindresorhus/caprine/issues/206#issuecomment-307985319. I think the border one looks the best.

mirshko:
[Archive.zip](https://github.com/sindresorhus/caprine/files/1171107/Archive.zip)

Okay these should be what you need. I think I understand now.

antonio-ramadas:
> If I choose "Disable Continuity" while the Caprine window is hidden, it should be shown again.

Done

> Clicking the Caprine menu bar icon fast stops it from being toggled.

I believe I know what you're talking about, but I'm not figuring out  a solution for it. I created a dummy app just (https://github.com/antonio-ramadas/electron-tray) to test if it was caused by Caprine. I concluded that there is not correlation.

> Menu bar icon

Done

> README.md

Done

> Why is it named "Continuity"?

While working I sometimes need to type quick responses, but I do not like switching context. That's why I'm doing this. This reminds me Apple's [continuity](https://support.apple.com/en-gb/HT204681) as the switch is seamless.

antonio-ramadas:
While working on this, I found a bug. While the continuity is enabled, if the window is maximised, you can't minimize it easily because the menu bar does not show. I can think of 3 solutions while continuity is enabled:

1 - Prevent maximize

2 - Show the app on the dock when the window is maximised (which will enable the menu bar)

3 - Add a button to minimize the window

sindresorhus:
@antonio-ramadas I'm getting an error:

screen shot 2017-09-29 at 13 51 24

antonio-ramadas:
Seems that I forgot to upload the icons. I'll take care of that.

antonio-ramadas:
@sindresorhus Done!

sindresorhus:
> While working I sometimes need to type quick responses, but I do not like switching context. That's why I'm doing this. This reminds me Apple's continuity as the switch is seamless.

I think users will confuse this with the actual system Continuity thing, which is totally different from this. Can we just be explicit an call it `Menu Bar Mode`?

> 1 - Prevent maximize

👍

sindresorhus:
The icon is way to large now. Compare it to other menu bar icons:

screen shot 2017-11-20 at 17 47 36

AllanLRH:
Any progress of this? Would really like the menu-bar functionality, so that I (hopefully) can hide the app from the app-switcher by adding 
```
<key>LSUIElement</key>
<true/>
```
to `/Applications/Caprine.app/Contents/Info.plist`, yet still access the app when needed.

Currently I just start the app when I need to write to someone, or my phone notifies me that someone are writing to me :)

antonio-ramadas:
@sindresorhus I'll take care of that when I've some more time available.

@AllanLRH Any help is welcome. Anyway, according to the latest changes requested, this pull request seems to be close to being accepted.

sindresorhus:
Closing in favor of #374

PR #233, opened 2017-06-19
Work around Electron window focus issue

### LINKED ISSUE #202: Caprine captures F5 key presses while minimized to tray
Hello,
I am using Manjaro GNU/Linux with Gnome 3. This issue has affected me some Caprine 2.0.

When Caprine is minimized to the tray, I can no longer use the f5 key to refresh pages in my browser. I have used the `xev` utility to see whats going on, and I can see while Carpine is minimized to the tray pressing f5 is being capture by another process.

If I bring Caprine back from the tray, or kill Caprine, then I can start using f5 normally again. I'm not sure if its just a coincidence that Caprine is built on web technologies and the f5 key happens to be the one affected.

Is there something I can do to help track down the origin of the issue and identify a fix?

veniversum:
Fixes #202.

Seems to be caused by window keeping focus, so we just have to unfocus before hiding on minimize to tray.

### DIFF
--- a/index.js
+++ b/index.js
@@ -165,7 +165,7 @@ function createMainWindow() {
 	win.on('close', e => {
 		if (!isQuitting) {
 			e.preventDefault();
-
+			win.blur();
 			if (process.platform === 'darwin') {
 				app.hide();
 			} else {
--- a/index.js
+++ b/index.js
@@ -165,7 +165,10 @@ function createMainWindow() {
 	win.on('close', e => {
 		if (!isQuitting) {
 			e.preventDefault();
-			win.blur(); // Workaround for electron/electron#10023
+			
+			// Workaround for electron/electron#10023
+			win.blur();
+			
 			if (process.platform === 'darwin') {
 				app.hide();
 			} else {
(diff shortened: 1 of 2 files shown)


sindresorhus:
If it's actually an issue, it should be fixed in `electron-localshortcut`.

// @parro-it

parro-it:
@veniversum I'm not sure I understood the problem: why you want to avoid shortcuts being blocked?
I looked at the source code you changed, it seems the window is hided anyway, so why you need shortcuts to be enabled?

veniversum:
I think I misunderstood the issue in #202. Currently when minimizing to tray with the exit button, the Caprine window disappears but the previous focused window doesn't regain focus after Caprine is hidden.

sindresorhus:
> Currently when minimizing to tray with the exit button, the Caprine window disappears but the previous focused window doesn't regain focus after Caprine is hidden.

This happens even without `electron-localshortcut`. Seems like an Electron bug. Could you open an Electron issue? If they say it's expected behavior, it should at least be documented that hiding a window does not blur it (which is surprising behavior)

https://electron.atom.io/docs/api/browser-window/#winhide

sindresorhus:
@veniversum No activity on the Electron issue, so we can land the workaround in the meantime. Can you add comment above the workaround code with a link to the Electron issue? And remove the `electron-localshortcut` change?

PR #237, opened 2017-06-22
Exclude toolbar buttons from draggable region

### LINKED ISSUE #236: Disable window dragging via toolbar buttons
It's currently possible to drag the window via the New Message, Voice Call, Video Call & Conversation Information buttons.

This is a very minor issue, but it's sometimes annoying to move the window when clicking these buttons.

jamiestraw:
Fixes #236 🎉

### DIFF
--- a/browser.css
+++ b/browser.css
@@ -25,6 +25,12 @@ body {
 ._5742 {
 	-webkit-app-region: drag;
 }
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thank :)

PR #242, opened 2017-07-01
Add "Don't ask again" option when confirming whether to paste image from the clipboard

ggrelet:
Adding a feature to something that annoyed me: the dialog box asking whether I wanted to paste an image directly from the clipboard.

This use @nathanbuchar's [electron-settings](https://github.com/nathanbuchar/electron-settings).


index.js
```diff
@@ -207,19 +207,34 @@ app.on('ready', () => {
 
 	const argv = require('minimist')(process.argv.slice(1));
 
+	const settings = require('electron-settings');
```
sindresorhus: Should use https://github.com/sindresorhus/caprine/blob/master/config.js

index.js
```diff
@@ -207,19 +207,34 @@ app.on('ready', () => {
...
+					checkboxLabel: 'Never ask me again',
+					checkboxChecked: false
+				}, (resp, checkboxChecked) => {
+					if (resp === 0) {
+						// User selected send
+						webContents.paste();
+						settings.set('neverAskMeAgain.state', checkboxChecked);
+					}
+				});
+			} else if (settings.get('neverAskMeAgain.state')) {
```
sindresorhus: No point in having a `else if` here with just the inverted version of the original `if`. Just use plain `else`.

index.js
```diff
@@ -207,19 +207,34 @@ app.on('ready', () => {
...
-					webContents.paste();
-				}
-			});
+			if (!settings.get('neverAskMeAgain.state')) {
+				electron.dialog.showMessageBox({
+					type: 'info',
+					buttons: ['Send', 'Cancel'],
+					message: 'Are you sure you want to send the image in the clipboard?',
+					icon: electron.clipboard.readImage(),
+					checkboxLabel: 'Never ask me again',
```
sindresorhus: `Never ask me again` => `Don't ask me again`

index.js
```diff
@@ -207,19 +207,34 @@ app.on('ready', () => {
 
 	const argv = require('minimist')(process.argv.slice(1));
 
+	const settings = require('electron-settings');
+
+	if (!settings.has('neverAskMeAgain')){
```
sindresorhus: The setting key needs a better and more relevant name.

sindresorhus:
Don't commit the package-lock.json file. I'm not interested in having that.

index.js
ggrelet: I think both are correct but I changed it as you requested.

sindresorhus: @ggrelet "Don't ask me again" is more common in the system dialog boxes: https://www.google.no/search?q=%22don%27t+ask+me+again%22&source=lnms&tbm=isch&sa=X&ved=0ahUKEwiY2tOqt-jUAhWjQJoKHcZIBWQQ_AUICigB&biw=1440&bih=821

index.js
ggrelet: This was marked as an error when launching `$ np minor --no-publish`

index.js
ggrelet: Got it

index.js
ggrelet: Got it

index.js
ggrelet: You're right

ggrelet:
Thank you 👍

PR #252, opened 2017-07-10
Fix issue with paste confirmation when AlwaysOnTop is activated

chabou:
On Windows, if user has activated AlwaysOnTop option, when pasting an image: confirmation popup  appears behind (and maybe not visible).
Changing this popup as a main window modal fixes this issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -217,7 +217,7 @@ app.on('ready', () => {
 		const clipboardHasImage = electron.clipboard.availableFormats().some(type => type.includes('image'));
 
 		if (clipboardHasImage && config.get('confirmImagePaste')) {
-			electron.dialog.showMessageBox({
+			electron.dialog.showMessageBox(mainWindow, {
 				type: 'info',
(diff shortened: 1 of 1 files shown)


sindresorhus:
That seems like an Electron bug as windows should ever be above modal dialogs. Would you mind opening an issue on Electron?

chabou:
Paste confirmation dialog was a MessageBox but not modal (without any parent).
In OSX, MessageBox are always in front of previous window but not on Windows. But we can see that as a limitation. Maybe open a MessageBox not always on top could make sense if you can declare it modal to counter this behavior.

I really can open an issue on Electron but I'm not sure there is a problem here

sindresorhus:
I see. Seems like that should be an issue on Windows instead then, but good luck with that... I guess it wouldn't hurt to add a note to the Electron docs that `alwaysOnTop` goes even above message boxes on Windows. That's not expected behavior at all.

chabou:
You're right, I'll do that.

PR #256, opened 2017-07-18
Add Preferences keyboard shortcut to readme

aude:
Simply add information about the Preferences menu to the readme. I use a minimal window manager on Linux, so I have no menu bar, and therefore had to browse the source code to find the Preferences keyboard shortcut :)


sindresorhus:
> I use a minimal window manager on Linux, so I have no menu bar,

Just curious, how do you then access the Preferences in other apps? Most Linux apps doesn't have a shortcut for Preferences.

aude:
Thanks for merging. I don't use many apps, I mainly use a web browser and the terminal. Afaik, Cmd/Ctrl+, is a Mac idiom for Preferences, might be why I didn't think of it immediately.

aude:
Oh, I just noticed what happens if you press `Alt`. Hehe.

PR #257, opened 2017-07-18
Add "Table of Contents" in readme.md

PeterDaveHello:


sindresorhus:
Thanks, but I don't see the need for a TOC just yet. The important parts in the readme are the feature highlights and the install instructions, both of which are at the top.

PR #259, opened 2017-07-19
Add insert emoji shortcut

aude:
I like the insert GIF shortcut, I'd like one for insert emoji as well.


kevva:
Looks good and works fine :+1:

PR #270, opened 2017-08-06
Fix conversation action shortcuts (mute, archive, delete)

### LINKED ISSUE #228: Archive menu item/shortcut doesn't work
Hi there,

Caprine's menu item/shortcut for archiving conversations doesn't seem to work.

Also, is there a way to bind the archive shortcut to another key?

Great app btw!

Tim

-

Caprine 2.3.1
Electron 1.6.7
darwin x64 16.6.0

veniversum:
Fixes #228. Messenger changed the layout of the settings menu so selectors had to be updated.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -209,8 +209,7 @@ function openMuteModal() {
 	}
 
 	const selector = '._54nq._2i-c._558b._2n_z li:nth-child(1) a';
-	const nodes = document.querySelectorAll(selector);
-	nodes[nodes.length - 1].click();
+	document.querySelector(selector).click();
 }
 
 function openArchiveModal() {
@@ -219,8 +218,7 @@ function openArchiveModal() {
 	}
 
 	const selector = '._54nq._2i-c._558b._2n_z li:nth-child(3) a';
-	const nodes = document.querySelectorAll(selector);
-	nodes[nodes.length - 1].click();
+	document.querySelector(selector).click();
 }
 
 function openDeleteModal() {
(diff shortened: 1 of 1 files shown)


browser.js
```diff
@@ -222,17 +213,24 @@ function openMuteModal() {
...
 	if (!openConversationMenu()) {
 		return;
 	}
 
-	const selector = `._54nq._2i-c._558b._2n_z ul`;
+	const selector = '._54nq._2i-c._558b._2n_z li:nth-child(3) a';
 	const nodes = document.querySelectorAll(selector);
-	const menuList = nodes[nodes.length - 1];
-	const position = menuList.childNodes.length - 3;
+	nodes[nodes.length - 1].click();
```
veniversum: From testing, the selectors for Mute, Archive & Delete only matches 1 element. In that case we could just do 
```javascript
document.querySelector(selector).click();
```

Are there any cases where the selector will match multiple elements?

daveseco7:
First of all great job, I was going to suggest to clean up the openDeleteModal function, since it was a little bit confusing. 

I only have one concern, the selector `('._5blh._4-0h')[index]` looks for the anchor of the current conversation, so it can open the options menu. Let's suppose, for some reason a user clicks the options menu, as a result the classes you are looking for are no longer available. If I click the conversation option menu and then shortcut "_mute conversation_", **such action will not work**. At first glance, my I suggest using `('._2j6._5l-3 ._3d85'[index].firstChild;` Since it looks for the parent's divs, they will always be there.

browser.js
sindresorhus: Not that I can think of. Do the change.

sindresorhus:
Looks good :)

PR #271, opened 2017-08-07
Add option to turn off badges - fixes #216

### LINKED ISSUE #216: Option to turn off badges?
Hey there! Been using this app for awhile and really like it :-). 

I was wondering if it's possible to turn off the badge count. I found the preference to disable the notifications, but I'd also like to be able to turn off the red badge on OSX. Is that possible?

daveseco7:
Implements what was discussed in #216 . The _Default behaviour_ is incrementing badge counter and it can be changed on the **menu option**.
This feature has only tested in MacOS, but logic for all Operating Systems was added.


daveseco7:
As it stands, the current logic of **disabling badge counter** and **bounce on notification** can be seen as a binary truth table, this means that there's a total of 4 states regarding these 2 menu options.

menu.js
```diff
@@ -138,6 +138,14 @@ const macosTpl = [
 				}
 			},
 			{
+				label: 'Enable notification counter',
```
sindresorhus: > Show Unread Badge

menu.js
```diff
@@ -138,6 +138,14 @@ const macosTpl = [
...
+				type: 'checkbox',
+				checked: config.get('incrementDockOnMessage'),
```
sindresorhus: `incrementDockOnMessage` -> `showUnreadBadge`

PR #275, opened 2017-08-12
Add mute notification option

gerhardberger:
This PR adds a checkable dock option with what the user can mute or unmute the app's notifications, so they don't appear, kind of like with the do not disturb mode on macOS.

Also refactors and fixes the show preferences mechanism, because if the preferences was open and the user opened it again, it stacked.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@ const tray = require('./tray');
 
 const domain = config.get('useWorkChat') ? 'facebook.com' : 'messenger.com';
 
-const {app, ipcMain} = electron;
+const {app, ipcMain, Menu} = electron;
 
 app.setAppUserModelId('com.sindresorhus.caprine');
 app.disableHardwareAcceleration();
@@ -25,6 +25,7 @@ require('electron-context-menu')();
 let mainWindow;
 let isQuitting = false;
 let prevMessageCount = 0;
+let dockMenu;
 
 const isAlreadyRunning = app.makeSingleInstance(() => {
 	if (mainWindow) {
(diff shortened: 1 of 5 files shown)


sindresorhus:
Why put it in the Dock menu?

gerhardberger:
@sindresorhus my use case was that I needed a way to quickly disable the notifications coming from Messenger, while other apps are still able to send notifications. Putting it in the dock allowed to do that without opening the app directly.

sindresorhus:
I'm ok with this change, but there should be an option for it in the App menu too, as the Dock menu is not very discoverable. So [here](https://github.com/sindresorhus/caprine/blob/e245df9edaa5605e1c2fb996f460c31433230d4c/menu.js#L139) and [here](https://github.com/sindresorhus/caprine/blob/e245df9edaa5605e1c2fb996f460c31433230d4c/menu.js#L421).

sindresorhus:
And you need to handle the lint issues: https://travis-ci.org/sindresorhus/caprine/jobs/263910734 :)

sindresorhus:
Ping @gerhardberger :)

gerhardberger:
@sindresorhus sorry for the late response. I fixed the lint errors, put it in the app's main menu and added to the config too.

therealparmesh:
+1

index.js
```diff
@@ -136,6 +137,20 @@ function setUserLocale() {
 	electron.session.defaultSession.cookies.set(cookie, () => {});
 }
 
+function setNotificationsMute(status) {
+	const mainMenuItemIndex = process.platform === 'darwin' ? 3 : 14;
```
sindresorhus: This is too fragile. You should find the position by title instead.

index.js
```diff
@@ -136,6 +137,20 @@ function setUserLocale() {
...
+	const muteMenuItem = Menu.getApplicationMenu().items[0].submenu.items[mainMenuItemIndex];
+
+	console.log(status);
```
sindresorhus: Leftover that should be removed.

sindresorhus:
@gerhardberger Still interested in finishing this?

gerhardberger:
@sindresorhus yes, yes. sorry for being late, but i'm covered with work and college. but i'm finishing next week and then i can resolve them.

gerhardberger:
@sindresorhus fixed them!

sindresorhus:
Looks good now. Thank you and Merry Xmas :)

PR #276, opened 2017-08-14
Added inline titlebar buttons on non-darwin devices to mirror the macos theme

tjhz:


tjhz:
Closing, there are a lot of issues that I didn't account for (i.e. lack of menu bar and colouring on night mode)

PR #284, opened 2017-09-02
Fix calling for groups and other locales

### LINKED ISSUE #280: Group Calling
It seems that voice/video calls work for conversations with individual people. However, they do not seem to work in group conversations.  Nothing seems to happen when I select the people I wish to talk to and click "Call". 
screen shot 2017-08-25 at 4 30 00 pm
When I receive a call from others I cannot join the call as well

---

Caprine 2.6.1
Electron 1.6.7
darwin x64 16.6.0

veniversum:
Fixes #280.

Removed whitelist since it is difficult to support other locales.

### DIFF
--- a/index.js
+++ b/index.js
@@ -262,6 +262,7 @@ app.on('ready', () => {
 			if (frameName !== 'about:blank') {  // Voice/video call popup
 				options.show = true;
 				options.titleBarStyle = 'default';
+				options.webPreferences.sandbox = true;
 				e.newGuest = new electron.BrowserWindow(options);
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


chabou:
I think this PR could fix an issue with Workchat and external SSO.
When your session expire, you need to revalidate credentials with external auth form. Then Workchat tries to popup a new window. For now, I was forced to reload Caprine with `cmd+r` to see this authentication form and renew my SSO session.

⚠️  be sure to sandbox this new BrowserWindow. I don't know if it is already done.

sindresorhus:
Yes, we should turn on the [`sandbox` option](https://electron.atom.io/docs/api/sandbox-option/), as long as video calling still works. Can't imagine it doesn't as the sandbox option just disintegrates the window from Node.js.

sindresorhus:
@veniversum :arrow_up:

Issue #272, opened 2017-08-08
No setting menu on Windows


dudeedud4:
As stated in the title, I see no settings in the Windows version. Sorry if this is a duplicate, I looked but didn't see any other issue.

http://i.imgur.com/qk54kSr.png

daveseco7:
Hi, 
You are correct, by default on Windows the upper toolbar it's not visible, but you can toggle it on by pressing the **_ALT_** key.

This behaviour is also present in all Linux DE

#208

sindresorhus:
It's also documented in the readme.

Issue #157, opened 2017-01-12
Linux settings


MageJohn:
There doesn't seem to be any way to access settings on Linux. The indicator icon in the bar just has the option to toggle the window, and there are no settings buttons visible in the app. I was able to access some settings with the shortcut <ctrl>+<comma>, but that didn't include things like a logout button. Am I missing something, or is there something broken?

sindresorhus:
It's right in the menubar:

<img width="517" alt="screen shot 2017-01-13 at 10 46 31" src="https://cloud.githubusercontent.com/assets/170270/21917850/98c4bfee-d97d-11e6-9cce-6232e27a4ab3.png">

<img width="502" alt="screen shot 2017-01-13 at 10 47 12" src="https://cloud.githubusercontent.com/assets/170270/21917855/ab0e1c36-d97d-11e6-9e4f-3622e27699ae.png">

MageJohn:
Ah, that's the problem. That menu bar seems to be only visible in the Ubuntu Unity interface, where it's been integrated into the global panel at the top. Outside of that interface, though, the menubar is not visible. See for example Gnome 3, what I'm using currently: 

![Screenshot 1](https://cloud.githubusercontent.com/assets/3520558/21923669/52013f2a-d96d-11e6-8fb2-99284b34f71e.png)
![Screenshot 2](https://cloud.githubusercontent.com/assets/3520558/21923668/51fd57c0-d96d-11e6-835f-63b437746855.png)

So, am I still missing something?

sindresorhus:
I don't know. That would be something to ask on the [Electron issue tracker](https://github.com/electron/electron/issues). I'm just using their API to set the menu.

airtonix:
@MageJohn what's the keyboard shortcut for opening the settings? i think you missed that from your initial message.

AydenMurray97:
@MageJohn have you tried using pressing Alt? I'm using Xubuntu which also didn't show the menu bar by default but using Alt toggles it

MageJohn:
Hey guys, sorry I've been quiet. 

@airtonix Sorry, the shortcut is there, I just didn't make it very clear. It's <Ctrl+Comma>. 

@AydenMurray97 That works! That's great, thanks. Maybe this should be documented somewhere?

Issue #115, opened 2016-08-07
Menu bar is hidden on Windows


c99koder:
There's no way to access the settings on Windows because the menu bar is hidden and you've hidden the gear button that's normally part of the web UI.  Pressing the alt key temporarily shows the menu, but that's super confusing for most users.

sindresorhus:
> Pressing the alt key temporarily shows the menu, but that's super confusing for most users.

That how Windows works.

c99koder:
Huh, I've never seen a Windows app hide the menu bar like that. Doing some quick Googling, I guess ,the Windows 7 version of Explorer had an option to do that?  Definitely had me stumped for quite a few minutes trying to turn off the dark theme and turn on notifications.  Either way, there's no reason to be hiding the gear icon on Windows.

workboot:
>> Pressing the alt key temporarily shows the menu, but that's super confusing for most users.

> That how Windows works.

Well do I misunderstand your answer? This is not standard behaviour for electron apps and not specific to Windows is it? The same thing happens on Fedora 23 and surely because of

``` javascript
autoHideMenuBar: true
```
in [index.js](https://github.com/sindresorhus/caprine/blob/2415e8d93bb505e5fd0b901f4e4d0ee3df1e7a33/index.js#L65). So is this a design choice for caprine?
In that case I agree with @c99koder that it is confusing for most users. At least a menu entry in the tray icon should be available to make it visible.

Issue #113, opened 2016-07-20
Preferences in Gnome 3


davidcortesortuno:
I can't find the `Preferences` option in Gnome 3:

![screenshot from 2016-07-20 21-41-38](https://cloud.githubusercontent.com/assets/12200541/17002423/1d1104d8-4ec3-11e6-87af-c8e3ef4dc3af.png)

Is it in  different place? The tray icon only shows me the `Toggle` option.

sindresorhus:
Preferences (and all other menu items) are in the app menu, not in the tray menu.

forivall:
Press the "alt" key for the menu to appear inside the app's window. Otherwise, electron needs to support the gnome 3 topbar application menu.

PR #302, opened 2017-10-08
Show menu bar by default on Linux and Windows and add option to disable that

klauscfhq:
A minor feature that adds ability to toggle the menu in a persistently visible mode.
Once activated makes the menu bar visible and disables the <kbd>ALT</kbd> key toggling, while the opposite happens on deactivation.

- Available on Linux/Windows only 
- Preserves last state on relaunch
- Preserves **current default state** where menu bar initializes hidden

 @sindresorhus feel free to suggest edits : )

caprine


sindresorhus:
Why?

klauscfhq:
It mainly targets users that are not much confortable with keyboard shortcuts and prefer using the menu options instead. Issues  #115, #113, #272 and #157 are to a degree relatable.
In addition, [Ubuntu dropped Unity for Gnome](https://arstechnica.com/information-technology/2017/04/ubuntu-unity-is-dead-desktop-will-switch-back-to-gnome-next-year/), that has an identical window structure to Windows, thus lots of ex-Unity users who were used in having their menu bar constantly visible by default, will be now pressing the <kbd>ALT</kbd> key to access it momentarily. Also, this can be considered out of context, but Atom is a popular app among Linux/Windows users that also [features the same menu bar behavior](https://github.com/atom/atom/pull/4215). 

I completely get your view, to not unreasonably increase Caprine's sustainability cost, the PR could be frozen until there is enough interest by people who might actuall use it : )

sindresorhus:
I've been thinking about this and I'm gonna accept, but it should be done like in Atom, where the menu bar is visible by default and you can opt into auto-hiding it. The option should be named `Auto Hide Menu Bar` (like in Atom).

klauscfhq:
All good. Let me know if there are any changes you may want to make : )

browser.js
```diff
@@ -118,7 +118,7 @@ ipc.on('toggle-vibrancy', () => {
 });
 
 ipc.on('toggle-menu-bar', () => {
```
sindresorhus: The title here needs to be updated too.

config.js
```diff
@@ -5,6 +5,7 @@ module.exports = new Store({
 	defaults: {
 		darkMode: false,
 		vibrancy: false,
+		autoHideMenuBar: false,
```
sindresorhus: Should be added to the bottom.

menu.js
```diff
@@ -49,6 +49,15 @@ const viewSubmenu = [
 			sendAction('toggle-sidebar');
 		}
 	},
+	{
+		label: 'Auto Hide Menu Bar',
+		type: 'checkbox',
+		checked: config.get('autoHideMenuBar'),
+		visible: !(process.platform === 'darwin'),
```
sindresorhus: Should be `process.platform !== 'darwin'`

menu.js
```diff
@@ -49,6 +49,15 @@ const viewSubmenu = [
...
+		click() {
+			sendAction('toggle-menu-bar');
```
sindresorhus: Don't really need IPC here as it's run in the same context. Just put the toggle in a function in this file and call it here as a normal function call.

browser.js
```diff
@@ -239,6 +248,9 @@ document.addEventListener('DOMContentLoaded', () => {
 	document.body.appendChild(style);
 	setZoom(zoomFactor);
 
+	// Activate the menu bar if it was set before quitting
+	setMenuBar();
+
```
sindresorhus: None of the changes in `browser.js` needs to be here. It can all me in `menu.js`.

index.js
```diff
@@ -287,6 +287,20 @@ ipcMain.on('set-vibrancy', () => {
...
+		// Hide the menu bar
+		mainWindow.setMenuBarVisibility(false);
+		// Enable ALT key toggling
+		mainWindow.setAutoHideMenuBar(true);
+	} else {
+		// Make the menu bar persistently visible
+		mainWindow.setMenuBarVisibility(true);
+		// Disable ALT key toggling
+		mainWindow.setAutoHideMenuBar(false);
+	}
```
sindresorhus: This can be simplified to:

```js
const autoHideMenuBar = config.get('autoHideMenuBar');
mainWindow.setAutoHideMenuBar(autoHideMenuBar);
mainWindow.setMenuBarVisibility(!autoHideMenuBar);
```

klauscfhq: Nice thinking! Maybe we could simplify it even more, just by setting the `autoHideMenuBar` attribute of `electron.BrowserWindow()` to: 

- `autoHideMenuBar: config.get('autoHideMenuBar')`

And perhaps do directly the following on `menu.js`:

```js
label: 'Auto Hide Menu Bar',
type: 'checkbox',
checked: config.get('autoHideMenuBar'),
visible: process.platform !== 'darwin',
click(item, focusedWindow) {
    config.set('autoHideMenuBar', item.checked);
    focusedWindow.setAutoHideMenuBar(item.checked);
    focusedWindow.setMenuBarVisibility(!item.checked); 
}
```

Let me know if you would care for any of these changes, or prefer to go with the initial ones : )

sindresorhus: > autoHideMenuBar: config.get('autoHideMenuBar')

But you still have to do 	`.setMenuBarVisibility`, right? Because there's no BrowserWindow option for that.

klauscfhq: It is not that well documented, but both on Linux and Windows, if `autoHideMenuBar` is set to `false`, the app launches with its menu bar visible and has the `ALT` key toggling disabled. On the contrary, if it is set to `true`, the app launches with its menu bar hidden and the `ALT` key toggling enabled. Here is a quick demo;

demo

Unfortunately, the only available [piece of documentation](https://electronjs.org/docs/api/browser-window#new-browserwindowoptions) does not describe this default behavior thoroughly. But I agree with you, it would be better to use the `.setMenuBarVisibility()` & `.setAutoHideMenuBar()` methods, as they provide a more reliable and clearer way to deliver the feature.

sindresorhus: Seeing as the default is what we want, I'm fine with going with your proposal.

klauscfhq: Sending it over now : )

sindresorhus:
I think it should be placed here: https://github.com/sindresorhus/caprine/blob/100a437ed8ab294384d200d49b303a387aaf51b0/menu.js#L461 Together with the other persistent options. You can then also remove the `visible` check as it's a Linux/Windows-only menu.

sindresorhus:
And you need to fix the merge conflict.

klauscfhq:
Good to go : )

sindresorhus:
Thank you! :)

klauscfhq:
No problem at all! Happy holidays : )

PR #304, opened 2017-10-09
Changed color when blocked

dusansimic:
This commit changes color of the message box text to light when the
conversation is blocked.


sindresorhus:
You need to add a comment for the selectors in the file so it's maintainable, like done with the other selectors there.

PR #306, opened 2017-10-12
Fix event bar colors in dark mode

dusansimic:
This commit fixes colors for event bar in dark mode.


sindresorhus:
You need to add a comment for the selectors in the file so it's maintainable, like done with the other selectors there.

PR #311, opened 2017-10-17
Remove redundant background containing emoji

### LINKED ISSUE #310: Single emoticon has redundant background
You are probably aware of it, but I decided to report it anyways. Some of the emoticons still have message-like background:

And some of them do not:

I believe the second case is correct.

I am using Caprine 2.7.1 on macOS with dark theme and vibrancy turned on.

Great app, really like it.

sjoloughlin:
Fixes #310


sindresorhus:
Thanks :)

PR #319, opened 2017-11-03
Don't use mainWindow.minimize() to run Caprine without startup window

### LINKED ISSUE #318: Launch Caprine without startup window (run in background)
```sh
caprine --no-startup-window
```

KSXGitHub:
There's `--minimize` flag which should stop startup window from appearing but
it uses `mainWindow.minimize()` which causes [undesired effect](https://goo.gl/629GdB) in Linux

On the other hand, `--no-startup-window` stop startup window from appearing by
using `mainWindow.hide()` which doesn't produce any side-effect

[Demo video](https://goo.gl/629GdB)

Fixes #318 

### Alternate solution

Instead of adding `--no-startup-window`, just edit [this line](https://github.com/KSXGitHub/caprine/blob/flag-minimize/index.js#L257)

[Branch](https://github.com/KSXGitHub/caprine/tree/flag-minimize)

### DIFF
--- a/index.js
+++ b/index.js
@@ -253,7 +253,9 @@ app.on('ready', () => {
 			webContents.insertCSS(fs.readFileSync(path.join(__dirname, 'workchat.css'), 'utf8'));
 		}
 
-		if (argv.minimize) {
+		if (argv['startup-window'] === false) {
+			mainWindow.hide();
+		} else if (argv.minimize) {
 			mainWindow.minimize();
 		} else {
 			mainWindow.show();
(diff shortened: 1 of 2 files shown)


KSXGitHub:
Few questions:

1. Should `--minimize` be deprecated? Or should we just re-implement `--minimize` with `mainWindow.hide()` instead of adding a new flag (`--no-startup-window`)?

2. Convert entire code to [standard](https://standardjs.com/) code-style? (Because their tools and the ecosystem around it are awesome)

sindresorhus:
There's now a "Launch Minimized" option for Linux/Window: https://github.com/sindresorhus/caprine/blob/9aa6532fb37ad4ce441e9b4113e00fbaa904d6fe/menu.js#L487

macOS will get a menu bar mode: #374

PR #320, opened 2017-11-05
Update electron version

dusansimic:
I'm suggesting to update electron version since it has fixes in chromium
(version 60/61) for tray icon. With electron version 1.8.1 tray icons is
displayed in Ubutnu 17.10 (Gnome).

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
 		"minimist": "^1.2.0"
 	},
 	"devDependencies": {
-		"electron": "1.6.11",
+		"electron": "1.8.1",
 		"electron-builder": "19.42.2",
 		"xo": "*"
 	},


sindresorhus:
I wish I could upgrade, but Electron is in a weird state now. 1.6 is really the only stable release. 1.7 has some brokenness, and 1.8 is just totally broken and even officially marked as a prerelease. I'm holding it out for 1.8 for a bit longer until they can fix the worst regressions. Caprine works pretty well right now and I don't have time to implement lots of workaround for Electron bugs.

I'll keep this PR open for discovery and merge it when Electron 1.8 is stable enough.

dusansimic:
Electron has been updated so I will close this PR.

PR #321, opened 2017-11-05
Persist "Toggle Sidebar" setting after quitting

jdanford:


jdanford:
@sindresorhus any idea what's wrong with the AppVeyor build? Not sure about you, but I'm just getting this message:
> Project not found or access denied.

sindresorhus:
Just ignore the AppVeyor failure. I've been trying to remove AppVeyor without success.

sindresorhus:
Looks pretty good, but you need to add a default value for the config here: https://github.com/sindresorhus/caprine/blob/master/config.js Should be `true` by default.

jdanford:
Default value has been added!

browser.js
```diff
@@ -82,8 +77,6 @@ function setDarkMode() {
 function setVibrancy() {
 	document.documentElement.classList.toggle('vibrancy', config.get('vibrancy'));
 	ipc.send('set-vibrancy');
-
-	document.documentElement.style.backgroundColor = 'transparent';
```
sindresorhus: I just noticed this and it breaks vibrancy mode. Please don't do unrelated changes.

sindresorhus: I don't remember, but I think it's so it set right away instead of waiting for `vibrancy.css` to be loaded.

jdanford: If you enabled vibrancy and then disabled it, i.e. called `setVibrancy()` twice, wouldn't the background color still be set to transparent?

sindresorhus: I dunno. I don't have time to look into it. It's not really relevant to this regardless. Happy to have it fixed/improved in a separate PR though.

jdanford: Alright, I removed the unrelated change, so this PR should be good to merge.

sindresorhus:
Thanks :)

PR #330, opened 2017-12-02
Call immediately when starting a audio/video call

vadimdemedes:
Removes this pointless step and calls the person right away:

screen shot 2017-12-02 at 7 12 56 pm

I think this step isn't needed, because I already expressed my intent to call that person by clicking on the button.

This PR is extracted from the Touch Bar work.

### DIFF
--- a/browser-call.js
+++ b/browser-call.js
@@ -1,3 +1,6 @@
+'use strict';
+const elementReady = require('element-ready');
+
 (async () => {
 	const startCallButton = await elementReady('._3quh._30yy._2t_');
 	startCallButton.click();
--- a/browser-call.js
+++ b/browser-call.js
@@ -1,23 +1,4 @@
-'use strict';
-const electron = require('electron');
-
-const {ipcRenderer: ipc} = electron;
-
-const onReady = fn => {
-	window.addEventListener('load', () => {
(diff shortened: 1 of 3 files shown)


browser-call.js
```diff
@@ -0,0 +1,23 @@
...
+		};
+
+		check();
+	});
+}
+
+onReady(() => {
+	// Start call
+	document.querySelector('._3quh._30yy._2t_').click();
+});
```
sindresorhus: Use https://github.com/sindresorhus/element-ready here:

```js
(async () => {
	const startCallButtonSelector = '._3quh._30yy._2t_';
	await elementReady(startCallButtonSelector);
	document.querySelector(startCallButtonSelector).click();
});
```

browser-call.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+const electron = require('electron');
+
+const {ipcRenderer: ipc} = electron;
```
sindresorhus: This is unused.

browser-call.js
vadimdemedes: Good tip!

vadimdemedes:
Damn, I made ridiculous errors, thanks for fixing!

Issue #350, opened 2017-12-28
Compact mode doesn't work with the right sidebar


Rezonansowy:
When you resize the Caprine window to a smaller size, it should present appearance as follows:
![](https://github.com/sindresorhus/caprine/raw/master/media/screenshot-compact.png)

The problem is that the right sidebar is still showing, which breaks the layout making the whole GUI hard to use.

Here's the current problem screenshot:
![sidebar layout problem](https://user-images.githubusercontent.com/6929363/34421734-0809cb0a-ebdf-11e7-9ff2-58b9813ab197.png)

---

Caprine 2.9.0
Electron 1.7.9
darwin x64 16.7.0

sindresorhus:
Yes, it should automatically hide the right-sidebar when the window is too small to show it.

PR #354, opened 2017-12-31
Fix opened right sidebar in compact mode

dusansimic:
The right-sidebar was kept opened in compact mode and it was difficult
to read messages. This fixes it and hides right-sidebar in compact mode.

Closes: https://github.com/sindresorhus/caprine/issues/350


sindresorhus:
Thanks :)

PR #357, opened 2018-01-02
WIP: Add upload shortcut

aude:
**This does not work now.**

Focusing a `<input type="file">` via JavaScript seems to not work.
I think it does not work because of browser policy. https://stackoverflow.com/q/210643

Do you know anything that can be done to make it work? Maybe allow it somehow?

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -48,6 +48,10 @@ ipc.on('insert-emoji', () => {
 	document.querySelector('._5s2p').click();
 });
 
+ipc.on('insert-files', () => {
+	document.querySelector('._260t').click();
+});
+
 ipc.on('next-conversation', nextConversation);
 
 ipc.on('previous-conversation', previousConversation);
--- a/menu.js
+++ b/menu.js
@@ -327,6 +327,13 @@ const macosTpl = [
 				}
 			},
 			{
+				label: 'Insert Files',
+				accelerator: 'Cmd+U',
+				click() {
+					sendAction('insert-files');
+				}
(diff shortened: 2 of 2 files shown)


sindresorhus:
You could probably use https://electronjs.org/docs/api/web-contents#contentssendinputeventevent

aude:
I'm closing this pull request, for two reasons:

* `Ctrl+U` does not seem to work when the message text input field is focused, which is when I need it the most.
* I tried [webContents.sendInputEvent()](https://electronjs.org/docs/api/web-contents#contentssendinputeventevent) like you suggested, and it turned out really hacky. Something like finding the element, focusing it, then sending an `Enter` key press.

I could probably find a better solution to the last point if I wanted, but the first point is really stopping me.

sindresorhus:
> Ctrl+U does not seem to work when the message text input field is focused, which is when I need it the most.

Did you try any other key combination? Maybe Ctrl+U is overriden and used by the system for something else.

aude:
Yeah, I tried `Ctrl+T`, and it worked. But I want `Ctrl+U`, that's the only one I could think of that makes sense. So I lost the motivation to continue.

sindresorhus:
`Ctrl+Shift+U` could have been an alternative.

aude:
That's a good idea, thanks. I'll give it a try.

aude:
Nah, it didn't work either. I bet `Ctrl+U` is used for something internally.

PR #358, opened 2018-01-02
Add search shortcut

aude:
This adds a keyboard shortcut to search in previous messages.

I think it makes sense to let this shortcut be Ctrl+F. Thus I changed "Find Conversation" from Ctrl+F to Ctrl+K.


sindresorhus:
Makes sense. Thanks :)

PR #364, opened 2018-01-10
Upgrade `electron-builder` and `electron-updater`

duongdev:


sindresorhus:
Thanks, but I only update before a release. By the time I do a new release, there's another update.

PR #372, opened 2018-01-19
Add `Show Message Buttons` option

### LINKED ISSUE #355: Optionally hide feature buttons (pictures, emojis, gifs, etc)
The Messenger Lite App removes these icons along the bottom of the app. It would be great to have the ability to optionally remove them. Or add a mode that replaces all of them with one "send" icon.

screen shot 2018-01-01 at 12 17 17 pm

dusansimic:
This commit suggests to hide options in message box (insert emoji,
image, git...) and show them with hover over like/send button. This
would make interface clean without removing those options.

Fixes #355

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -92,12 +92,9 @@ ipc.on('toggle-mute-notifications', (event, defaultStatus) => {
 	}
 });
 
-ipc.on('toggle-message-buttons', () => {
-	elementReady('._39bj').then(messageButtons => {
-		messageButtons.style.display = config.get('showMessageButtons') ? 'flex' : 'none';
-	}).catch(err => {
-		console.error(err);
-	});
+ipc.on('toggle-message-buttons', async () => {
+	const messageButtons = await elementReady('._39bj');
+	messageButtons.style.display = config.get('showMessageButtons') ? 'flex' : 'none';
 });
 
 function setDarkMode() {
--- a/browser.js
+++ b/browser.js
@@ -1,5 +1,6 @@
 'use strict';
 const electron = require('electron');
+const elementReady = require('element-ready');
 const config = require('./config');
 
 const {ipcRenderer: ipc} = electron;
@@ -92,8 +93,11 @@ ipc.on('toggle-mute-notifications', (event, defaultStatus) => {
 });
 
 ipc.on('toggle-message-buttons', () => {
-	const messageButtons = document.querySelector('._39bj');
-	messageButtons.style.display = config.get('showMessageButtons') ? 'flex' : 'none';
+	elementReady('._39bj').then(messageButtons => {
+		messageButtons.style.display = config.get('showMessageButtons') ? 'flex' : 'none';
+	}).catch(err => {
+		console.error(err);
+	});
 });
 
 function setDarkMode() {
--- a/browser.js
+++ b/browser.js
@@ -91,6 +91,22 @@ ipc.on('toggle-mute-notifications', (event, defaultStatus) => {
 	}
 });
 
+ipc.on('toggle-message-buttons', () => {
+	const messageButtons = document.querySelector('._39bj');
+
+	let messageButtonsStatus = window.getComputedStyle(messageButtons, null).getPropertyValue('display');
+
+	if (config.get('showMessageButtons')) {
+		messageButtons.style.display = 'flex';
+	} else {
+		messageButtons.style.display = 'none';
+	}
+
+	messageButtonsStatus = window.getComputedStyle(messageButtons, null).getPropertyValue('display');
+
+	ipc.send('message-buttons-toggled', messageButtonsStatus === 'flex');
+});
+
 function setDarkMode() {
 	document.documentElement.classList.toggle('dark-mode', config.get('darkMode'));
 	ipc.send('set-vibrancy');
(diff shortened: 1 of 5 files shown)


sindresorhus:
I'm not sure if it's clear enough that the user has to hover the 👍 to see more. Maybe we need a separate icon to indicate more items. I'm not really sure. I just have a feeling we'll get complaints about this.

dusansimic:
Yea, you're right. I'll work on that.

sindresorhus:
When clicking "Show Message Buttons", I get an error:

screen shot 2018-03-09 at 14 32 01

menu.js
```diff
@@ -55,6 +55,15 @@ const viewSubmenu = [
...
 		}
+	},
+	{
+		label: 'Show Message Buttons',
+		type: 'checkbox',
+		checked: config.get('showMessageButtons'),
+		click() {
+			config.set('showMessageButtons', !config.get('showMessageButtons'));
+			sendAction('toggle-message-buttons');
+		}
```
sindresorhus: Move this below `Toggle Sidebar`.

index.js
```diff
@@ -321,6 +332,10 @@ ipcMain.on('mute-notifications-toggled', (event, status) => {
 	setNotificationsMute(status);
 });
 
+ipcMain.on('message-buttons-toggled', (event, status) => {
+	setMessageButtonsVisible(status);
```
sindresorhus: I don't see the point of moving everything here into a separate function. The logic can just be here.

dusansimic: Done.

sindresorhus:
I'm still getting the error from https://github.com/sindresorhus/caprine/pull/372#issuecomment-371736010

dusansimic:
I think this will now fix that error (https://github.com/sindresorhus/caprine/pull/372#issuecomment-371736010) but I'm not 100% sure since I can't test it on real Mac.

sindresorhus:
Why are you putting it in the Dock-menu? It's also not a Mac-specific issue. So I'm lost on what you're actually talking about.

browser.js
```diff
@@ -91,6 +91,22 @@ ipc.on('toggle-mute-notifications', (event, defaultStatus) => {
...
+
+	if (config.get('showMessageButtons')) {
+		messageButtons.style.display = 'flex';
+	} else {
+		messageButtons.style.display = 'none';
+	}
+
+	messageButtonsStatus = window.getComputedStyle(messageButtons, null).getPropertyValue('display');
+
+	ipc.send('message-buttons-toggled', messageButtonsStatus === 'flex');
```
sindresorhus: You don't need to use IPC for this. You can use the config from the renderer process too.

browser.js
```diff
@@ -91,6 +91,22 @@ ipc.on('toggle-mute-notifications', (event, defaultStatus) => {
...
+
+	let messageButtonsStatus = window.getComputedStyle(messageButtons, null).getPropertyValue('display');
+
+	if (config.get('showMessageButtons')) {
+		messageButtons.style.display = 'flex';
+	} else {
+		messageButtons.style.display = 'none';
+	}
+
+	messageButtonsStatus = window.getComputedStyle(messageButtons, null).getPropertyValue('display');
```
sindresorhus: Why are you measuring here and on line 97? The config should be the source of truth.

dusansimic:
There, I think this should be it. I'm sorry for this mess. I'm still kind of new to electron and learning how it works. The thing about that error You were getting is that checkbox for show/hiding message buttons wasn't added to the dock and method inside `index.js` was trying to change it's state. I've completely removed that method now and everything is in `browser.js` event and `menu.js` checkbox that will trigger that event. Again sorry for that spaghetti code that i wrote.

sindresorhus:
No worries. I was just confused about what we were talking about.

sindresorhus:
You still need to handle the case where "Show Message Buttons" is unchecked on startup, and hide the icons then.

dusansimic:
I've just found a very big problem. When Caprine starts, in my os (Ubuntu 17.10), it takes some time to load all elements (you can see that document is empty for some time). The problem is that `dom-ready` event in `webComponents` is called before all elements are loaded for some reason so every config we want to set that involves manipulating dom is done before you can event access settings for example. My proposal is to add a delay (`setTimeout` for example) and do it after some period of time but that is not a very good solution since on some other machine it might take longer to load elements. Any ideas?

sindresorhus:
You can use https://github.com/sindresorhus/element-ready

browser.js
```diff
@@ -91,6 +92,14 @@ ipc.on('toggle-mute-notifications', (event, defaultStatus) => {
 	}
 });
 
+ipc.on('toggle-message-buttons', () => {
+	elementReady('._39bj').then(messageButtons => {
```
sindresorhus: Use async/await

browser.js
```diff
@@ -91,6 +92,14 @@ ipc.on('toggle-mute-notifications', (event, defaultStatus) => {
...
+		messageButtons.style.display = config.get('showMessageButtons') ? 'flex' : 'none';
+	}).catch(err => {
```
sindresorhus: No need to catch this. We have global rejection handlers installed.

PR #394, opened 2018-02-25
Add messenger menu options to the "View" menu

### LINKED ISSUE #352: Missing Messenger menu in Caprine
A menu which allows user view Archived Threads, Unread Threads, Active Contacts and Connection Requests is missing. It looks like this on Messenger website:
missing menu

---

Caprine 2.9.0
Electron 1.7.9
darwin x64 16.7.0

dusansimic:
I've added options from messenger settings menu to view menu as
suggested.

Fixes #352


Jay1:
This was needed, thanks a lot

browser.js
```diff
@@ -277,6 +293,38 @@ function closePreferences() {
 	doneButton.click();
 }
 
+function openActiveContactsView() {
+	// Create the menu for the below
+	document.querySelector('._30yy._2fug._p').click();
+
+	const nodes = document.querySelectorAll('._54nq._2i-c._558b._2n_z li:nth-child(3) a');
+	nodes[nodes.length - 1].click();
+}
```
sindresorhus: No need for external functions for these. Just put the logic in the event handler.

browser.js
```diff
@@ -277,6 +293,38 @@ function closePreferences() {
 	doneButton.click();
 }
 
+function openActiveContactsView() {
+	// Create the menu for the below
+	document.querySelector('._30yy._2fug._p').click();
```
sindresorhus: This is used in many functions. Make it a reusable function in case we need to update the selector at some point.

menu.js
```diff
@@ -55,6 +55,33 @@ const viewSubmenu = [
 		click() {
 			sendAction('toggle-dark-mode');
 		}
+	},
+	{
+		type: 'separator'
+	},
+	{
+		label: 'Open Active Contacts',
```
sindresorhus: I would use `Show` instead of `Open` for these.

menu.js
```diff
@@ -55,6 +55,33 @@ const viewSubmenu = [
...
+		}
+	},
+	{
+		label: 'Open Archived Threads',
+		click() {
+			sendAction('open-archived-threads-view');
+		}
+	},
+	{
+		label: 'Toggle Unread Threads',
```
sindresorhus: `Show` here too

menu.js
```diff
@@ -55,6 +55,33 @@ const viewSubmenu = [
...
+		click() {
+			sendAction('toggle-unread-threads-view');
+		}
```
sindresorhus: Needs a separator here.

sindresorhus:
If I click, for example, "Open Active Contacts", I can not then click "Open Message Requests". I think it should always go back to the main before going to one of these.

menu.js
dusansimic: I would suggest to keep `Toggle` on this label since it toggles the view between read/unread threads and it doesn't show it. Maybe we could change label text to show/hide depending on the current view if nothing else. :smile:

menu.js
sindresorhus: Not on macOS

browser.js
```diff
@@ -8,6 +8,20 @@ const listSelector = 'div[role="navigation"] > div > ul';
...
+}
+
+function selectMenuItem(itemNumber) {
+	const selector = document.querySelector(`._54nq._2i-c._558b._2n_z li:nth-child(${itemNumber}) a`);
+	selector.click();
+}
+
+function clickBackButton() {
+	const backButton = document.querySelector('._30yy._2oc9');
+	backButton ? backButton.click() : null; // eslint-disable-line no-unused-expressions
```
sindresorhus: This should be an if-statement

sindresorhus:
Looks good, but the "Toggle Vibrancy" menu item should be together with the other Toggle menu items on macOS. This is how it currently looks:

screen shot 2018-04-18 at 14 24 10

dusansimic:
Done.

menu.js
```diff
@@ -123,10 +123,7 @@ ${process.platform} ${process.arch} ${os.release()}`;
 ];
 
 if (process.platform === 'darwin') {
-	viewSubmenu.push({
-		type: 'separator'
-	},
-	{
+	viewSubmenu.splice(6, 0, {
```
sindresorhus: Using numeric splice here is very fragile and makes it hard to do changes. Check out this instead for how to place a menu item in a specific position: https://electronjs.org/docs/api/menu#menu-item-position

dusansimic: Fixed it, thanks for that note. :+1:

sindresorhus:
Looks good. Can you fix the merge conflict?

dusansimic:
If you mean all other files and some new changes inside `browser.js`, those are from rebasing branch to master.

PR #399, opened 2018-03-04
Fix title bar transparency on MacOS High Sierra

### LINKED ISSUE #295: Transparent titlebar in macOS High Sierra when using vibrancy
screen shot 2017-09-28 at 09 26 43

Bhullnatik:
(Temporary) fixes #295 

Not sure if that's the right solution, but it works so I figured I'd open a PR to start a discussion.
There is still an issue, see https://github.com/electron/electron/issues/10886.

### DIFF
--- a/index.js
+++ b/index.js
@@ -174,8 +174,9 @@ function createMainWindow() {
 		minWidth: 400,
 		minHeight: 200,
 		alwaysOnTop: config.get('alwaysOnTop'),
+		titleBarStyle: 'hidden-inset',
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think it's better to have the default frame than square corners. I also tested this PR with Electron 2.0.0-beta.1 and the toolbar has no background, so it's broken again there.

sindresorhus:
Vibrancy mode is now disabled until it's fixed in Electron: https://github.com/sindresorhus/caprine/releases/tag/v2.13.0

PR #419, opened 2018-03-25
Fix keyboard shortcut for archiving does not work for group conversations

### LINKED ISSUE #391: Keyboard shortcut for archiving does not work for group conversations
Using the keyboard shortcut to archive chats (Shift+Cmd+A) does not work for group chats. It triggers the button to 'Leave Conversation' instead and pops up a confirmation alert.

Caused by group chats having the button for 'Leave Conversation' as the third child element.

Solution: Setup a branch to select the fourth child element if conversation is a group chat.

---

Caprine 2.9.0
Electron 1.7.9
darwin x64 17.4.0

daveseco7:
Fixed #391
It relies on a canary approach, where it checks if there is a `<li>` element with the role "**separator**" at position 5.


sindresorhus:
Looks good, but the code needs some comments and better variable names. When looking at the code without the context of this PR, it's impossible to know what it does.

PR #420, opened 2018-03-26
Fix mute notifications setting on startup

### LINKED ISSUE #414: Muted notification option checkbox is in reverse with option
OS: Ubuntu 17.10 x64
Caprine version: 2.10

image
I'm getting notifications with checked Muted Notifications option enabled.
image

I think option should called "Show Notifications" :)

### LINKED ISSUE #417: Desktop notifications turned off
After every system reboot the option to send desktop notifications turns itself off. The app is set as a startup app and the option start minimized is checked if that matters.

Deepin Linux 15.5
Caprine 2.10.0

dusansimic:
This is a fix for long-time issue in Caprine. Mute notifications option will now be saved and set after startup again.

Closes #417
Closes #414

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,5 +1,6 @@
 'use strict';
 const electron = require('electron');
+const elementReady = require('element-ready');
 const config = require('./config');
 
 const {ipcRenderer: ipc} = electron;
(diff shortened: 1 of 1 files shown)


fnwbr:
@sindresorhus Do you think you can do a new release now?

PR #422, opened 2018-04-01
Added emoji color config

dusansimic:
I've added event listeners for buttons of emoji colors in preferences. This will save the settings and set saved emoji color on startup.
This PR requires https://github.com/sindresorhus/caprine/pull/420 because it fixes the problem when setting saved config on startup.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,5 +1,6 @@
 'use strict';
 const electron = require('electron');
+const elementReady = require('element-ready');
 const config = require('./config');
 
 const {ipcRenderer: ipc} = electron;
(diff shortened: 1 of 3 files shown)


sindresorhus:
I don't understand, why is this needed? Facebook already persists this.

dusansimic:
It doesn't work on my machine (emoji color is always yellow after startup). I think I should've opened an issue first.

sindresorhus:
@dusansimic Then we should rather look into *why* it doesn't work. I don't want lots of code and complexity for something that Facebook already should handle.

dusansimic:
This seems like a problem with messenger.com. It doesn't work on either Chrome or Firefox. I guess this is not a Caprine issue so you're right about keeping the code clean and not adding this.

Issue #170, opened 2017-03-08
Touch Bar support


sindresorhus:
The [next version of Electron](https://github.com/electron/electron/releases/tag/v1.6.3) will come with Touch Bar support. See: https://github.com/electron/electron/pull/8095

I'm interested in hearing people's opinion on what it should contain? What shortcuts would be useful to have? What are the most common actions you do?

I think it should at least have buttons to go to the next/previous conversation.

mirshko:
It would be cool to have top chatted with people on it. Similar to the favorites chat heads on the mobile Messenger app. Maybe shortcut to start new private chat (not sure if those are supported on the web tho)

marcooliveira:
Will be adding several comments, so people can *thumbs up* on what they like.

- Emojis:
  - Could be the most common.
  - Ideally would be the most frequently used.

marcooliveira:
Next/previous conversation.

marcooliveira:
Favourite contacts: Like @mirshko said, their "heads".

marcooliveira:
Start audio & video call.

marcooliveira:
Send a "Like". I feel like this is a big one. Separating it from the "Emojis" one, since it just has a lot of meaning and is very frequent.

marcooliveira:
"New message":
  - Could use the same icon that's already on top left.
  - After tap, would show friends with some heuristics, based on combination of favourites and most recent friends.

sindresorhus:
If anyone needs to test the Touch Bar and don't have a new MacBook Pro, there's https://github.com/sindresorhus/touch-bar-simulator

CvX:
Touch Bar support has been added in #426! 🎉 Closing.

PR #426, opened 2018-04-11
Add TouchBar support

cawa-93:
image
image

#170


sindresorhus:
Nice! Thanks for starting this :)

I think it would be nice to also include the profile picture in the button to make it faster to visually scan the different conversations.

cawa-93:
@sindresorhus It may not be so easy to implement:
1. It seems that there is no possibility to display in the button two icons: the avatar and the marker of the new message.
2. The button icon should be [nativeImage](https://github.com/electron/electron/blob/master/docs/api/native-image.md)

CvX:
@cawa-93 wouldn't it be possible to create nativeImage icons on the fly (using `nativeImage.createFromBuffer()` or `nativeImage.createFromDataURL()`) by composing a new message marker onto an avatar image?

cawa-93:
@sindresorhus I do not know how to do this with minimal loss in performance (Without the use of canvas, etc.)

sindresorhus:
> I do not know how to do this with minimal loss in performance (Without the use of canvas, etc.)

It wouldn't cause any noticeable performance loss. It's just combining a few tiny avatar images together with a "new message" marker. That can easily be done with `<canvas>`, for example: https://github.com/sindresorhus/caprine/blob/967f2202be7e3dd5c537b9220470867381b3fddb/browser.js#L112 We could cache the avatars, so they're only generated once per session. But it's not a blocker for this PR if you don't want to do it (?). It can be done later. This PR is already awesome.

cawa-93:
Maybe sometime later

sindresorhus:
Thank you, @cawa-93! 

I've opened a new issue about showing avatars: https://github.com/sindresorhus/caprine/issues/430

PR #431, opened 2018-04-18
Show profile picture in the Touch Bar

### LINKED ISSUE #430: Show profile picture in the Touch Bar
https://github.com/sindresorhus/caprine/pull/426 added Touch Bar support, but a button can only have one image and we use it for the unread message indicator. What we could do instead is to combine the profile picture with the unread message indicator using `<canvas>`, as done here: https://github.com/sindresorhus/caprine/blob/967f2202be7e3dd5c537b9220470867381b3fddb/browser.js#L112-L127
We should cache the profile pictures, so they're only generated once per session.

cawa-93:
Fixes #430 

image

Added rendering of user avatars to the touch bar. I do not know much about `<canvas>, so please correct my code in potentially problematic places.

Also, I was not able to figure out how to work with group dialogs in which there are several avatars. Therefore, only the first picture is used for such dialogues


browser.js
```diff
@@ -328,6 +331,65 @@ function initTouchBar() {
...
+			ctx.fill();
+			resolve(canvas.toDataURL());
+		});
+		im.src = avatar;
+	});
+}
+
+// Generates the date url for a picture for storing it for reuse
+function getAvatarDataUrl(img) {
+	return new Promise(resolve => {
```
sindresorhus: The code here is almost the same as the above. It could be abstracted into a method like:

```js
await canvasToImage(img.src, canvas => {
```

browser.js
```diff
@@ -328,6 +331,65 @@ function initTouchBar() {
...
+			ctx.beginPath();
+			ctx.ellipse(this.naturalWidth - markerSize, markerSize, markerSize, markerSize, 0, 0, 2 * Math.PI);
+			ctx.fill();
+			resolve(canvas.toDataURL());
+		});
+		im.src = avatar;
+	});
+}
+
+// Generates the date url for a picture for storing it for reuse
```
sindresorhus: `date` => `data`

browser.js
```diff
@@ -328,6 +331,65 @@ function initTouchBar() {
...
+function getRenderedUnreadMarker(img, unread) {
+	return new Promise(async resolve => {
+		const avatar = await getAvatarDataUrl(img);
+		if (!unread) {
+			return resolve(avatar);
+		}
+
+		const im = new Image();
+		im.crossOrigin = 'anonymous';
+		im.addEventListener('load', function () {
```
sindresorhus: Use arrow function. You can just refer to `im` instead of `this`.

sindresorhus:
Can you reduce the space between the avatar and the name?

touch bar shot 2018-04-27 at 15 36 48

cawa-93:
@sindresorhus about space between the avatar and the name — https://github.com/electron/electron/issues/12658

sindresorhus:
This looks really great now. Thanks for doing all this 🙌

Issue #437, opened 2018-04-30
Notifications Muted... and dark mode unreaded messages...


dachronos:
when aplication starts the option "Mute notification" are checked... I uncheck but if close the app and start again, again "Mute notification" are checked...

In "Dark Mode" the unreaded messages no have a hightlight color.

---

Caprine 2.12.1
Electron 1.8.6
linux x64 3.16.0-4-amd64

dusansimic:
@dachronos the 'highlight color' in dark mode is actually the bold font on unread messages and conversation name. But I agree this is not visible enough, maybe it should get a lighter shade or something like that.

PR #438, opened 2018-04-30
Fix mute notifications

dusansimic:
Config variable `notificationsMuted` is true if notifications should be
muted, not unmuted. There was a bug in if statement in `browser.js` when
it checks config state and real state.

Closes: #437


sindresorhus:
Thanks for fixing :)

dusansimic:
No problem :+1:

PR #439, opened 2018-05-02
Remove left and right padding in touch bar buttons

cawa-93:
Decrease the indentations to the left and to the right. Thus, the image area became more compact, which made it possible to slightly increase the resolution of the image

Before:
tg_image_4091352590
After:
image


sindresorhus:
Nice! :)

Issue #432, opened 2018-04-25
Spell check


tomhoward7:
Is it possible to add a spell check for the message box?

Issue #14, opened 2015-09-24
Implement spell checking


fregante:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2465%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/caprine/issues/14)

All apps on OS X can use the system's spellchecker, but it appears to be disabled in Caprine

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $65.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($60.00)
- [<img src='https://avatars3.githubusercontent.com/u/115112?v=4' alt='hazrpg' width=24 height=24> hazrpg](https://issuehunt.io/u/hazrpg) ($5.00)

### Submitted pull Requests
- [#1336 Spellcheck with electron-context-menu update!](https://issuehunt.io/r/sindresorhus/caprine/pull/1336)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/caprine/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

sindresorhus:
I opened an Electron issue regarding this  →https://github.com/atom/electron/issues/3189

sindresorhus:
Follow-up issue → https://github.com/atom/electron/issues/3211

bfred-it:
:+1:

dudeofawesome:
Has there been any progress on this?

I might try to tackle it using [electron-spellchecker](https://github.com/electron-userland/electron-spellchecker).

sindresorhus:
@dudeofawesome No progress. A pull request would be lovely. I think it should be added to [`electron-context-menu`](https://github.com/sindresorhus/electron-context-menu) though, as it's what Caprine uses for the context menu, and that way it would benefit many apps, not just Caprine.

nahive:
Any progress on that? I'm using macOS text replacement keyboard feature and it's the only thing blocking me from using Caprine.

nahive:
[...]

PR #443, opened 2018-05-04
Add spellcheck

cawa-93:
For #432


sindresorhus:
Please see https://github.com/sindresorhus/caprine/issues/14 and https://github.com/sindresorhus/caprine/pull/247 This needs to be integrated into the right-click menu to be useful. We're using https://github.com/sindresorhus/electron-context-menu for the context menu.

PR #445, opened 2018-05-08
Support the `flashWindowOnMessage` option on Linux too

### LINKED ISSUE #435: Set urgency hint when a message is received and Caprine isn't focused
I use i3wm on Linux and it would be nice if Caprine set the urgency hint flag when an new message was received (and the window isn't focused) so that you'd know that you have unread messages. This is useful if you're not at the computer when the message is received, when you get back you'll still get notified through the workspace icon/window border.

While googling for this I stumbled upon a similar feature request (https://github.com/yakyak/yakyak/issues/101) for a Google Hangouts desktop application called YakYak and it seems like it is doable in Electron.

LackOfGumption:
Resolves #435.
As flashFrame is now platform agnostic (https://electronjs.org/docs/all#winflashframeflag).

Verified it worked on Ubuntu 18.04 i3wm.

### DIFF
--- a/index.js
+++ b/index.js
@@ -80,10 +80,11 @@ function updateBadge(title, titlePrefix) {
 				mainWindow.webContents.send('render-overlay-icon', messageCount);
 			}
 		}
+	}
 
-		if (config.get('flashWindowOnMessage')) {
-			mainWindow.flashFrame(messageCount !== 0);
-		}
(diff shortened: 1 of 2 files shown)


sindresorhus:
I just checked and it does nothing on macOS, so it should not be shown there.

sindresorhus:
Closing for lack of response.

PR #464, opened 2018-05-19
Add snap support

### LINKED ISSUE #460: Publish to Snap Store
I would be interested in working on this if you think it is a good idea :slightly_smiling_face: 

https://snapcraft.io/store

dusansimic:
Adding snap config and fixing issues regarding this PR.

* [x] Fix links opening in default browser instead app window

---

Fixes #460

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,12 +72,10 @@
 				"AppImage",
 				"deb",
 				"snap"
-			]
-		},
-		"snap": {
+			],
 			"synopsis": "Elegant Facebook Messenger desktop app",
 			"description": "Caprine is an unofficial and privacy focused Facebook Messenger app with many useful features.",
-			"category": "Chat"
+			"category": "Network;Chat"
 		}
 	}
 }
--- a/package.json
+++ b/package.json
@@ -70,8 +70,14 @@
 		"linux": {
 			"target": [
 				"AppImage",
-				"deb"
+				"deb",
+				"snap"
 			]
+		},
+		"snap": {
+			"synopsis": "Elegant Facebook Messenger desktop app",
+			"description": "Caprine is an unofficial and privacy focused Facebook Messenger app with many useful features.",
+			"category": "Chat"
 		}
 	}
 }


package.json
```diff
@@ -68,8 +68,14 @@
 		"linux": {
 			"target": [
 				"AppImage",
-				"deb"
+				"deb",
+				"snap"
 			]
+		},
+		"snap": {
```
kounelios13: I think that you should also include the `confinement` field(Not sure though if this is required)
https://www.electron.build/configuration/snap

sindresorhus:
@dusansimic Please make sure you review all the config in https://www.electron.build/configuration/snap and determine what's needed. This also needs to be manually tested on Ubuntu to ensure it works.

package.json
dusansimic: Added `confinement` property. It's not required but it needs to be set to `classic` because Caprine uses users directories. Thanks for the note.

dusansimic:
There is an issue when using sandbox mode in electron (link in PR message). The fix is being reviewed and will probably be in next electron version (current v2.0.2). I'll watch on updates about that. Other than that there are no issues with snap build.

index.js
```diff
@@ -180,6 +180,7 @@ function createMainWindow() {
 		autoHideMenuBar: config.get('autoHideMenuBar'),
 		darkTheme: isDarkMode, // GTK+3
 		webPreferences: {
+			sandbox: true,
```
sindresorhus: You need to read up on how this option works: https://electronjs.org/docs/api/sandbox-option It's not enough to just enable it and currently it throw an error now:

```
VM22:949 Uncaught Error: module not found
    at preloadRequire (<anonymous>:949:9)
    at eval (eval at <anonymous> (unknown source), <anonymous>:4:22)
    at Object.<anonymous> (<anonymous>:987:3)
    at Object.require.14.../common/atom-binding-setup (<anonymous>:990:4)
    at o (<anonymous>:2:273)
    at r (<anonymous>:2:439)
    at <anonymous>:2:468
```

package.json
klauscfhq: There is no need to switch to `classic` confinement to access user directories. The snap can be build under the default recommended `strict` confinement, where all user directories are exposed via the [`home`](https://docs.snapcraft.io/reference/interfaces) plug/interface, which is also included by default by the [`electron-builder`](https://www.electron.build/configuration/snap). If you insist on building with `classic` confiment, then a [request should be sent to the snapcraft team](https://forum.snapcraft.io/t/process-for-reviewing-classic-confinement-snaps/1460), in oder for the snap to be released to the store, but again that would be totally unneeded since `strict` confinement should work just fine.

index.js
dusansimic: Seems like sandbox can't be used in Caprine since we need some of libraries but they cant be imported whit `require()` into renderer process. I'll disable sandbox option.

package.json
dusansimic: Removed confinement option since it's not needed. Thanks for review @klauscfhq :smile:

dusansimic:
Tested all features and fixed merge conflicts. @sindresorhus please review.

package.json
klauscfhq: Everything works fine on both Ubuntu 16.04 & 18.04:



My only suggestion would be to move the `category`, `synopsis` and `description` options under the top-level [`linux`](https://www.electron.build/configuration/linux) key in order for them to be inherited by the [snap](https://www.electron.build/configuration/snap) configuration as well as the the [AppImage](https://www.electron.build/configuration/linux#appimage-options) one.

Also, here are a few known issues to keep in mind regarding electron and snapcraft:
 - All snaps are self-contained and confined from the OS and other apps through security mechanisms, which means that the electron auto-updater will fail to update the isolated, under its own dedicated [$HOME](https://docs.snapcraft.io/reference/env#generic-variables) environment variable, snap and that updates can only be delivered through the [Snap Store](https://docs.snapcraft.io/build-snaps/electron#upload-your-snap).
 - Currently snapcraft does not support [input methods](https://forum.snapcraft.io/t/cant-use-input-method-in-snap-apps/4712/19) like [`fcitx`](https://fcitx-im.org/wiki/Fcitx) & [`sogoupinyin`](https://pinyin.sogou.com/) which are widely used for typing in Chinese.
 - Currently the mouse cursor appears too small on snaps that are running under [hight DPI displays](https://github.com/klauscfhq/tusk/issues/62).

For the users facing the last two issues, until they are resolved, switching to an other Linux package should do the trick : )

package.json
```diff
@@ -70,8 +70,14 @@
...
+			"synopsis": "Elegant Facebook Messenger desktop app",
+			"description": "Caprine is an unofficial and privacy focused Facebook Messenger app with many useful features.",
+			"category": "Chat"
```
sindresorhus: This should be `Network;Chat` as `Chat` is not a main category and it has to contain a main category according to the spec.

dusansimic: Fixed it :+1:

PR #470, opened 2018-05-25
Open conversations from the Dock menu

cawa-93:
image

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -329,22 +329,19 @@ function closePreferences() {
 async function sendConversationList() {
 	const sidebar = document.querySelector('[role=navigation]');
 
-	const conversations = [];
-	for (const el of sidebar.querySelectorAll('._1ht1')) {
-		conversations.push({
-			label: el.querySelector('._1ht6').textContent,
-			selected: el.classList.contains('_1ht2'),
-			unread: el.classList.contains('_1ht3'),
-			icon: await getDataUrlFromImg( // eslint-disable-line no-await-in-loop
-				el.querySelector('._55lt img'),
-				el.classList.contains('_1ht3')
-			)
-		});
-
-		if (conversations.length >= 10) {
-			break;
-		}
-	}
+	const conversations = await Promise.all(
+		Array.from(sidebar.querySelectorAll('._1ht1'))
+			.splice(0, 10)
+			.map(async (el) => ({
+				label: el.querySelector('._1ht6').textContent,
+				selected: el.classList.contains('_1ht2'),
+				unread: el.classList.contains('_1ht3'),
+				icon: await getDataUrlFromImg(
+					el.querySelector('._55lt img'),
+					el.classList.contains('_1ht3')
+				)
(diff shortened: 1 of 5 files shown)


sindresorhus:
Can you mention it in the readme too? It's useful, but most apps don't have anything useful in the Dock menu, so users wouldn't think to look there.

cawa-93:
@sindresorhus Would it be better to make such an image from the English system? Also, I think it would be worth adding the same functionality to other systems. However, I can not add and check it.

sindresorhus:
@cawa-93 Sure, I can fill in the screenshot later if you add a short sentence about it in the readme.

sindresorhus:
> Also, I think it would be worth adding the same functionality to other systems. However, I can not add and check it.

Windows has jump lists. We can open an issue after this for the Windows implementation with a "PR welcome" label.

cawa-93:
@sindresorhus How about Linux? Gnome?

sindresorhus:
I'm not very familiar with Linux, but I don't think it's possible there. If any Linux users know, they can do a pull request.

readme.md
```diff
@@ -12,10 +12,17 @@ Caprine is an unofficial and privacy focused Facebook Messenger app with many us
...
-	<img src="https://c5.patreon.com/external/logo/become_a_patron_button@2x.png" width="160">
+	<img src="https://user-images.githubusercontent.com/1662812/40548807-66c154aa-603e-11e8-89b5-b514d426eafd.png" width="100">
 </a>
 
+### And Touch Bar support
+<img src="media/screenshot-touchbar.png" width="1085">
+
+<a href="https://www.patreon.com/sindresorhus">
+	<img src="https://c5.patreon.com/external/logo/become_a_patron_button@2x.png" width="160">
+</a>
```
sindresorhus: These should be in the Highlights section, not here.

browser.js
```diff
@@ -326,42 +326,29 @@ function closePreferences() {
...
-	const conversationListObserver = new MutationObserver(sendTouchBar);
-	conversationListObserver.observe(sidebar, {
-		subtree: true,
-		childList: true,
-		attributes: true,
-		attributeFilter: ['class']
-	});
+	ipc.send('conversations', conversations);
 }
+window.s = ipc.send;
```
sindresorhus: ?

cawa-93: Wow. Sorry. This is a piece of code I used for experimentation

readme.md
```diff
@@ -12,10 +12,6 @@ Caprine is an unofficial and privacy focused Facebook Messenger app with many us
 	<img src="media/screenshot.png" width="846">
 </a>
 
-<a href="https://www.patreon.com/sindresorhus">
-	<img src="https://c5.patreon.com/external/logo/become_a_patron_button@2x.png" width="160">
-</a>
-
```
sindresorhus: ?

sindresorhus:
Looks good :)

PR #476, opened 2018-05-29
Fix the `New Conversation` button overlapping the traffic light controls on macOS

cawa-93:
Before
image
After
image

### DIFF
--- a/browser.css
+++ b/browser.css
@@ -164,6 +164,10 @@ html.sidebar-hidden ._1enh {
 	._4_j5 {
 		display: none !important;
 	}
+	
+	._36ic._5l-3 {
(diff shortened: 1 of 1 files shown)


browser.css
```diff
@@ -164,6 +164,10 @@ html.sidebar-hidden ._1enh {
 	._4_j5 {
 		display: none !important;
 	}
+	
+	._36ic._5l-3 {
```
sindresorhus: Can you add a comment on what this does to make the code readable?

sindresorhus:
This should only be applied on macOS, as Linux and Windows have a real window frame. You can use `.os-darwin` like done elsewhere.

PR #482, opened 2018-06-06
New tray icons

jimmy213:
I added a grey version of the icon (using a b&w filter) to show that there aren't any unread messages, and I put the old icons to show that there are new messages.


jimmy213:
I built it earlier today to see how it really looks and personally I'll keep that way.

No new message
caprine_new

New message arrived
caprine_new2

sindresorhus:
I think the gray icon should be slightly darker gray.

jimmy213:
Yes, a darker one looks a bit nicer. Something like that?

caprine_new3

sindresorhus:
Perfect

alekksander:
and gradient stays?

PR #488, opened 2018-06-07
Don’t show the main window on conversations change

### LINKED ISSUE #483: Caprine window brought to front whenever a message is read or received
Version 2.14.0.797. A Caprine window is opened and brought to front whenever a message is read or received. This occurs even when notifications are muted and Caprine is running in the background.

CvX:
Fixes #483 (a bug introduced in #470)

I don't think there's any reason to display the window on any change in the conversations list.

### DIFF
--- a/index.js
+++ b/index.js
@@ -249,7 +249,6 @@ app.on('ready', () => {
 				return;
 			}
 
-			mainWindow.show();
 			const items = conversations.map(({label, icon}, index) => {
 				return {
 					label: `${label}`,


cawa-93:
This should be called in the callback function
```js
const items = conversations.map(({label, icon}, index) => {
  return {
    label: `${label}`,
    icon: nativeImage.createFromDataURL(icon),
    click: () => {
      mainWindow.show();
      sendAction('jump-to-conversation', index + 1);
    }
  };
});
```

CvX:
@cawa-93 You're right, fixed. Thank you!

Issue #498, opened 2018-06-19
Notifications preference not remembered


petko:
Every time I start Caprine, Desktop Notifications are enabled, even though I disabled them in the Preferences during the previous launch (and this is respected for the duration of the previous session). I use Caprine 2.14.2 under Windows 10.

petko:
The problem is back in 2.16.0, I am afraid..

sindresorhus:
// @dusansimic

dusansimic:
@petko you're supposed to toggle notifications option in menu bar, not in preferences. There is no event listener for toggling notifications options through preferences, but that feature could be added.

petko:
Sorry, I am not aware of the internals of Caprine. But I think that it should be possible in the Preferences too. If not, it either should be disabled there or there should be some kind of reminder that it is not remembered between sessions. Because if you change it in the Preferences, it stays for the duration of the session.

dusansimic:
I agree, @sindresorhus any thoughts on adding event listeners for preferences dialog or hiding options from preferences?

sindresorhus:
@dusansimic We can just hide it from Preferences.

PR #500, opened 2018-06-20
Fix some preferences issues

dusansimic:
Fixed color of options in preferences in dark mode and issue with
setting notifications preference.

Fixes: #498

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -105,13 +105,11 @@ function setSidebarVisibility() {
 }
 
 ipc.on('toggle-mute-notifications', async (event, defaultStatus) => {
-	const wasPreferencesOpen = isPreferencesOpen();
-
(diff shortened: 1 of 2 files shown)


cawa-93:
@dusansimic I'm wondering: why did you use such a structure
```js
if (isPreferencesOpen()) {
	// ...
}

if (isPreferencesOpen()) {
	// ...
}
```
instead of this:
```js
const isPreferencesOpen = isPreferencesOpen()
if (isPreferencesOpen) {
	// ...
}

if (isPreferencesOpen) {
	// ...
}
```
Is the second option not more effective at the expense of smaller operations with the DOM?

dusansimic:
Sorry, I forgot to change that to the old version (second one) when i was debugging that method. I'll get that fixed right away. Thanks.

Issue #298, opened 2017-10-03
Reply via notification [macOS]


Vilscon:
Other apps in general, but Goofy (as Caprine alternative) have the option to reply via a notification. I hereby request the same feature 👍

sindresorhus:
👍 Would be useful. Pull request welcome if anyone wants this :)

Goofy is open source, so I guess one could look at the source to see how it's implemented: https://github.com/danielbuechele/goofy-core Don't copy-paste the code though.

salmanulfarzy:
Does Goofy [still][1] support [inline replies][2] in notifications ? Looks like they are using simple notifications from electron. 

https://github.com/danielbuechele/goofy-core/blob/c1c88a46b6e4cd7a296228b921e527ad1870945e/src/assets/fb.js#L117-L121

Electron [supports][3] inline replies from [v1.7.3][4] or we could make use of [node-mac-notifer][5]

[1]: https://github.com/danielbuechele/goofy/pull/321
[2]: https://github.com/danielbuechele/goofy/issues/370
[3]: https://github.com/electron/electron/blob/v1.7.3/docs/api/notification.md
[4]: https://github.com/electron/electron/commit/e82af41591e00589d990a901cdd98ea617e8f34b
[5]: https://github.com/CharlieHess/node-mac-notifier

Rezonansowy:
I'd like to provide some reference pics of sample implementations:

![click-on-reply-in-message-notification-on-mac](https://user-images.githubusercontent.com/6929363/34421200-e9224c6a-ebdb-11e7-9579-2095985f2152.png)
<img width="379" alt="image 2016-04-14 at 2 37 47 pm" src="https://user-images.githubusercontent.com/6929363/34421210-f93ce84e-ebdb-11e7-8333-76c77eafb2fc.png">
![reply-to-messags-mac-notification](https://user-images.githubusercontent.com/6929363/34421212-ff832362-ebdb-11e7-82d0-1bc52c0b04c1.png)
![send-message-reply-notification-mac](https://user-images.githubusercontent.com/6929363/34421223-09c99400-ebdc-11e7-8985-901da197d0f0.png)

tobychin:
I'd like to take a gander at this! This would be a great feature to have.

IssuehuntBot:
@jonivr has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/42574339/issues/298)

IssuehuntBot:
[...]

PR #504, opened 2018-06-21
Make notifications stay in the Notification Center on macOS

### LINKED ISSUE #317: Notifications don't show up in the Notification Center on macOS
When a new message notification appears, it can't be found in the Notification Center after it disappears from the screen. Interestingly, when I had Do Not Disturb turned on, the notifications could be seen in the notification center, however they disappeared when I turned Do Not Disturb off. 
I'm currently on the newest version of High Sierra, but I remember this issue occurred in Sierra, too.
I checked my notification settings in System Preferences and I have the same settings for Caprine (including checked 'Show in Notification Center') as for other apps, whose notifications are remaining in the Notification Center. 

---

macOS High Sierra 10.13.1
Caprine 2.7.1
Electron 1.6.7
darwin x64 17.2.0

janosorcsik:
Fixes #317


sindresorhus:
It's not clear why you have reimplemented notifications in the main process. Can you explain your changes?

janosorcsik:
The messenger.com use the HTML5 Notification API. I guess this differs from Electron Notification because the notification does not stay in the Notification Center.
You mentioned the Goofy App in #298 and I found this solution in it.

janosorcsik:
Ok. Thank you.
I am going to try monkey patch but I am not an advanced JS developer. Therefore I cannot say this will be fast.
Do you have any suggestions?

sindresorhus:
@janosorcsik Is this ready to be reviewed?

janosorcsik:
@sindresorhus Yes, it is.
Otherwise, I got a PR in my fork. We discussed the reply function. 
[Please check it out.](https://github.com/janosorcsik/caprine/pull/1)

sindresorhus:
This works well. Sorry for the delay.

sindresorhus:
Would you be able to add a click action to the notification so when it's clicked it opens the correct conversation?

browser.js
```diff
@@ -481,3 +481,28 @@ document.addEventListener('keydown', event => {
...
+			canvas.width = img.width;
+			canvas.height = img.height;
+
+			ctx.drawImage(img, 0, 0, img.width, img.height);
+
+			ipc.send('notifications', title, body, canvas.toDataURL(), silent);
+		});
+		return false;
+	};
+	return Object.assign(customNotification, notification);
```
sindresorhus: Use object-spread.

browser.js
```diff
@@ -481,3 +481,28 @@ document.addEventListener('keydown', event => {
 		jumpToConversation(num);
 	}
 });
+
+window.Notification = (function (notification) {
```
sindresorhus: Use arrow function.

browser.js
```diff
@@ -481,3 +481,28 @@ document.addEventListener('keydown', event => {
...
+		img.addEventListener('load', () => {
+			const canvas = document.createElement('canvas');
+			const ctx = canvas.getContext('2d');
+
+			canvas.width = img.width;
+			canvas.height = img.height;
+
+			ctx.drawImage(img, 0, 0, img.width, img.height);
+
+			ipc.send('notifications', title, body, canvas.toDataURL(), silent);
```
sindresorhus: `notifications` => `notification`

sindresorhus: And I think it would be better to send the arguments as an object instead.

```js
ipc.send('notifications', {title, body, canvas: canvas.toDataURL(), silent});
```

browser.js
janosorcsik: This does not work because it needs to return a function.

janosorcsik:
@sindresorhus Everything is done :)

.vscode/launch.json
```diff
@@ -0,0 +1,30 @@
+{
```
sindresorhus: Don't add unrelated files.

sindresorhus:
Really nice work on this, @janosorcsik 🎉

PR #506, opened 2018-06-24
Let Ctrl+1..Ctrl+9 work for Azerty keyboards too

Smile4ever:
To make Ctrl+1 till Ctrl+9 work with Azerty keyboard layouts, without needing to use the shift key every time, I created this simple pull request.

event.code contains either:
* Digit1..Digit9
* Numpad1..Numpad9

Getting the last character from the string should be a number, so we check that number.

I think this change is rather trivial and I would love to see it in a new release of Caprine.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -469,7 +469,7 @@ document.addEventListener('keydown', event => {
 		previousConversation();
 	}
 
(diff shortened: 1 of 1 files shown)


Smile4ever:
Thank you.

PR #516, opened 2018-07-02
Make the "Toggle dark mode" option a checkbox

cawa-93:
I changed the type of the menu item "Toggle dark mode" to the "Enable dark mode" checkbox.


sindresorhus:
Good idea 🙌

If you're interested, would be great to do this with the "Toggle Vibrancy" and "Toggle Sidebar" options too.

PR #517, opened 2018-07-02
Don't hide the right sidebar in compact mode

cawa-93:
I think it's a mistake to forcibly hide the right panel in compact mode.
This prevents users from accessing some important features, such as viewing attachments, discussion settings, and more.
It's best to leave the right panel visibility at the discretion of each individual user.


sindresorhus:
You're right. The user can just hide it if they don't want it.

Issue #515, opened 2018-07-02
Add Sparkle.framework for auto updating on macOS


gklka:
https://sparkle-project.org

Almost every macOS app which is not in the App Store uses this handy framework for updating. I think Caprine needs is very much too.

sindresorhus:
Caprine already has silently auto-updates using https://www.electron.build/auto-update

gklka:
Then it is not working. I’m using Caprine since 1 year now and not even once updated itself.

janosorcsik:
The auto-update only checks when the app started. We need to check more often.

gklka:
I practically never quit, so this is the problem for me.

PR #525, opened 2018-07-06
Check for new version more often

janosorcsik:
Check update every 4th hour.

Related to #515

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,7 +26,8 @@ app.setAppUserModelId('com.sindresorhus.caprine');
 if (!isDev) {
 	autoUpdater.logger = log;
 	autoUpdater.logger.transports.file.level = 'info';
-	autoUpdater.checkForUpdates();
+	const updateInterval = 60 * 60 * 1000 * 4; // 4 hours
+	setInterval(autoUpdater.checkForUpdates(), updateInterval);
 }
 
 let mainWindow;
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -26,7 +26,8 @@ app.setAppUserModelId('com.sindresorhus.caprine');
 if (!isDev) {
 	autoUpdater.logger = log;
 	autoUpdater.logger.transports.file.level = 'info';
-	autoUpdater.checkForUpdates();
+	const updateInterval = 60 * 60 * 1000 * 4; // 4 hours
+	setInterval(autoUpdater.checkForUpdates(), updateInterval);
```
sindresorhus: This won't actually work since you're passing the function result, not the function to `setInterval`.

index.js
```diff
@@ -26,7 +26,8 @@ app.setAppUserModelId('com.sindresorhus.caprine');
 if (!isDev) {
 	autoUpdater.logger = log;
 	autoUpdater.logger.transports.file.level = 'info';
-	autoUpdater.checkForUpdates();
+	const updateInterval = 60 * 60 * 1000 * 4; // 4 hours
```
sindresorhus: Make this a `FOUR_HOURS` constant instead. Then you don't need the comment.

sindresorhus: And this order would make more sense: `1000 * 60 * 60 * 4`

cawa-93: Perhaps it's better to put into the utility the constant `HOUR` then in the code simply multiply it by 4 `HOUR * 4`? In this case, it will be easier to further change the frequency of the interval
`HOUR * 5`

sindresorhus: Nah, I don't see the point for something as simple and specific as this.

PR #560, opened 2018-08-14
enter selects first result in search

rahulgs12:
Currently, Enter in the search field isn't useful because the search refreshes on keydown, it should instead select the first search result. This also applies to the GIF picker.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -232,6 +232,21 @@ function jumpToConversation(key) {
 	selectConversation(index);
 }
 
+function selectFirstOption(inputField) {
+	if (!inputField || !inputField.placeholder) { return }
+	if (inputField.placeholder.includes('GIF')) {
+		let firstGif = document.querySelector('._358')
+		if (firstGif) {
+			firstGif.click()
+		}
+	} else if (inputField.placeholder.includes('Messenger')) {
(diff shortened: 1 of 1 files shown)


browser.js
```diff
@@ -455,6 +472,10 @@ window.addEventListener('load', () => {
 document.addEventListener('keydown', event => {
 	const combineKey = process.platform === 'darwin' ? event.metaKey : event.ctrlKey;
 
+	if (event.key === 'Enter' && document.activeElement && document.activeElement.className === '_58al') {
```
sindresorhus: `'_58al'` should be a constant variable with a descriptive name.

browser.js
```diff
@@ -232,6 +232,23 @@ function jumpToConversation(key) {
 	selectConversation(index);
 }
 
+function selectFirstOption(inputField) {
```
sindresorhus: This function name needs to be more specific.

sindresorhus:
This is useful, but it's not clear enough to the user. I think it would be better to move the focus to the first element and add some style for when it's focused. That way it's totally clear that when you press enter you choose the first element. It also makes it easy to tab to select the next element.

sindresorhus:
Closing for lack of response.

Issue #557, opened 2018-08-13
"Keep me signed in" should be checked by default


manfreed:
Reason: This is the typical behavior of desktop apps.

sindresorhus:
Agreed. PR welcome :)

PR #564, opened 2018-08-19
Make the login "Keep me signed in" checkbox checked by default

dusansimic:
Added as requested.

Closes: https://github.com/sindresorhus/caprine/issues/557

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -451,9 +451,9 @@ window.addEventListener('load', () => {
 		});
 	}
 
-	const isKeepMeSignedInChecked = document.getElementById('u_0_0').checked;
-	if (!isKeepMeSignedInChecked) {
-		document.getElementById('u_0_0').checked = true;
+	const keepMeSignedInCheckbox = document.querySelector('#u_0_0');
+	if (!keepMeSignedInCheckbox.checked) {
+		keepMeSignedInCheckbox.checked = true;
 	}
 });
 
--- a/browser.js
+++ b/browser.js
@@ -451,8 +451,8 @@ window.addEventListener('load', () => {
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


browser.js
```diff
@@ -439,15 +439,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
+		const conversationListObserver = new MutationObserver(sendConversationList);
+		conversationListObserver.observe(sidebar, {
+			subtree: true,
+			childList: true,
+			attributes: true,
+			attributeFilter: ['class']
+		});
+	}
+
+	const isKeepMeSignedInChecked = document.getElementById('u_0_0').checked;
```
sindresorhus: Use `document.querySelector()` and make it a variable so you don't have to get the element twice.

browser.js
```diff
@@ -439,15 +439,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
+		const conversationListObserver = new MutationObserver(sendConversationList);
+		conversationListObserver.observe(sidebar, {
+			subtree: true,
+			childList: true,
+			attributes: true,
+			attributeFilter: ['class']
+		});
+	}
+
+	const keepMeSignedInCheckbox = document.querySelector('#u_0_0');
```
sindresorhus: It should only be applied at the login page.

```js
location.pathname.startsWith('/login');
```

browser.js
```diff
@@ -439,15 +439,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
+	if (!keepMeSignedInCheckbox.checked) {
```
sindresorhus: No point in checking here. We can just always set it to `true`.

PR #583, opened 2018-09-16
fixed notification icon for windows & linux

### LINKED ISSUE #578: Icon no longer switching to blue upon new message.
---

Caprine 2.17.0
Electron 2.0.8
linux x64 4.18.5-arch1-1-ARCH

kolinger:
Tray icon/badge is no longer updated on any other platform than darwin. This was introduced by commit a281594b5f975833ff736573a08dd2a4bf7745f3.

Event registration for updating tray icon was moved to darwin specific condition - this is wrong. It should be registered for all platforms. This commit fixes this issue and enables icon to be update on notification  for all platforms. Fixes #578

### DIFF
--- a/index.js
+++ b/index.js
@@ -256,10 +256,10 @@ app.on('ready', () => {
 			});
 			app.dock.setMenu(electron.Menu.buildFromTemplate([firstItem, {type: 'separator'}, ...items]));
 		});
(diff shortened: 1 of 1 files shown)


kolinger:
See #587

Issue #582, opened 2018-09-16
no notifications


ForumLiker:
Hello. On the [deepin](https://www.deepin.org/) desktop, there are no notifications about the incoming message. Icon does not change color in tray, it is always gray.

kolinger:
Probably duplicate of #578?

sindresorhus:
Duplicate of #578

PR #587, opened 2018-09-17
Fix tray icon

dusansimic:
Tray icons was not turning into `unread` state when new messages are
received. There was a bug for group chat photo because it uses
background image on div instead img tag. Also `updateBadge()` method was
inside if statement that is true only on darwin platform so tray icon
wasn't changed on linux/win32.

Closes: https://github.com/sindresorhus/caprine/issues/578
Closes: https://github.com/sindresorhus/caprine/issues/582
Closes: #594

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -333,8 +333,9 @@ async function sendConversationList() {
 				selected: el.classList.contains('_1ht2'),
 				unread: el.classList.contains('_1ht3'),
 				icon: await getDataUrlFromImg(
-					el.querySelector('._55lt img'),
-					el.classList.contains('_1ht3')
+					el.querySelector('._55lt img') ? el.querySelector('._55lt img') : el.querySelector('._4ld- div'),
+					el.classList.contains('_1ht3'),
+					Boolean(el.querySelector('._55lt img'))
 				)
 			}))
 	);
@@ -378,7 +379,7 @@ function urlToCanvas(url, size) {
 }
 
 // Return data url for user avatar
-function getDataUrlFromImg(img, unread) {
+function getDataUrlFromImg(img, unread, isImg) {
 	// eslint-disable-next-line no-async-promise-executor
 	return new Promise(async resolve => {
 		if (!unread) {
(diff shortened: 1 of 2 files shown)


piomar123:
I tested it under Windows 7 and the icon still doesn't change on new messages.

remmyle1102:
when will this bug be fixed? :(

yermulnik:
When this is going to be released?

dusansimic:
cc @sindresorhus

browser.js
```diff
@@ -333,8 +333,9 @@ async function sendConversationList() {
 				selected: el.classList.contains('_1ht2'),
 				unread: el.classList.contains('_1ht3'),
 				icon: await getDataUrlFromImg(
-					el.querySelector('._55lt img'),
-					el.classList.contains('_1ht3')
+					el.querySelector('._55lt img') ? el.querySelector('._55lt img') : el.querySelector('._4ld- div'),
```
sindresorhus: `el.querySelector('._55lt img')` should be a variable with a descriptive name as it's used multiple times.

sindresorhus: `el.querySelector('._4ld- div')` should also be a descriptive variable as it's not clear what it is when just reading the code.

browser.js
```diff
@@ -389,6 +390,9 @@ function getDataUrlFromImg(img, unread) {
 			return resolve(img.dataUnreadUrl);
 		}
 
+		if (!isImg) {
+			img.src = img.style.backgroundImage.substring(5, img.style.backgroundImage.length - 2);
+		}
```
sindresorhus: This logic should not be in this function. This function is purely to make a data URl from an image. Move it to somewhere around line 327.

browser.js
```diff
@@ -389,6 +390,9 @@ function getDataUrlFromImg(img, unread) {
 			return resolve(img.dataUnreadUrl);
 		}
 
+		if (!isImg) {
+			img.src = img.style.backgroundImage.substring(5, img.style.backgroundImage.length - 2);
```
sindresorhus: Use `slice` instead of `substring` and add a short comment on what this does.

browser.js
```diff
@@ -389,6 +390,9 @@ function getDataUrlFromImg(img, unread) {
 			return resolve(img.dataUnreadUrl);
 		}
 
+		if (!isImg) {
```
sindresorhus: `isImg` => `isImage`

sindresorhus:
Thanks for fixing, @dusansimic :)

PR #608, opened 2018-10-10
Fix border on links in light and dark mode

janosorcsik:
In white mode no border on liks:
kepernyofoto 2018-10-10 - 9 18 52

In dark mode ugly:
kepernyofoto 2018-10-10 - 9 23 43


browser.css
```diff
@@ -353,3 +353,8 @@ a,
 ._6f6l {
 	display: block !important;
 }
+
+.__6j._43kk {
```
sindresorhus: Add a comment about what this selector targets. Otherwise, the code is totally unreadable because of the obfuscated selectors.

dark-mode.css
```diff
@@ -699,6 +699,11 @@ html.dark-mode .__6l {
 	color: var(--base-seventy);
 	background: var(--container-color);
 }
+
 html.dark-mode ._4gd0 {
 	background: transparent;
 }
+
+html.dark-mode .__6j._43kk {
```
sindresorhus: Same as https://github.com/sindresorhus/caprine/pull/608/files#r224308662

PR #618, opened 2018-10-25
Re-enable vibrancy support for macOS users

phoomparin:
Since Caprine is now using Electron version 3, I wonder if the rendering bugs are gone and it is safe to re-enable vibrancy support for macOS users? I haven't noticed any graphical artifacts from enabling dark mode and vibrancy as of now, so I've submitted this PR in case it's okay to re-enable vibrancy.

I've also added some selectors to make the chat's title bar and the shared link's preview transparent.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -148,13 +148,13 @@ ipc.on('toggle-unread-threads-view', () => {
 
 function setDarkMode() {
 	document.documentElement.classList.toggle('dark-mode', config.get('darkMode'));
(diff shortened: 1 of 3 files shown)


omarojo:
:) please re enable vibrancy. So we all can enjoy it. The official download from the website doesnt have the vibrancy feature anymore.

sindresorhus:
Great! I don't see any rendering artifacts either.

PR #620, opened 2018-10-25
Add an option to disable hardware acceleration

### LINKED ISSUE #469: Request - Power safe mode
Hello,
I recently noticed that Caprine (MacOS) is switching graphics card from Integrated to High Performance which causes undesired energy impact.
Would be great if it was possible to switch.
Thank you.

kevinjalbert:
There is some demand to disable hardware acceleration to prevent electron from kicking GPUs into high-performance modes (effectively using a lot more energy).

This change adds a new menu option that toggles a config value to disable the hardware acceleration. By default, the hardware acceleration is enabled to stay parity with the current behaviour.

-----

There is some concern around rendering artifacts as mentioned in #469, although some reports are indicating with the recent build they aren't present.

---

Fix #469

### DIFF
--- a/config.js
+++ b/config.js
@@ -3,6 +3,7 @@ const Store = require('electron-store');
 
 module.exports = new Store({
 	defaults: {
+		hardwareAcceleration: true,
 		darkMode: false,
 		vibrancy: false,
 		zoomFactor: 1,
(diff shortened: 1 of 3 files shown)


janosorcsik:
If the hardware acceleration setting has changed, you need to restart the app.

kevinjalbert:
@janosorcsik correct, which is why I've named the menu item _"Hardware Acceleration (after restart)"_.

PR #623, opened 2018-10-26
Change close window behaviour on macOS to the default behaviour instead of hide

### LINKED ISSUE #288: Closing Caprine on macOS
I don't have any specifics here (maybe I can dig later), but I think closing caprine on macOS is not done "by the book".

I'm using Amethyst (tiled window manager) for macOS and it re-renders (moves and resizes) windows in a screen whenever one window is closed/moved or otherwise changed.

I guess then Caprine is closed, such window-event Amethyst listens to is never fired off, and the space Caprine filled, stays filled until I force-kill Caprine which sometimes results so that the space re-renders, but not always.

Similar electron-apps, such as Slack, have said window-event fired correctly I am assuming and space re-renders correctly.

(This is a little vague, I know. Again, my assumption is that a particular window-event is not fired which macOS normally receives.)

Edit: Funny enough. If I close Caprine normally, and ask Amethyst to force-render space, it will not fill the space Caprine had. Maybe Caprine is claiming said space even if minimised. Killing Caprine, and then force re-render results in space being filled.

pcjl:
Fixes #288

I'm proposing to change the app's macOS close window behaviour to be the default behaviour for closing macOS apps, which uses `win.hide()` instead of the "hide behaviour" (<kbd>Cmd</kbd><kbd>H</kbd>), which uses `app.hide()`.

I realize that this was an intentional change, and I saw the issue you opened in the Electron repo [here](https://github.com/electron/electron/issues/3694). However, I think this should be reverted for the following reasons:

1. It causes problems with window managers such as [Amethyst](https://github.com/ianyh/Amethyst). As described in #288, when using Amethyst, closing and re-opening Caprine doesn't re-render windows properly.

2. The default behaviour of all macOS apps is to have the "close behaviour" when closing an app, and the "hide behaviour" when hiding an app. However, as it stands right now, closing and hiding have the same behaviour. I'm assuming the reason why this was done in the first place was so that once the app closes, it'll focus on another app instead of keeping the focus on Caprine. While I understand that that may be useful, it's not standard and is inconsistent with the behaviour of virtually every other macOS app.

### DIFF
--- a/index.js
+++ b/index.js
@@ -203,12 +203,7 @@ function createMainWindow() {
 
 			// Workaround for electron/electron#10023
 			win.blur();
-
-			if (process.platform === 'darwin') {
-				app.hide();
-			} else {
-				win.hide();
-			}
+			win.hide();
 		}
 	});
 


sindresorhus:
👍

Issue #619, opened 2018-10-25
Improve dark mode skin - Decline message


rtpHarry:
Accidentally clicked on a decline message as I didn't see it:

![image](https://user-images.githubusercontent.com/1038062/47495190-b2f96300-d87d-11e8-928a-919609a2c3e5.png)

Taking the red from here:

![image](https://user-images.githubusercontent.com/1038062/47495099-804f6a80-d87d-11e8-8255-ed1bcd12c73a.png)

This CSS will fix it:

    button._3quh._30yy._17u4._3yvh._1d-z {
        color: #f03d25;
    }

Snippet of the html:

![image](https://user-images.githubusercontent.com/1038062/47495383-35822280-d87e-11e8-88f2-7d8f42141e3c.png)

Preview:

![image](https://user-images.githubusercontent.com/1038062/47495338-1388a000-d87e-11e8-8a03-62644f375719.png)

Can somebody incorporate it into the theme, please?

PR #629, opened 2018-10-28
Fix decline button in dark mode

dusansimic:
Decline button is now using more visible color in dark mode as recommended.

Closes: https://github.com/sindresorhus/caprine/issues/619


sindresorhus:
👌

rtpHarry:
Thanks! I guess I should be making these into PRs directly but they usually crop up in the middle of me working so I figure its better to just submit my notes :)

PR #631, opened 2018-10-28
Change default notification style to banner

janosorcsik:


sindresorhus:
We can actually just drop the config. `banner` is the default.

Issue #392, opened 2018-02-23
Normal theme to light, dark theme too dark(and too blue). Both bad for eyes


mzso:
Hi!

It'd be nice to have something in between that's pleasant for the eyes. The normal theme with the white background is eye-scroching as always, the dark theme is difficult to read. (Also the blue hue is not good for melatonin levels in the evening)

Even better would be if we could just tune the brightness hue with a slider without a need for themes.

dusansimic:
I agree, the dark theme is not really "dark" as you'd expect. I would rather recommend to add a functionality  for custom themes (adding a custom css file). This would make it more customizable.

skymakai:
I like how dark the 'dark' theme is with 'vibrancy' turned off.  However, the text is a gray color, which makes it hard to read on the dark theme.  I'd like the text to be lighter.

elvarb:
The YakYak dark theme is almost perfect, good baseline for a dark theme

https://github.com/yakyak/yakyak

kitor:
I'm missing this (or similar) theme from original m4d - https://userstyles.org/styles/134433/dark-night-facebook-messenger
Indeed ability to just load own css would be great.

sindresorhus:
I come to agree that the current dark mode is too bright and blue too. Happy to accept a pull request that makes it slightly darker and less blue.

sindresorhus:
I'm not going to do anything with the light theme though. It's already enough work to maintain the dark theme.

One way to support custom themes could be to just try reading a certain CSS file in the user's home directory and if it exists, apply it.

Issue #633, opened 2018-10-31
Dark mode + vibrancy broken on app restart


neegool:
When I toggle dark mode + vibrancy while the app is open, it looks like this:
<img width="641" alt="screen shot 2018-10-31 at 3 49 55 pm" src="https://user-images.githubusercontent.com/2795476/47773373-ce081f00-dd24-11e8-8064-e4ccf7a78747.png">

But when I restarted the app, it looks like this:
<img width="634" alt="screen shot 2018-10-31 at 3 49 41 pm" src="https://user-images.githubusercontent.com/2795476/47773399-dd876800-dd24-11e8-9314-4301b465a2b5.png">

re-toggling Vibrancy fixes the problem. The problem does not appear to affect Light mode.

I'm currently using 2.21.0.

sindresorhus:
// @phoomparin

neegool:
**UPDATE:** I've identified that this problem is being caused by the workaround for the white flash on startup. Commenting out the following line in `browser.js` eliminates the issue:

```
// Prevent flash of white on startup when in dark mode
// TODO: find a CSS-only solution
if (config.get('darkMode')) {
	document.documentElement.style.backgroundColor = '#192633';
}
```

PR #634, opened 2018-10-31
Change dark mode color

dusansimic:
Dark mode color is changed to dark gray and added code for loading custom css file for dark mode from config directory.

The one problem that exists is that it's complicated to get backgorund color for fixing `white flash on startup` issue. Any ideas on that or we'll just have to get the color with regex. If https://github.com/sindresorhus/caprine/issues/633 gets pulled there won't be need for implementing this though.

Closes: https://github.com/sindresorhus/caprine/issues/392


index.js
```diff
@@ -270,7 +270,11 @@ app.on('ready', () => {
 
 	webContents.on('dom-ready', () => {
 		webContents.insertCSS(fs.readFileSync(path.join(__dirname, 'browser.css'), 'utf8'));
-		webContents.insertCSS(fs.readFileSync(path.join(__dirname, 'dark-mode.css'), 'utf8'));
+		try { // Try inserting custom css
+			webContents.insertCSS(fs.readFileSync(path.join(app.getPath('userData'), 'custom-dark-mode.css'), 'utf8'));
```
sindresorhus: It's not meant for just dark mode, it's meant to let you override the defaults. It should be injected no matter what. It also needs to be documented.

dusansimic:
The dark mode screenshot is outdated since the dark mode color is changed. For consistency I recommend we use macOS screenshot.

sindresorhus:
The custom styles should be a separate pull request. It's unrelated.

readme.md
```diff
@@ -75,6 +75,22 @@ On *macOS*, you can toggle the window vibrancy effect in the `View` menu.
...
+
+Put `custom.css` into `~/Library/Application Support/Caprine`.
+
+#### Windows
+
+Put `custom.css` into `%APPDATA%\Caprine`.
+
+#### Linux
+
+Put `custom.css` into `$XDG_CONFIG_HOME/Caprine` or if it doesn't exist into `~/.config/Caprine`
```
sindresorhus: Instead of having to document the custom paths here, I think we instead should just add a menu item in `Caprine Preferences` called `Custom Styles` that when clicked, opens `custom.css` and creates it first if it doesn't exist. Then we can just document to click this item. The custom.css file should contain all the variables from `:root` in `dark-mode.css`, so users can quickly just change the colors if wanted.

readme.md
```diff
@@ -75,6 +75,22 @@ On *macOS*, you can toggle the window vibrancy effect in the `View` menu.
 
 <img src="media/screenshot-vibrancy.jpg" width="1165">
 
+### Custom styles
```
sindresorhus: This should be mentioned in the https://github.com/sindresorhus/caprine#highlights section.

readme.md
sindresorhus: > and editing css file there won't be an easy way for disabling the feature

You just comment out the custom styles (we can make that clear as a comment in the default `custom.css` file). I would really prefer not to over-design this. It's just meant as a nice feature for power-users.

dusansimic:
@sindresorhus if you're satisfied with the dark mode colors I will fix the merge conflict and pr will be finished.

sindresorhus:
Yes, the colors look good.

dusansimic:
Rebased 👍

browser.js
```diff
@@ -438,8 +438,8 @@ document.addEventListener('DOMContentLoaded', () => {
 
 	// Prevent flash of white on startup when in dark mode
 	// TODO: find a CSS-only solution
-	if (config.get('darkMode') && !config.get('vibrancy')) {
-		document.documentElement.style.backgroundColor = '#192633';
+	if (config.get('darkMode')) {
```
sindresorhus: ?

https://github.com/sindresorhus/caprine/commit/ec5d74d0ecba1c4deccc9233841b097d2995faec

PR #655, opened 2018-11-14
Add FB Messenger notification sound

### LINKED ISSUE #201: Customize the notification sound
I think it would be great to customize the incoming messages tone, because on Macs it sounds like an app just installed. Giving more customization options could be a huge opportunity for caprine.

clintandrewhall:
Fixes #201.

This adds the correct FB messenger sound on a notification.  I downloaded and added the ringtone, as well, but wasn't able to determine how to toggle it based on the notification.


sindresorhus:
I don't think we should add `ringtone.mp3` until it's actually used.

sindresorhus:
It also needs to respect the `Sound` toggle in the Messenger settings.

sindresorhus:
Closing for lack of response.

PR #660, opened 2018-11-20
After downloading quit and install the update

janosorcsik:


sindresorhus:
No thanks. I don't want to bother users with that. Updates should just happen seamlessly like Chrome.

PR #664, opened 2018-11-26
Fix unread count for muted chats

dusansimic:
Muted conversations are included into unread messages count. This PR will exclude all muted conversations from the count and only include unread and unmuted ones.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -340,7 +340,7 @@ async function sendConversationList() {
 				return {
 					label: el.querySelector('._1ht6').textContent,
 					selected: el.classList.contains('_1ht2'),
-					unread: el.classList.contains('_1ht3'),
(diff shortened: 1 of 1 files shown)


browser.js
```diff
@@ -340,7 +340,7 @@ async function sendConversationList() {
 				return {
 					label: el.querySelector('._1ht6').textContent,
 					selected: el.classList.contains('_1ht2'),
-					unread: el.classList.contains('_1ht3'),
+					unread: el.classList.contains('_1ht3') && !el.classList.contains('_569x'),
```
sindresorhus: It's not clear from reading the code what the class `_569x` means. Move `el.classList.contains('_569x')` into a variable with a descriptive name.

PR #667, opened 2018-11-28
Make sidebar vibrancy the default and add opt-in full-window vibrancy

### LINKED ISSUE #656: More Mojave-like dark theme
screenshot 2018-11-15 at 11 05 35

MacOS Mojave includes a Dark Theme which is (IMHO) much better than Caprine's current theme. I understand that some might prefer the current theme, so I think you should introduce a third, more OS-native feeling theme.

I think two main problems are visible with the current theme:

- The currently selected conversation is not highlighted enough, it's barely visible
- Either nothing or the whole window can be set "vibrant". I think only the sidebar should be transparent, the message view should remain opaque.

Beside this, Caprine's dark theme uses a nice blue-ish color tint, which also adds an odd, akward-look for the app. It stands out from the rest of the dark apps.

What do you think about it?

neegool:
Fixes #656.

The Vibrancy toggle in the menu has been replaced by a new submenu wherein the user can select among three different vibrancy modes:

Disabled: turns off vibrancy
Native-like: only the left sidebar has vibrancy enabled
Full-window: the whole window has vibrancy enabled

It now pulls style and vibrancy settings from `vibrancy-config`


sindresorhus:
I think you're over-complicating it. Instead, I would just rename the current vibrancy mode menu item to `Full-Window Vibrancy` and make the default native-like with sidebar vibrancy.

neegool:
What approach would be better in your opinion: utilize os-specific css (using `.os-darwin`) to implement the default vibrancy (native-like), or injecting a new css conditionally when the OS is detected as macOS?

neegool:
I deferred using the `.os-darwin` approach as I'm having problems with cascading the styles properly. `fd6a7aa` implements conditional injection of `vibrancy-native.css`, only happening if macOS is detected.

config.js
```diff
@@ -4,7 +4,7 @@ const Store = require('electron-store');
 module.exports = new Store({
 	defaults: {
 		darkMode: false,
-		vibrancy: false,
+		vibrancy: 'disabled',
```
sindresorhus: This can remain unchanged as we only have two modes now.

index.js
```diff
@@ -275,6 +275,9 @@ function createMainWindow() {
 	webContents.on('dom-ready', () => {
 		webContents.insertCSS(fs.readFileSync(path.join(__dirname, 'browser.css'), 'utf8'));
 		webContents.insertCSS(fs.readFileSync(path.join(__dirname, 'dark-mode.css'), 'utf8'));
+		if (is.macos) {
+			webContents.insertCSS(fs.readFileSync(path.join(__dirname, 'vibrancy-native.css'), 'utf8'));
+		}
```
sindresorhus: This is also moot. Just move the sidebar vibrancy stuff to browser.css

sindresorhus:

sindresorhus:
https://github.com/sindresorhus/caprine/pull/667#discussion_r241989867 also applies to `vibrancy.css`.

browser.css
```diff
@@ -390,7 +390,54 @@ a,
 	color: rgba(255, 255, 255, 0.8);
 }
 
-/* Remove top Facebook cookie banner */
```
sindresorhus: ?

vibrancy.css
```diff
@@ -64,3 +65,17 @@ html.dark-mode.vibrancy button {
 html.dark-mode.vibrancy ._5iwm ._58al {
 	background: rgba(255, 255, 255, 0.1) !important;
 }
+
+/* Message container + right sidebar */
+html.vibrancy ._4_j4,
+html.vibrancy ._4_j5,
+html.dark-mode.vibrancy ._4_j4,
+html.dark-mode.vibrancy ._4_j5 {
```
sindresorhus: These two selectors are moot.

neegool: I find these selectors necessary, since with the way I implemented sidebar vibrancy, the body must be transparent, and therefore the message div needs to be re-styled to be an opaque color.

sindresorhus:
Thank you :)

Issue #625, opened 2018-10-27
Automatically disable hardware acceleration when on battery


sindresorhus:
Requires https://github.com/electron/electron/issues/1615

whitecrownclown:
@sindresorhus https://github.com/electron/electron/issues/1615 seems to be solved. I'm getting the `on-battery` event, however, how would we disable hardware acceleration without restarting `Caprine` afterwards? Do we just prompt the user with a message and then let him decide?

sindresorhus:
> how would we disable hardware acceleration without restarting `Caprine` afterwards?

Someone needs to open an issue on [Electron](https://github.com/electron/electron) requesting the ability to change hardware acceleration without having to restart. Prompting would be annoying and not much better than just manually changing and we can't automatically change as the user might be in the middle of writing a message.

sindresorhus:
Electron declined https://github.com/electron/electron/issues/17011, so I don't see this being feasible to implement.

PR #676, opened 2018-12-10
Added  automatic power state setting

killbom:
This is a PR for #625 

In order to change the hardware acceleration electron needs to reload. It is no very pretty as the whole window needs to close. I think it would be better if electron could "soft" reload or so. 

Not sure if this should even be a PR :/

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,6 +30,22 @@ if (!config.get('hardwareAcceleration')) {
 	app.disableHardwareAcceleration();
 }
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't think this would work well as the user might be in the middle of typing a message, then unplugs the power adapter accidentally and Caprine restarts and loses the message.

I would open an issue on the Electron issue tracker about ability to change the hardware acceleration thing without having to restart the app.

PR #697, opened 2019-01-07
Add setting to follow system dark mode on macOS

### LINKED ISSUE #510: Follow system dark mode
macOS Mojave will have a system-wide dark mode. It would be nice to have an option to follow the system theme.

### LINKED ISSUE #694: Electron 4 and vibrancy
In reference to #667, release 2.24.0's electron version now detects the system-wise dark mode status. It can be seen in the sidebar. Selecting Caprine's light mode while using system-wide dark mode makes the sidebar look wrong. It's because internally I've set the sidebar vibrancy mode to `sidebar`, which I think changes based on the state of the system dark mode.

With that being said, I plan on making a new pull request which fixes the issue. We have two options here:

1. *Stick with the old behavior*: app-level dark mode can be set independently of the system-level dark mode. This means that I should avoid using `sidebar` vibrancy profile, and revert to using `light`

2. *Adapt to the system-level dark mode*: This means that we will remove the option to set the app-level dark mode.

I'd like to know your thoughts regarding the direction you want the app to move towards.

neegool:
Fixes #694 
Fixes #510


index.js
```diff
@@ -160,6 +161,16 @@ function setUserLocale() {
 	electron.session.defaultSession.cookies.set(cookie, () => {});
 }
 
+function setUpSystemPreferencesListener(win) {
+	// Listen to system-wide appearance changes (macOS)
+	const id = systemPreferences.subscribeNotification('AppleInterfaceThemeChangedNotification', () => {
+		console.log('changed', systemPreferences.isDarkMode());
+		win.webContents.send('set-dark-mode');
+	});
```
sindresorhus: You can use https://github.com/sindresorhus/electron-util#darkmode to simplify some of the listening code here.

index.js
```diff
@@ -160,6 +161,16 @@ function setUserLocale() {
 	electron.session.defaultSession.cookies.set(cookie, () => {});
 }
 
+function setUpSystemPreferencesListener(win) {
+	// Listen to system-wide appearance changes (macOS)
+	const id = systemPreferences.subscribeNotification('AppleInterfaceThemeChangedNotification', () => {
+		console.log('changed', systemPreferences.isDarkMode());
```
sindresorhus: Leftover

sindresorhus:
There’s no need to unsubscribe on quit. The system does that for us then.

sindresorhus:
Looks good. Thank you! :)

sindresorhus:
@neegool You've done some good PRs. Would you be interested in joining the project as a maintainer? No worries if not though.

PR #706, opened 2019-01-09
Crop label on Touch Bar button

### LINKED ISSUE #705: Crop chat names on Touchbar
touch bar shot 2019-01-09 at 11 22 52

Sometimes chat names can be very long. I think you should just crop them to ~25-30 characters.

janosorcsik:
Fix #705


CvX:
May I suggest using an ellipsis ("…" U+2026) instead of three full stops ("...")?

Minor difference but it's slightly more visually pleasing and in some cases it could squeeze in enough horizontal space for an additional label that otherwise wouldn't fit. 😉

Issue #704, opened 2019-01-09
v4.5.10: A JavaScript error occurred in the main process


yermulnik:
```
Uncaught Exception:
TypeError: systemPreferences.setAppLevelAppearance is not a function
    at EventEmitter.ipcMain.on (/opt/Caprine/resources/app.asar/index.js:382:21)
    at EventEmitter.emit (events.js:182:13)
    at WebContents.<anonymous> (/opt/Caprine/resources/electron.asar/browser/api/web-contents.js:334:13)
    at WebContents.emit (events.js:182:13)
```
Linux Mint 17.3 Rosa (Ubuntu 14.04 based).
Previous version works fine for me.

sindresorhus:
// @neegool

rtsp:
Same problem on Windows 7 (On launch and reload)

![Caprine Error](https://user-images.githubusercontent.com/314429/50914848-17ce5d80-146a-11e9-8595-a1c8ed653c51.png)

petko:
Same problem on Windows 10..

malyzeli:
Same here on Ubuntu 18.10, installed via Snap

![image](https://user-images.githubusercontent.com/10441355/50932261-5a416b80-1465-11e9-9b1b-45eb1cc811db.png)

malyzeli:
I noticed Facebook updated Messenger appearance (and maybe also some functionality?) the same time this error started happening. 

I guess it might be related.

sindresorhus:
Note this part:

<img width="434" alt="screen shot 2019-01-10 at 08 45 47" src="https://user-images.githubusercontent.com/170270/50940598-23467680-14b4-11e9-88b7-15b272ccaa36.png">

zhartus:
Seems that problem returns after last update

adoster:
Windows 10 64-bit:

A potential solution I found was having an existing and working installation of Caprine (2.36) and then updating to 2.44. I tested it out going from 2.36 to 2.37, to 2.38... etc. first, so it may be the case that you need some combination or collection of pre-2.44 versions of Caprine working first (not necessarily *any* working version).

PR #708, opened 2019-01-09
Add OS check before calling setAppLevelAppearance

neegool:
This is a fix for #704


index.js
```diff
@@ -376,7 +376,7 @@ function createMainWindow() {
 
 ipcMain.on('set-vibrancy', () => {
 	mainWindow.setVibrancy('sidebar');
-	if (config.get('followSystemAppearance')) {
+	if (is.macos && config.get('followSystemAppearance')) {
```
sindresorhus: This is not correct. `setAppLevelAppearance` will still be called now when `followSystemAppearance` is `false`.

PR #716, opened 2019-01-12
Hide the preferences dialog during the automated action

CvX:
Gets rid of the preferences dialog flash on startup and when changing the “Mute Notifications” setting.

**Application startup**
before/after
before after


sindresorhus:
Yay! That flash was really annoying :)

PR #717, opened 2019-01-13
Fix notifications

### LINKED ISSUE #700: Clicking on notifications doesn't open window on macOS
Since v2.24 clicking on system notifications when Caprine's window is closed (with ⌘W) does not do anything. It should (and it used to) open and bring Caprine to the front, then focus the selected conversation.

CvX:
Electron’s context isolation is preventing any `window.Notification` modification done by the preload script (`browser.js`).

This change works around that limitation by first injecting custom `window.Notification` in the main process using `webContents.executeJavaScript` and then by passing messages between browser context (Messenger code) and the Isolated Context (Caprine’s `browser.js`) via `window.postMessage/window.addEventListener('message', …)`.

As a bonus, this PR adds support for `onclick` and `onclose` Notification events used by Messenger. That allows us to remove `selectConversationByImg`, since Messenger selects the conversation in the `onclick` callback! 😃 

Fixes #700

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -502,9 +502,11 @@ document.addEventListener('keydown', event => {
 	}
 });
 
-window.Notification = (notification => {
-	const customNotification = function (title, options) {
-		let {body, icon, silent} = options;
+// Pass events sent via `window.postMessage` on to the main process
+window.addEventListener('message', ({data: {type, data}}) => {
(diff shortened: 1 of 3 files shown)


sindresorhus:
Wow. Nice work! This is a very clever way to solve it 🙌

PR #720, opened 2019-01-14
Add emoji style setting

mikigal:
Add option to use old Messenger emoji. You can enable it in options.


CvX:
1. Emoji replacement shouldn't really be in `setUpPrivacyBlocking`. That code should have its own function instead.
2. The URL replacement algorithm is too complex because it not only changes the emoji style but also its size. If we keep the original size/pixel ratio (and I think we should), then the replacement comes down to just a single letter change, e.g. https://static.xx.fbcdn.net/images/emoji.php/v9/tf8/2/32/1f61b.png becomes https://static.xx.fbcdn.net/images/emoji.php/v9/zf8/2/32/1f61b.png ('t' -> 'z')
3. As far as I know there are three styles available: 't' (the current default), 'z', and 'f' (e.g. https://static.xx.fbcdn.net/images/emoji.php/v9/ff8/2/32/1f61b.png). Should we offer all three?

mikigal:
I think we can offer thee types of emoji, user will can select it in menu.
I will follow your suggestions.

mikigal:
I've created menu with 3 emoji options, separated emoji code from `setUpPrivacyBlocking`. I haven't idea to name options so I simply name it 'F style', etc.

CvX:
@mikigal I've added more fixed and slightly reorganized the code. Please take a look if everything looks good to you. 🙂

Before we merge, I have some questions:
1. Is `*://*.${domain}/images/emoji.php/v9` needed here? https://github.com/sindresorhus/caprine/blob/f4eabacf112189ac7584dbe66f8b7caa6a8075c9/index.js#L151
2. We probably should handle in some way emoji that isn't available in older emoji sets, right? Currently using the latest emoji ends up like this: screen shot 2019-01-16 at 19 04 39
3. This one is to @sindresorhus – should menu options like this be all placed in "Caprine > Caprine Preferences"? I feel like at least some of them should be moved to e.g. "View" or "Window".

mikigal:
1. Emoji can be fetched from facebook.com too. I'm not sure when it fetch from static.xx.fbcdn.net or facebook.com. It can't be downloaded from messenger.com. I replace {domain} for facebook.com for additional safety.

2. I created exclude list. It is needed only for Messenger 1.0 version. For these emoji we don’t do anything and give emoji from newest version. For Facebook 2.2 everything seems to work.
I changed default emoji style to the newest version too.

I think everything is correct now :)

mikigal:
It seem's it can be only one `webRequest.onBeforeRequest`. It seems second onBeforeRequest override first, and privacy blocking didn't worked, because emoji override it. I worked around it by creating one onBeforeRequest that trigger privacy blocking and emoji methods. Now everything works :)

CvX:
You're right about `webRequest.onBeforeRequest`! I just assumed it allows registering multiple listeners… Should have read the docs! 😄

I'm pushing more changes:
- Added more emoji codes to the excludes list (Facebook 3.0 set introduced 161 new emojis), removed duplicates and sorted it alphanumerically
- cleaned up the code and extracted it again to the separate file (who knew it would grow so much 😜)

Let me know if I broke anything again. 😉

Also, two more questions:
1. What's the reason for these changes? https://github.com/sindresorhus/caprine/blob/f07aef98af2960181e4ab14286d5fb6be066bbea/index.js#L426-L436
2. Since changing active emoji set requires app restart, should we display a message box offering a restart?

CvX:
Oh, and because of all those new emoji I've enabled emoji replacement for Facebook 2.2 set. Without it, the support is kinda spotty:
screen shot 2019-01-16 at 23 20 35

mikigal:
I don't know why, but I haven't any problems with emoji from your screenshot. Please check by devtools(F12) from what URL it tries download.

nkj22ushr1

CvX:
Oh, they do work now (when using fallback). I'm just saying I've enabled the excludes list for Facebook 2.2 set because these were previously missing in e.g. 8f961df and f07aef9. Or did these work for you before my recent commits?

mikigal:
It worked for me, without exclude list for Facebook 2.2. But I found something interesting. It looks, that facebook automatically redirect some emoji from FB 2.2 (I think all missing for this set) to FB 3.0. It's why I made exclude list only for 1.0
i8mq7tsvgo

CvX:
That's how our replacement shows up in DevTools. 🙂

mikigal:
Not every! There are a lot of downloaded emoji without any redirect, but from selected set.
pas7wxxf80

CvX:
I'm guessing the first request for given emoji goes though the redirect and subsequent have the new URL cached.

mikigal:
Maybe you are right. But what about exclude for FB2.0? Try disable it for 2.0 and try does it work for you

CvX:
I did try fb2.0 without the excludes list. That how I took that screenshot. 🙂
Just try the toothmoji 😉 in the browser:
t: https://static.xx.fbcdn.net/images/emoji.php/v9/t7f/2/32/1f9b7.png 
f: https://static.xx.fbcdn.net/images/emoji.php/v9/f7f/2/32/1f9b7.png 
z: https://static.xx.fbcdn.net/images/emoji.php/v9/z7f/2/32/1f9b7.png 
No redirects there.

mikigal:
Okay, you are right about redirects ;)
So excludes for FB2.0 will fix issues about missing emoji. I tried creating modal for restart but I'm not sure how to style it. Can you create this modal?

CvX:
How about this? 
restart dialog

(it should use the app's icon after packaging)

mikigal:
Nice! God job :)

menu.js
```diff
@@ -145,6 +196,19 @@ const preferencesSubmenu = [
...
+				detail: 'Do you want to restart the app now?',
+				buttons: ['Restart', 'Ignore'],
+				defaultId: 0,
+				cancelId: 1
+			}, response => {
+				if (response === 0) {
+					app.relaunch();
+					app.exit();
+				}
+			});
```
sindresorhus: This is almost identical to the other emoji restart prompt? Can you move it to a function that accepts a message argument? As that's the only unique thing about them.

menu.js
```diff
@@ -194,6 +258,10 @@ const preferencesSubmenu = [
 			config.set('quitOnWindowClose', !config.get('quitOnWindowClose'));
 		}
 	},
+	{
+		label: 'Emoji style (requires restart)',
```
sindresorhus: Since we have a prompt now, we don't need the `(requires restart)` part. Same for the hardware acceleration menu.

menu.js
```diff
@@ -23,6 +23,57 @@ const newConversationItem = {
...
+		type: 'checkbox',
+		checked: (config.get('emojiStyle') === 'f'),
+		click() {
+			handleEmojiClick('f');
+		}
+	}
+];
+
+function handleEmojiClick(type) {
+	config.set('emojiStyle', type);
```
sindresorhus: The config should use a more descriptive value. I would suggest:

- `facebook-3-0`
- `messenger-1-0`
- `facebook-2-2`

menu.js
```diff
@@ -23,6 +23,57 @@ const newConversationItem = {
...
+
+	// Update menu
+	emojiSubmenu[0].checked = (config.get('emojiStyle') === 'f');
+	emojiSubmenu[1].checked = (config.get('emojiStyle') === 't');
+	emojiSubmenu[2].checked = (config.get('emojiStyle') === 'z');
+	menu = electron.Menu.buildFromTemplate(template);
+	electron.Menu.setApplicationMenu(menu);
```
sindresorhus: These two lines should be abstracted to a `updateMenu`-kinda method.

menu.js
```diff
@@ -23,6 +23,57 @@ const newConversationItem = {
 	}
 };
 
+const emojiSubmenu = [
+	{
+		label: 'Facebook 3.0',
+		type: 'checkbox',
```
sindresorhus: Would be nice to have an `icon` here showing the emoji. It's impossible for a user to know what  `Facebook 3.0` and the other names mean. If they saw the same emoji in different styles, they would know.

emoji.js
```diff
@@ -0,0 +1,213 @@
...
+	'1f9fb',
+	'1f9fc',
+	'1f9fd',
+	'1f9fe',
+	'1f9ff',
+	'265f',
+	'267e'
+]);
+
+module.exports = function (url, callback) {
```
sindresorhus: Use arrow function

emoji.js
```diff
@@ -0,0 +1,213 @@
...
+		return callback({});
+	}
+
+	// Messenger 1.0 and Facebook 2.2 emoji sets support only emoji up to version 5.0.
+	// Fall back to default style for emoji >= 10.0
+	if (excludedEmoji.has(emojiCode)) {
+		return callback({});
+	}
+
+	const newURL = url.replace(/(emoji\.php\/v9\/)(.)(.+\/)/, `$1${emojiStyle}$3`) + '#replaced';
```
sindresorhus: Maybe use named groups for readability? http://2ality.com/2017/05/regexp-named-capture-groups.html

index.js
```diff
@@ -208,7 +222,7 @@ function createMainWindow() {
 	});
 
 	setUserLocale();
-	setUpPrivacyBlocking();
+	setUpRequestsFiltering();
```
sindresorhus: `setUp` => `init`

emoji.js
```diff
@@ -0,0 +1,213 @@
...
+	const emojiCode = url.substring(url.lastIndexOf('/') + 1, codeEnd);
+
+	if (emojiStyle === 't' || url.includes('#replaced') || codeEnd === -1) {
+		return callback({});
+	}
+
+	// Messenger 1.0 and Facebook 2.2 emoji sets support only emoji up to version 5.0.
+	// Fall back to default style for emoji >= 10.0
+	if (excludedEmoji.has(emojiCode)) {
+		return callback({});
```
sindresorhus: Would be nicer if this just returned the object and the callback handling was done in https://github.com/sindresorhus/caprine/pull/720/files#diff-168726dbe96b3ce427e7fedce31bb0bcR151

index.js
```diff
@@ -135,20 +136,33 @@ function enableHiresResources() {
...
-		let blocking = false;
-		if (details.url.includes('typ.php')) {
-			blocking = config.get('block.typingIndicator');
+function setUpRequestsFiltering() {
+	const filter = {
+		urls: [
+			`*://*.${domain}/*typ.php*`,
+			`*://*.${domain}/*change_read_status.php*`,
+			'*://static.xx.fbcdn.net/images/emoji.php/v9/*',
+			'*://facebook.com/images/emoji.php/v9/*'
```
sindresorhus: While it's fairly clear what these URLs are for, I think it would be nice to add comments to them to make it perfectly clear.

emoji.js
```diff
@@ -0,0 +1,213 @@
+'use strict';
+const config = require('./config');
+
+const excludedEmoji = new Set([
```
sindresorhus: Maybe add a comment on how you discovered this in case we need to update this list in the future.

menu.js
```diff
@@ -23,6 +23,59 @@ const newConversationItem = {
...
+		}
+	});
+}
+
+function updateEmojiMenu() {
+	emojiSubmenu[0].checked = (config.get('emojiStyle') === 'facebook-2-2');
+	emojiSubmenu[1].checked = (config.get('emojiStyle') === 'facebook-3-0');
+	emojiSubmenu[2].checked = (config.get('emojiStyle') === 'messenger-1-0');
+	menu = electron.Menu.buildFromTemplate(template);
+	electron.Menu.setApplicationMenu(menu);
```
sindresorhus: Note my comment in https://github.com/sindresorhus/caprine/pull/720#discussion_r248530171. Only the two lines (75 and 76).

menu.js
CvX: Done:
icons in emoji submenu

mikigal:
@CvX Can you add comment how to get all excludes?

CvX:
All requested changes have been made. (shouldn't Github automatically mark the review green?)

Though I have a question to @sindresorhus that got lost in the discussion:

> 3\. should menu options like this be all placed in "Caprine > Caprine Preferences"? I feel like at least some of them should be moved to e.g. "View" or "Window".

Anyway, that's a non-blocker. LGTM 🙂

sindresorhus:
> 3. should menu options like this be all placed in "Caprine > Caprine Preferences"? I feel like at least some of them should be moved to e.g. "View" or "Window".

Yeah, I missed that. No, Preferences is correct. "View" and "Window" should only contain transient and context-sensitive actions, meaning actions you do often and that are not meant to be permanent. You could argue the Vibrancy thing should be in Preferences too.

sindresorhus:

config.js
```diff
@@ -21,6 +21,7 @@ module.exports = new Store({
 			chatSeen: false,
 			typingIndicator: false
 		},
+		emojiStyle: 't',
```
sindresorhus: This should use the ID

menu.js
```diff
@@ -23,6 +42,51 @@ const newConversationItem = {
 	}
 };
 
+const emojiSubmenu = [
```
sindresorhus: Maybe we should move this and other related logic to `module.exports = {emoji}` in `emoji.js` to contain all the logic for this feature in one file?

CvX: That was a good idea! Done. Let me know how do you like the new code organization.

Issue #316, opened 2017-11-02
Bottom text field gets replaced with "Could not display composer." if its input is not active


the4kman:
I am using Caprine 2.7.1 on Mac. If the text input on the bottom is not in focus and I press an alphanumeric key on my keyboard, the input gets replaced with an area that says "Could not display composer." and it does not recover from that state until I refresh Caprine (Cmd+R). It does not disappear either when I switch between conversations.

<img width="507" alt="screen shot 2017-11-02 at 16 00 25" src="https://user-images.githubusercontent.com/6132647/32333149-6ce02bec-bfe7-11e7-8bd4-a2b60eed18eb.png">

BTW, you can blur the text input by opening an image in a conversation and dismissing it by clicking on the area where the right sidebar is behind the black overlay.

hckuo:
+1

massayoshi:
Here usually happens when I press CMD

happyearmuffs:
I've been able to reproduce by follow these steps:
- Click a link in a thread, focus is drawn to browser
- click back into thread body
- press any key

emileber:
I confirm that with these steps, I got the problem on my end as well.

effektsvk:
I can confirm as well but with a little note. You have to click back into thread body when Caprine window is not in focus (don't CMD-Tab back into it). That click into thread body will activate focus on the window.

emileber:
- Opening an image, 
- closing it 
- and typing 

will also cause the problem.

emileber:
It looks like the focus is always in the _composer_ box by default, and when the focus isn't on a valid element (search box, composer box, etc.), like when you open a link or an image and get back, it fails.

Vilscon:
Yet.. there isn't a solution for this? :s

effektsvk:
Except refreshing Caprine when it happens, no. :(

Vilscon:
Hm, I can live with that. Normally I force-quit the app and restart it. Apparently CMD+shift+R works as well.

effektsvk:
Just CMD + R should work ;)

veniversum:
This seems to have been fixed on FB's end -- is it still reproducible?

hron84:
Not with Caprine, but with MessengerForDesktop, but sometimes this still happens. I guess this is something that Facebook should improve.

notflip:
Has anyone been able to reproduce this?

PR #726, opened 2019-01-15
Add support for replying via notifications on macOS

### LINKED ISSUE #298: Reply via notification [macOS]

JoniVR:
Hi, 

First of all thanks for this great application.

Fixes #298.

## Some things:
* If you do consider merging, some testing is required on Windows and Linux as I do not currently have access to these devices. The inline reply should (I think) only work on macOS as that is the only platform that seems to be supported ([see documentation](https://electronjs.org/docs/api/notification)) but correct me if I'm wrong.
* I'm not an expert at electron apps, so forgive any things I overlooked, missed or did wrong, and let me know, I'll fix them. 
* If you have any further questions about implementation I'd be happy to answer them. 

Also, I made this the default, not configurable yet (since I thought it didn't make much difference in the user experience), but if you want me to I can add it as an extra option.


CvX:
Alright, I'm testing this change, here's what I found:

1. The reply isn't sent when Caprine's window is closed
2. Same if, window is visible but a different conversation is active
3. If the reply contains any emoji it sends only the first emoji (no text before or after it, or any later emojis)
4. Using `setTimeout` to wait for the interface is bound to break often. I'd suggest to move this: https://github.com/sindresorhus/caprine/blob/0f00f08ac87b6e91d539c165a2c43fead1d43378/browser.js#L559-L566 to https://github.com/sindresorhus/caprine/blob/0f00f08ac87b6e91d539c165a2c43fead1d43378/notifications-isolated.js#L4-L18 probably as a new `type`, e.g. `notification-reply` (just like we use `notification-callback` in two contexts). I'm not 100% sure, but I would hope that Messenger switches to to the active conversation synchronously in the `onclick` handler, which means that moving the reply-typing code here would solve the timing issue. Of course then, the switching back to the previous conversation is also getting more complex, since it has to be split between 'notifications-isolated.js' and 'browser.js' (i.e. after calling `onclick` and typing the reply, it would `window.postMessage()` back to the browser context where `selectConversation ` can be called).

I realize that number 4 is a large task. Let's say that after 1-3 are fixed, we'll just try it out and see if the current code is reliable enough. Or maybe doing the 4 will be necessary to fix previous issues… 🤷‍♂️😅

JoniVR:
Hi @CvX 

Thanks for looking into it, I'll fix 1-3 and then look into 4 again, I really struggled with nr. 4. 

I wanted to use the callback but couldn't get it to work without doing hacky stuff, so I implemented the least hacky solution that would work, but I'll look into it again. I should be able to get it to work because as you said, `setTimeout()` is not really a reliable solution.

JoniVR:
> 1. The reply isn't sent when Caprine's window is closed
> 2. Same if, window is visible but a different conversation is active

I just tested these again and they seem to work on my end, with "the window closed", you mean when you press the (x) button to just close the window right? Does the text field contain the message you intended to send?

CvX:
Hm, can't reproduce it now either. Maybe because of the emoji issue I mistook sent message for an earlier one…

JoniVR:
Okay so the emoji thing is a though one. 
I originally tried to implement it using:
```js
document.querySelector('div[contenteditable="true"]').focus();
document.execCommand('insertText', false, 'message');
```
but then it would display:

screenshot 2019-01-17 at 12 55 20


So I used:
```js
const event = document.createEvent('TextEvent');
event.initTextEvent('textInput', true, true, window, message, 0, locale);
```
instead. However, this is causing issues with the emoji being pasted (and this function is deprecated, so shouldn't be used anyways). 
I originally thought the "Could not display composer." message had something to do with the locale, and so I switched to the other method which allowed me to set a locale, but it doesn't seem like that's the issue. 

I've seen the error message on the internet on numerous forums, but it's most of the times related to messenger itself or an extension. I has also been mention in #316. Does anyone else have an idea? I'll keep looking into it.

JoniVR:
Thanks for looking into it! That would certainly work and is a great workaround!

Another solution to this that I thought of would be to quickly switch between your previous convo and back, because then the error goes away and the inputted text is still present and ready to be sent, this solution is equally hacky but doesn't make use of the deprecated API (and I haven't tested how reliable it is yet). 

I was still looking for another way but so far didn't have any luck (even posted a [stackoverflow question](https://stackoverflow.com/questions/54256517/make-document-execcommandinserttext-false-message-work-with-draftjs)). 

> If you have an unsent message typed in, the app should save that message, do the reply logic, and then restore that message.

I agree, that's a good idea, if it's possible to implement it 😅

@sindresorhus Any thoughts on this issue? Should we just implement a somewhat hacky solution?

CvX:
The conversation-switcheroo is a nice solution. It is certainly much cleaner and more straightforward than using deprecated events and a couple of text-manipulating commands. There are two downsides that come to mind, though:

1. what if you have just a single conversation? 😃
1. a scenario where Caprine's window is present (but not in focus) – the conversation switch would be visible

Restoring the previous message is possible, just did a quick test. The bottom line is: `document.querySelector('div[contenteditable="true"]').textContent` is just enough.
Here's the full example:

```js
const messageInput = document.querySelector('div[contenteditable="true"]');
let previousMessage = messageInput.textContent;

messageInput.focus();

if (previousMessage) {
  document.execCommand('selectAll', false, null);
  document.execCommand('delete', false, null);
}

document.execCommand('insertText', true, message);
nextConversation();
previousConversation();
sendMessage();

if (previousMessage) {
  document.execCommand('insertText', true, previousMessage);
  nextConversation();
  previousConversation();
}
```

JoniVR:
Yeah, I mean it's not the most elegant solution to use a deprecated API but I don't see any other way to currently get it done otherwise. So should I just implement it with your workaround then?

CvX:
Yeah, I suppose so. We can always change it if we come up with a better one. 🙂

JoniVR:
Alright, I'll take a stab at implementing it tomorrow, thanks for the help with this.

JoniVR:
Hi @CvX 
Quick question.	
So now to move it to `notifications-isolated.js` I should use: 
```js
ipc.on('notification-reply', (event, data) => {
	const previousConversation = getIndex();
	window.postMessage({type: 'notification-reply', data}, '*');
	window.addEventListener('message', async ({data: {type}}) => {
		if (type === 'reply-callback') {
			await sendReply(data.reply);
			if (previousConversation) {
				selectConversation(previousConversation);
			}
		}
	});
});
```
inside `browser.js`
and post back a message back from `notifications-isolated.js` to the browser:
```js
if (type === 'notification-reply') {
	window.postMessage({type: 'reply-callback'}, '*');
}
```
like that right? Or am I missing something?

CvX:
Yeah, but you'll have to move the `window.addEventListener` out of the IPC handler (otherwise you'd be adding a new listener after each reply). 🙂

JoniVR:
It's probably going to be an issue to switch to the previous conversation then, unless we send it back and forth through `postMessage` I guess.. or save it globally, but I think it'd be better to send it back and forth then.

CvX:
I'm currently testing this. I've found some issues, but they're not directly caused by this PR:

1. You reply via notification. The contact responds. Messenger doesn't trigger a notification at all, probably because it assumes you're active, in the app. This makes you miss the message (though there's the unread badge on app's icon in the Dock) and you can't continue the conversation through notifications.
  It may be possible to fix this. Messenger reacts to some presence events fired by the browser (`window.onblur`, [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API), etc.), so we could trigger those after sending the reply.
2. There's a bug in Messenger, when you switch to a conversation with a bot, sometimes the message composer is missing:
  missing composer composer is there
  When it's gone, the reply obviously doesn't go through.
3. Sometimes, the Messenger does fire a notification, but due to a bug (most likely in Electron) it doesn't appear on the screen and goes straight to the Notification Center. I'm gonna look into it.
4. There are also cases where you type a reply, click send or hit enter, and then… nothing. The notification disappears but `notification.on('reply')` isn't called. Neither is `notification.on('click')` nor `notification.on('close')`. That's a bug in either Electron or macOS… I'm going to look deeper into this one too.

CvX:
Update on issue 4:

It's a bug in Electron. Turns out [Notification](https://electronjs.org/docs/api/notification) objects are being prematurely garbage collected… so in many cases clicking, replying, or closing a notification doesn't trigger anything because the underlying Notification object (and its callbacks) is gone.
To fix this, Electron has to keep track of notifications that have event listeners, as stated in [Notification Web API spec](https://notifications.spec.whatwg.org/#garbage-collection) (I'll submit an issue on Electron's repo).

I've pushed a workaround to this branch.

About other issues:

1: Gotta fix that, will look into that next.
2: Still happens, but definitely is a non-blocker for this PR. Just a bit annoying when it comes to testing replies. 😉
3: I haven't seen it recently, should be easier to spot after 1 is fixed.

JoniVR:
Hi @CvX 

Any updates on this? Did you create an issue on electron (so I can track it myself)?
I resolved the merge conflicts after typescript migration.
I'm unable to reproduce issue 1. on my end at the moment.

CvX:
@JoniVR Sorry, I've been away for a while. Here's the issue on the Electron repo: https://github.com/electron/electron/issues/16922

About the issue "1.", the best way to reproduce it is to chat with a bot (I've created one to test this PR, but I'm waiting for it to go through FB's review)
- Send it a message
- Hide the window
- Reply to the first message
- Reply to the second message
- Reply to the third message
- Fourth message doesn't produce a notification

JoniVR:
@CvX 	
Yes, I'll be waiting for that bot then, unless you can recommend me a different one? 
I was previously testing it using the Memegenerator bot I saw from your screenshot earlier, but I think it has blocked me or something because my messages don't arrive anymore.

CvX:
Alright, Messenger Ping Bot is live! 😄 Just search for it in the app and send anything.

CvX:
I've pushed a fix: apparently the "Send" button lost one of its CSS classes recently. 😉

I was also investigating the issue with max 3 notifications. Turns out it's a "feature". 😉 Messenger tracks how many unread messages are in each thread and show notification only for the first 3 messages in each one. I haven't found any reliable way to work around it, so I'd say there's no reason to try to fight it and just accept it. 😉

There's a minor edge case that also comes to mind that I don't think I've mentioned before. Let's say you have unread messages in the currently active conversation. You reply via notification to a different conversation. But the first one is marked as seen/read even though you haven't.
Just something to keep in mind. I don't working around that is needed.

And with that, I think this PR is _almost_ ready to be merged? 🙂 
I haven't tested on Windows yet, but I'll be able to check that out tomorrow. 🚀

JoniVR:
> I was also investigating the issue with max 3 notifications. Turns out it's a "feature". 😉 Messenger tracks how many unread messages are in each thread and show notification only for the first 3 messages in each one. I haven't found any reliable way to work around it, so I'd say there's no reason to try to fight it and just accept it. 😉

Alright, makes sense, it was really weird that it did that anyways, couldn't really find a reason why it would happen. 

> There's a minor edge case that also comes to mind that I don't think I've mentioned before. Let's say you have unread messages in the currently active conversation. You reply via notification to a different conversation. But the first one is marked as seen/read even though you haven't.

Would it perhaps be possible to block the "seen indicator" in this case? The app already does this as a setting so perhaps that could be a workaround? I'm just guessing here.

Anyways, thanks a lot for all the help and feedback with this, I've been extremely busy lately so it's very helpful!

CvX:
> Would it perhaps be possible to block the "seen indicator" in this case? The app already does this as a setting so perhaps that could be a workaround? I'm just guessing here.

Yeah, that might be a good solution. 🙂 I'll try it later!
Anyway, that can be added later, after this PR is merged. 🙂

JoniVR:
Sounds good to me, would be great to have this feature! Looking forward 👌🏻

CvX:
Alright, no issues on Windows 10! 🚀

sindresorhus:
Works great for me 👌

Thanks for implementing this, @JoniVR 🙌

JoniVR:
Hi @sindresorhus 

Is it possible that you forgot to add the other assets to the 2.29.0 release? Or is it normal that they don't show up yet?

Thanks for everything.
Best regards,
Joni

sindresorhus:
Sometimes `electron-builder` fails to build the assets and I didn't notice. Fixed: https://github.com/sindresorhus/caprine/releases/tag/v2.29.0

Issue #712, opened 2019-01-10
Vibrancy issue, gray sidebar


hieuhuynh93:
I didn't use Dark mode and Vibrancy mode, but the sidebar is look like dark mode. Please check my screenshot. Thank you.
![Screenshot](https://i.imgur.com/V8dwnZZ.jpg)

dscos:
I am experiencing this same issue, but in reverse: Main window is obeying my dark area selection but the sidebar is light with light text.

cocoJambon01:
I have a similar issue I might say. Turning Dark Mode on will neglect the sidebar and only the message box will be black: http://prntscr.com/m5ttm7

filipruisl:
same here, however the Sidebar is a Dark vibrant theme with dark text on it.

ggrelet:
I have a similar issue, however I don't even have Drak Mode enabled. The white mode's white is simply different in the side bar.
<img width="1072" alt="screenshot_2019-01-12 15-02-25 2x" src="https://user-images.githubusercontent.com/10114131/51069883-263f9380-167b-11e9-96cd-bf73fd74221a.png">

cocoJambon01:
Just installed version 2.26 - Issue persists.

jackwilsdon:
Yeah I can confirm this is still broken on `master`. This patch seems to fix it but vibrancy mode still doesn't look right with dark mode enabled;

```diff
diff --git dark-mode.css dark-mode.css
index 570862b..344a1f2 100644
--- dark-mode.css
+++ dark-mode.css
@@ -764,8 +764,7 @@ html.os-darwin.dark-mode ._673w {
 }
 
 /* Styles for vibrancy in dark mode */
-html.os-darwin.dark-mode body,
-html.os-darwin.dark-mode ._4sp8 {
+html.os-darwin.dark-mode body {
 	background: transparent !important;
 }
 ```

*Dark Mode*
![Dark Mode](https://user-images.githubusercontent.com/1843197/51104343-d3322180-17dd-11e9-9ae9-5f10b5bfe4db.png)

*Dark Mode + Full-Window Vibrancy*
![Dark Mode + Full-Window Vibrancy](https://user-images.githubusercontent.com/1843197/51104342-d3322180-17dd-11e9-9455-eebc3c8adaf5.png)

sindresorhus:
// @CvX

hieuhuynh93:
I confirmed that the issue is still not solve yet. (2.26.0)
[...]

PR #727, opened 2019-01-16
Fix window resizing/moving issues and add ability to disable vibrancy

### LINKED ISSUE #721: Windows 10: window not vissible when toggled
So I'm having a problem when I open Caprine, it shows in taskbar but when I click on taskbar the window is not showing.

### LINKED ISSUE #724: Windows 10: Maximize button greyed out
On windows 10 I cannot maximize window. The button is greyed out. Double clicking on top bar doesn't do anything.

### LINKED ISSUE #725: Windows 10: Window cannot be resized
On windows 10, window of caprine cannot be resized.

CvX:
Fixes #721 
Fixes #724 
Fixes #725 by reverting a faulty fix (my bad).

Also, improves the situation in #712 by providing a way to disable the vibrancy.

vibrancy menu

In a way, this PR acts as a stop-gap before a complete fix for vibrancy issues is done (though some people might like the vibrancy-less look so an option for that is a plus).

cc: @sindresorhus, @neegool


config.js
```diff
@@ -5,7 +5,7 @@ module.exports = new Store({
 	defaults: {
 		followSystemAppearance: false,
 		darkMode: false,
-		vibrancy: false,
+		vibrancy: 'none', // Possible values: 'none', 'sidebar', 'full'
```
sindresorhus: I think we should default to `'sidebar'` to be as native-looking as possible by default.

You also need to add a migration from the boolean type, in this file. This is how I have handled migrations before: https://github.com/atomiclabs/hyperdex/blob/575056075278db372a724cde549743e0ba000bb2/app/config.js#L44-L59

sindresorhus: > I think we should default to 'sidebar' to be as native-looking as possible by default.

Actually, let's wait to turn this on by default until the underlaying issue is fixed in Electron. Add a TODO comment about it however which links to the Electron issue, so we don't forget to change the default in the future.

CvX:
As requested, I've added the TODO comment regarding the default vibrancy setting and implemented the setting migration (`true` -> `'full'`, `false` -> `'sidebar'`).

PR #731, opened 2019-01-18
Add option to use native emoji

### LINKED ISSUE #672: Option to use native Apple emoji
The Facebook emoji set looks terrible. The messenger app on iOS uses native Apple emojis. It would be great if Caprine could do this too

felixfbecker:
Closes #672 

image

image

Adds a new option "native" to the emoji menu.

Uses a MutationObserver to swap emoji `img` elements with the actual emoji text, and injects CSS to show the native emoji in the editor. 

Also added a bunch more docblocks with type info.

I did not enable it by default because unfortunately _some_ emoji are rendered as multiple characters, which is an Electron bug I wasn't able to work around: https://github.com/electron/electron/issues/11874

Despite this flaw (which I hope will be fixed in the future) I would still use this option and think it is good to have.


browser.css
```diff
@@ -25,6 +25,14 @@ body {
 	font-feature-settings: 'liga', 'clig', 'kern';
 }
 
+/* Size and align native emojis */
+.native-emoji {
+	font-size: 1.2em;
```
CvX: Emojis in emoji-only messages are supposed to be twice the size.

browser.js
```diff
@@ -465,6 +465,65 @@ document.addEventListener('DOMContentLoaded', () => {
...
+				text-indent: initial;
+				overflow: initial;
+				width: initial;
+			}
+		`;
+		document.body.appendChild(editorEmojiStyle);
+
+		/**
+		 * Replaces all `img` emojis found in the given node with native emojis.
+		 * @param {Node} node
```
CvX: I don't know if we want JSDoc in this project - that's up to @sindresorhus. But if we do, a lot of those in this PR added a bunch of linter warnings (see: `npm run lint`).

browser.js
```diff
@@ -465,6 +465,65 @@ document.addEventListener('DOMContentLoaded', () => {
...
+				if (!img.alt) {
+					continue;
+				}
+				const span = document.createElement('span');
+				span.className = 'native-emoji';
+				// Emoji in alt text is missing Variation Selector-16:
+				// "An invisible codepoint which specifies that the preceding character
+				// should be displayed with emoji presentation.
+				// Only required if the preceding character defaults to text presentation."
+				span.textContent = img.alt + '\uFE0F';
```
CvX: Some of "emoji" actually isn't emoji. 😃

When you type ":)", Messenger replaces that with "🙂", but only visually. The underlying message still has ":)" in it. And that's what ends up in the `alt` attribute.

So either a similar replacement logic should be done on `alt` attributes or the `src` attributes should be used instead. At least when no emoji could be found in `alt`.

emoji.js
```diff
@@ -273,7 +307,8 @@ module.exports = {
 		return [
 			emojiMenuOption('Facebook 3.0', 'facebook-3-0'),
 			emojiMenuOption('Messenger 1.0', 'messenger-1-0'),
-			emojiMenuOption('Facebook 2.2', 'facebook-2-2')
+			emojiMenuOption('Facebook 2.2', 'facebook-2-2'),
+			emojiMenuOption('🙂Native', 'native')
```
CvX: This sample emoji has to be saved like the others are. Otherwise it's misaligned and its color is off.

CvX: Oh, and I'd move this option between "Facebook 3.0" and "Messenger 1.0" - I think it makes sense to have most complete emoji sets at the top.

Or is that assumption wrong? How's the emoji support on e.g. Linux? Is it viable or should this option be disabled on some platforms?

CvX: Actually, if we end up supporting native emoji on multiple platforms, then sample emoji shouldn't be saved in a file (duh!), but rather dynamically rendered onto a `canvas`. How exciting! 😄

felixfbecker: How can I create an electron NativeImage from an emoji?

browser.js
felixfbecker: I find it hard to find a reason why one would NOT want documentation. It took me a non-zero amount of time to understand this code as I had to search for callsites of functions just to understand what paramters they expect and what the return values mean. These comments would have saved me all this time and will save it for the next developer.

The warnings are about missing descriptions. I believe that a docblock with types is better than no docblock at all. Apparently xo thinks otherwise, but it’s only a warning.

browser.css
felixfbecker: This is using em so it should make them (2 * 1.2) large then (assuming the font size is 2x)

emoji.js
janosorcsik: Load the image and try to draw onto the canvas like this:
https://github.com/sindresorhus/caprine/blob/00746cfce60b4332eb03945ad48e0888a90fa9e0/browser.js#L530-L541
After need to convert the canvas to a DataURL and create a NativeImage from this.
```js
const dataURL   = canvas.toDataURL();
const nativeImg = nativeImage.createFromDataURL(dataURL);
```

CvX: That's correct, just instead of loading an image you create canvas directly, and then call `fillText()` with given emoji on its context.

browser.css
CvX: The font size is the same, Messenger just uses 32px emoji imgs there, so it has to be handled in some way.

emoji.js
felixfbecker: Hmm, the menu code is called from index.js though which doesn't have access to the DOM and therefor can't call `document.createElement('canvas')`. Other ideas?

CvX: Showing that preview emoji is gonna be more complex than everything else in this feature. 😅

It has to be done with IPC:
* Instead of creating menu synchronously we `sendBackgroundAction('render-emoji')`
* in browser.js:
  ```js
  ipc.on('render-emoji', () => {
    const emoji = renderEmojiWithCanvas();
    ipc.send('update-preview-emoji', emoji.toDataURL());
  });
  ```
* in index.js:
  ```js
  ipcMain.on('update-preview-emoji', (event, renderedEmoji) => {
    updateMenu({emoji: nativeImage.createFromDataURL(renderedEmoji)})
  });
  ```

felixfbecker: I can't figure it out and it is becoming kind of a time sink. Can I just leave the icon out for "Native"?

CvX: Yeah, that can be added another time.

felixfbecker:
Resolved all outstanding issues

browser.js
felixfbecker: JSDoc and TS don't contradict each other and JSDoc annotations don't have to be removed if TS is adopted. I would love to see TS adopted here but that is a much larger task. Adding a few docblocks should be uncontroversial and I don't see any disadvantages :)

felixfbecker:
Build for testing: https://we.tl/t-9V4ZMO0boX

CvX:
Some things are still broken:

- [ ] Native emoji is displayed with various alpha color values (0.7 in the sidebar, 0.9 in the conversation when using the dark mode, etc.). `.native-emoji` should have its `color` set to something like `rgba(0, 0, 0, 255)`.
- [ ] Emojis in the reaction popover are misaligned:
  screen shot 2019-01-20 at 15 13 31
- [ ] Emojis in the "insert emoji" popover are misaligned too:
  screen shot 2019-01-20 at 15 13 44
- [ ] Emoticons (e.g. ":)") in the message composer are no longer being replaced with emoji while typing the message

felixfbecker:
>  Native emoji is displayed with various alpha color values (0.7 in the sidebar, 0.9 in the conversation when using the dark mode, etc.). .native-emoji should have its color set to something like rgba(0, 0, 0, 255).

Could you elaborate on this? Shouldn't the emoji be displayed like the surrounding text?

CvX:
Ouch, here's a biggie:

- [ ] Send ":)", restart Caprine, send another ":P". The last message in the conversation list doesn't update. Try sending "test", you end up with this: 
screen shot 2019-01-20 at 15 26 15

felixfbecker:
The error that causes the "Could not display composer" is
```
DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at Xg (https://static.xx.fbcdn.net/rsrc.php/v3iLl54/yp/l/en_US/EYCXyA9EU03.js:47:81303)
    at fi (https://static.xx.fbcdn.net/rsrc.php/v3iLl54/yp/l/en_US/EYCXyA9EU03.js:47:101019)
    at ei (https://static.xx.fbcdn.net/rsrc.php/v3iLl54/yp/l/en_US/EYCXyA9EU03.js:47:98043)
    at ci (https://static.xx.fbcdn.net/rsrc.php/v3iLl54/yp/l/en_US/EYCXyA9EU03.js:47:97387)
    at ji (https://static.xx.fbcdn.net/rsrc.php/v3iLl54/yp/l/en_US/EYCXyA9EU03.js:47:104465)
    at hd (https://static.xx.fbcdn.net/rsrc.php/v3iLl54/yp/l/en_US/EYCXyA9EU03.js:47:32140)
    at HTMLDocument.n (https://static.xx.fbcdn.net/rsrc.php/v3/yF/r/KAUB_fijJwW.js:103:2698)
    at Object.applyWithGuard (https://static.xx.fbcdn.net/rsrc.php/v3/yF/r/KAUB_fijJwW.js:55:2237)
    at t.fire (https://static.xx.fbcdn.net/rsrc.php/v3/yF/r/KAUB_fijJwW.js:263:7487)
    at HTMLDocument.<anonymous> (https://static.xx.fbcdn.net/rsrc.php/v3/yF/r/KAUB_fijJwW.js:263:6644)
```

It seems like messenger doesn't like us messing with their DOM nodes?

CvX:
Yeah… 😕

…I have an idea… that involves removing the `MutationObserver` and rendering emojis to canvas. 🙈

You know how currently the emoji replacement introduced in #720 uses redirects to change which emoji image is getting displayed? Well, it's possible to redirect to a dataURL! 😃

So we can render emoji (taken from the original URL and processes with `String.fromCodePoint()`) on-demand to a canvas, convert them to dataURL format and redirect to it. 😃

felixfbecker:
I'm kind of out of ideas on this one. I think what we're seeing here is React holding a references to the old `img` that we replaced with a `span`. When it discovers that it needs to remove that `img` as part of a rerender, it calls `removeChild()`, which fails because the `img` is not anymore in the DOM and it really should be removing the new `span` instead.

The emojis in the message editor are possible to replace without this error because they use `span`s with `background-image`, so we can modify the element's `textContent` in place and just remove the `background-image` style. But emojis in the message list are `img`s, which cannot have contents, not even with `display: contents` or pseudo elements.

felixfbecker:
> Yeah… 😕
> 
> …I have an idea… that involves removing the `MutationObserver` and rendering emojis to canvas. 🙈
> 
> You know how currently the emoji replacement introduced in #720 uses redirects to change which emoji image is getting displayed? Well, it's possible to redirect to a dataURL! 😃
> 
> So we can render emoji (taken from the original URL and processes with `String.fromCodePoint()`) on-demand to a canvas, convert them to dataURL format and redirect to it. 😃

Nice, that should work. But it'll have to be done through IPC I think since the interceptor is defined in the main process.

CvX:
Yup, the strategy would be pretty much like in https://github.com/sindresorhus/caprine/pull/731#discussion_r249248727. 🙂

felixfbecker:
It's a bit annoying that there is no real IPC layer used yet, like JSON RPC or [comlink](https://github.com/GoogleChromeLabs/comlink).

felixfbecker:
Need to tweak sizing on Windows

image

felixfbecker:
Should be all fixed! 🎉

CvX:
Almost there! 🎉😃

Just four more issues:

- [ ] There are a couple of broken emojis:
  screen shot 2019-01-21 at 03 23 27 screen shot 2019-01-21 at 03 23 57
- [ ] When you have a custom emoji set for a conversation (see the "ⓘ" sidebar) you can send an enlarged emoji (max 128px it seems) by holding it down. Messenger actually downloads the 128px versions. It would be nice if native ones would also be rendered at a higher quality in that situation.
  screen shot 2019-01-21 at 03 34 51
- [ ] If there are more than 10 emojis loaded at a time we get a warning `axListenersExceededWarning: Possible EventEmitter memory leak detected. 11 native-emoji listeners added. Use emitter.setMaxListeners() to increase limit`. We don't actually leak any listeners, and all the warnings I've seen said there's exactly 11 of them, so we should probably raise that limit. (to 11?)
- [ ] In the "Pick the emoji for this conversation" modal, facebook's thumbs-up is broken. We should either pass 'f0000' through or replace it with emoji thumbs-up.
  screen shot 2019-01-21 at 03 43 43

Also, I'm adding a couple of commits:

* Changed ESLint rules regarding JSDoc and fixed a warning so that this PR doesn't introduce a bunch of new warnings
  screen shot 2019-01-21 at 03 48 41
* Moved and renamed the "Native" option. It's after "Facebook 3.0" now (since that set is the default). And the new label is platform-dependent ("macOS", "Windows", or "Linux")
* Added the native emoji preview to the menu
  screen shot 2019-01-21 at 03 50 18

emoji.js
```diff
@@ -209,32 +219,117 @@ function codeForEmojiStyle(style) {
 	}
 }
 
-const menuIcons = new Map();
+function nativeEmojiLabel() {
+	if (is.macos) {
+		return 'macOS';
```
felixfbecker:
```suggestion
		return 'Apple';
```

This is the "Apple" emoji set (also used on iOS), see the bottom of https://emojipedia.org/

emoji.js
```diff
@@ -209,32 +219,117 @@ function codeForEmojiStyle(style) {
...
+	}
+
+	if (is.windows) {
+		return 'Windows';
```
felixfbecker:
```suggestion
		return 'Microsoft';
```

This is the "Microsoft" emoji set (also used on Windows Phone, Xbox etc), see the bottom of https://emojipedia.org/

emoji.js
```diff
@@ -209,32 +219,117 @@ function codeForEmojiStyle(style) {
...
+	}
 
-function getEmojiIcon(style) {
-	if (menuIcons.has(style)) {
-		return menuIcons.get(style);
+	if (is.linux) {
+		return 'Linux';
```
felixfbecker: There isn't really a "Linux" emoji set, not sure what to show here. Detecting the actual set in use is probably not worth it. Honestly I think I would prefer to show "Native" for all three options because showing a different label makes it seem like we support that emoji set specifically and e.g. I would be able to select the "Apple" emoji set on Windows too, which is not what it does - it's just the native emoji set used by the OS.

felixfbecker:
>  There are a couple of broken emojis:

This is https://github.com/electron/electron/issues/11874 as mentioned in the PR description. As said I don't consider this a blocker because I would hope it gets fixed eventually and users need to opt-in to use native emojis (and this is simply what the native rendering returned).

emoji.js
```diff
@@ -357,8 +357,8 @@ module.exports = {
 		});
 
 		return [
-			emojiMenuOption('Native', 'native'),
 			emojiMenuOption('Facebook 3.0', 'facebook-3-0'),
+			emojiMenuOption('Native', 'native'),
```
felixfbecker: I would prefer it to be at the bottom then, because it should be separate from the Facebook emoji sets.

emoji.js
CvX: I see your point.
But how about changing the label to e.g. "System"? "Native" is very tech-speak.

emoji.js
CvX: Ok, you prefer to have Facebooks sets grouped together, I find important having more complete sets at the top and highlighting the default option.

How about this? 🙂 
screen shot 2019-01-21 at 16 11 41

CvX:
> * There are a couple of broken emojis:
>   screen shot 2019-01-21 at 03 23 27 screen shot 2019-01-21 at 03 23 57

An update on "broken emojis" issue:

As per https://github.com/electron/electron/issues/16530 Electron 5.0.0-beta.1 contains a newer Chromium version that fixes most (though not all) emoji rendering problems:

emojis 1 emojis 2

It seems that there's a change in emoji's vertical offset when rendering, so on 5.0.0-beta.1 they're cut off a bit at the top. It's something to look out for when we upgrade after 5.0 stabilizes.

Anyway, I'm marking this one as "fixed".
The other 3 issues remain (128px emojis, "MaxListenersExceededWarning", and facebook's thumbs-up).

felixfbecker:
Merged in master with change to TypeScript

CvX:
Thank you! 🎉🚀

CvX:
Merged in the master once again, and fixed the Facebook's thumbs-up:

felixfbecker:
Btw I just discovered [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) which is also available in WebWorkers. I wonder if it's available in the Electron main process?

sindresorhus:
https://electronjs.org/docs/tutorial/offscreen-rendering

CvX:
Hm, I don't think Electron's offscreen rendering is a good fit here. Keeping a spare (hidden) frameless window just for rendering emoji seems excessive. 😃

felixfbecker:
All done!

sindresorhus:
Nice work, everyone! 🎉

felixfbecker:
:shipit: 🚢

Issue #591, opened 2018-09-20
Create shortcut for "Archive and go to next message"


jonleung:
Instead of just archiving and going to a limbo state that requires me to advance to the next conversation with another shortcut key, have a new shortcut that allows me to archive a message AND go to the next message. I would KILL to have this feature!

Thank you!

sindresorhus:
I think it should just do this by default. No option needed.

IssuehuntBot:
@dusansimic has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/42574339/issues/591)

IssuehuntBot:
@sindresorhus has rewarded $18.00 to @dusansimic. [See it on IssueHunt](https://issuehunt.io/repos/42574339/issues/591)

- :moneybag: Total deposit: $20.00
- :tada: Repository reward(0%): $0.00
- :wrench: Service fee(10%): $2.00

Issue #590, opened 2018-09-20
Archiving group messages tries to leave the group convo instead of archiving it


jonleung:
OSX 10.12.6

Instructions to reproduce:

1. Click any group conversation in the sidebar
2. `Cmd + Shift + a`
3. I get this message:
> 
![image](https://user-images.githubusercontent.com/654941/45850584-ac1e8400-bd04-11e8-8dfb-fb6baf6e91c3.png)

Archiving messages is the primary reason I have downloaded this app. SO EXCITED TO HAVE THIS WORKING! Thank you for creating this!

IssuehuntBot:
@issuehuntfest has funded $30.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/42574339/issues/590)

IssuehuntBot:
@dusansimic has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/42574339/issues/590)

IssuehuntBot:
@sindresorhus has rewarded $27.00 to @dusansimic. [See it on IssueHunt](https://issuehunt.io/repos/42574339/issues/590)

- :moneybag: Total deposit: $30.00
- :tada: Repository reward(0%): $0.00
- :wrench: Service fee(10%): $3.00

PR #732, opened 2019-01-18
Fix archive conversation issues

dusansimic:
Fixed archive conversation issues. Now when a conversation is archived, it automatically switches to the next conversation so there wont be an empty display. Also with this PR fixed the issue with archiving group conversations.

Closes: https://github.com/sindresorhus/caprine/issues/590
Closes: https://github.com/sindresorhus/caprine/issues/591

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -11,14 +11,14 @@ const conversationSelector = '._4u-c._1wfr > ._5f0v.uiScrollableArea';
 const selectedConversationSelector = '._5l-3._1ht1._1ht2';
 const preferencesSelector = '._10._4ebx.uiLayer._4-hy';
 
-async function withSettingsMenu(callback) {
+async function withMenu(menuButtonElement, callback) {
 	const {classList} = document.documentElement;
 
 	// Prevent the dropdown menu from displaying
 	classList.add('hide-dropdowns');
 
-	// Click the Settings icon
-	(await elementReady('._30yy._2fug._p')).click();
+	// Click the menu button
+	menuButtonElement.click();
 
 	// Wait for the menu to close before removing the 'hide-dropdowns' class
 	const menuLayer = document.querySelector('.uiContextualLayerPositioner:not(.hidden_elem)');
@@ -31,6 +31,10 @@ async function withSettingsMenu(callback) {
 	observer.observe(menuLayer, {attributes: true, attributeFilter: ['class']});
 
 	await callback();
+}
+
+async function withSettingsMenu(callback) {
+	await withMenu(await elementReady('._30yy._2fug._p'), callback);
 }
 
 function selectMenuItem(itemNumber) {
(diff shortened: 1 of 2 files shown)


browser.js
```diff
@@ -256,8 +258,8 @@ function jumpToConversation(key) {
 }
 
 // Focus on the conversation with the given index
-function selectConversation(index) {
-	document.querySelector(listSelector).children[index].firstChild.firstChild.click();
+async function selectConversation(index) {
```
CvX: If this functions is now async, then each function that calls it should be converted to async as well (and should await for this one). Same with functions that call those functions, and so on.

browser.js
```diff
@@ -319,7 +321,11 @@ function openArchiveModal() {
 		return;
 	}
 
-	selectMenuItem(3);
+	// Check if selected conversation is a group
+	// In case it is we need to click on 4th element of conversation menu, otherwise 3rd
+	const isGroup = document.querySelector('._54nq._2i-c._150g._558b._2n_z li:nth-child(3) a span span').innerHTML === 'Leave Group';
```
CvX: You can't rely on innerHTML/text since Messenger's interface is internationalized, so this label can be different for different users.

browser.js
```diff
@@ -332,7 +338,7 @@ function openDeleteModal() {
 
 async function openPreferences() {
 	// Create the menu for the below
-	(await elementReady('._30yy._2fug._p')).click();
+	showSettingsMenu();
```
CvX: Not awaiting with `elementReady` renders the preferences menu invisible (`opacity: 0`).

dusansimic:
Here it is 😃

browser.js
CvX: Take a look at this branch: https://github.com/sindresorhus/caprine/compare/master...CvX:archive-fix

It's a solution I started working on a couple of days ago. It allows to operate on the active dropdown menu, without having to guess which menu is open.

The similar thing should be done for `openConversationMenu` and that would be it, right? Feel free to try it out, and merge it to your branch as needed.

dusansimic: Just merged your code and used that mechanism for `openConversationMenu` (now `withConversationMenu`) and it's fantastic.
Great work 🙌🏻.

CvX: This comment got lost in my inbox but I thought the same thing about merging those two functions when I saw the code. 🙂

CvX:
I've cleaned up the code a little bit. Please take a look, I want to be sure everything is as expected. 😉

dusansimic:
Looks good 👌🏻. I've just tested it and found out that actually delete chat option behaves just like archive option used to, gets removed from the list and leaves the chat panel empty. Should we add the same feature to delete chat option while were here?

CvX:
Oh, yeah. 😅That would be great! 😃

Issue #733, opened 2019-01-19
Dark mode greyed out


slayerbrk:
Dark mode appears greyed out and can not be enabled.

CvX:
You probably have the "View > Follow System Appearance" enabled. If you want to enable the "Dark Mode" independently of system theme setting, disable that first.

@sindresorhus, maybe it would be better to have the "Follow System Appearance" always enabled, and just toggle it off if you change the "Dark Mode" setting? Dunno which way is less confusing. 🤷‍♂️

19majkel94:
I have dark mode enabled in Windows 10 settings but the dark mode is still grayed out.

![image](https://user-images.githubusercontent.com/10618781/51424839-c6457180-1bd3-11e9-9a98-1955c3322a34.png)

I also don't have the "Follow system apperance" options 😕

janosorcsik:
The "Follow System Appearance" shows only on MacOS.
The "Dark mode" enabled if the "followSystemAppearance" config is false, but it is always true in config.
The "followSystemAppearance" needs to true only on MacOS.

dhuertasanchez:
The same thing happens to me, the dark mode is not activated, it happens to me in Linux, Mac and Windows, the strange thing is that a couple of days ago it worked correctly, but then the color changed to a gray one and with the texts in white color so that can not be distinguished, and in the conversations the conversation panel is shown in dark mode, but the contacts panel keeps the gray color with white texts...
![darkmode_broken](https://user-images.githubusercontent.com/8291300/51932872-af6a0080-23c5-11e9-844e-852ee8489f32.png)

CvX:
@dhuertasanchez This issue is tracked in #712.
The current workaround is to disable vibrancy "View -> Vibrancy -> No Vibrancy"

dhuertasanchez:
@CvX Hi, thank you, yep, i have already deactivated vibrancy, although in Linux and Windows the error persists with the dark mode, something strange happens there, I will have to continue with the light mode

CvX:
I'll look into dark mode on Windows tomorrow. Thanks for the info!

PR #735, opened 2019-01-19
Fix dark mode being disabled on non-macOS systems

dusansimic:
Dark mode option is disabled on non macOS systems since `followSystemAppearance` config value is always true by default when it should be true by default only on macOS.

Closes: https://github.com/sindresorhus/caprine/issues/733

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,10 +1,9 @@
 'use strict';
 const Store = require('electron-store');
-const {is} = require('electron-util');
 
 const defaults = {
 	defaults: {
-		followSystemAppearance: is.macos,
+		followSystemAppearance: true,
 		darkMode: false,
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Should I squash'n'merge ready PRs or just approve them and leave that to you, since you're preparing releases anyway?)

Feel free to merge bug fixes if you have reviewed and tested them thoroughly. New features should get multiple opinions before merging though.

sindresorhus:
We need to have the macOS check here too: https://github.com/sindresorhus/caprine/blob/00746cfce60b4332eb03945ad48e0888a90fa9e0/menu.js#L286 As current users now have `followSystemAppearance` as true no matter what system they're on.

sindresorhus:
Actually, we don't need the check in `config.js`. The checks should be where it's used, not in the config. We have not guarded other system-specific config like `vibrancy` either.

PR #738, opened 2019-01-20
Allow disabling video autoplay

### LINKED ISSUE #393: Stop autoplay for the videos and animated gif files
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/caprine/issues/393)

Hi all.

This is not a bug report but a feature request.
It will be very, very nice to have some option to stop autoplay for the videos and animated gif files
which are sent in messages.

And it would be best to have some kind of blocking of receiving msg'a with such attachments.
For example when the blocade is turned on I can recive only text from such msg's but not the attached video or gif.

Operating system: Linux Mint 18.3 Mate x64
Caprine version: v2.9.0

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/caprine/issues/393).

timomyllymaki:
Hi,

Fixes #393.

Adds an option to disable video autoplay.

Kind of implemented as suggested here https://github.com/sindresorhus/caprine/issues/393#issuecomment-452622199. It doesn't actually capture the first frame of the video but instead takes one from the video parent element background-image which holds a snapshot of the video. Its not the first frame but I didn't think it was that important as this was really easy to implement.


timomyllymaki:
Yes sure. I'll try to fix them asap.

timomyllymaki:
I fixed alI the requested changes. Also changed the mutation observers a little. Seems like it's a lot better to not iterate over mutations but to search for the videos by tag when we have mutations. By my tests this way seems to perform better.

spramij:
@timomyllymaki This needs a rebase from `origin:master` to accommodate for changes in code that were done since.

sindresorhus:
When this feature is enabled, I noticed that the videos lose their round corners. Can you handle that?

source/config.ts
```diff
@@ -36,7 +36,8 @@ const defaults = {
 	notificationsMuted: false,
 	hardwareAcceleration: true,
 	quitOnWindowClose: false,
-	keepMeSignedIn: true
+	keepMeSignedIn: true,
+	videoAutoplay: true
```
sindresorhus: `autoplayVideos`

source/menu.ts
```diff
@@ -173,6 +173,16 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				config.set('bounceDockOnMessage', !config.get('bounceDockOnMessage'));
 			}
 		},
+		{
+			label: 'Video Autoplay',
```
sindresorhus: `Autoplay Videos`

sindresorhus:
Can you move as much of the feature as possible into a separate file? For maintenance reasons.

timomyllymaki:
I refactored the feature to own .ts and .css -files. Also fixed the rounded corners (thus new .css -file) and changed videoAutoplay -> autoplayVideos. Let me know if there's something else to fix!

CvX:
Almost there! 🙂 The preview has only top corners rounded, as opposed to all four corners rounded with autoplay enabled:

autoplay disabled autoplay enabled

timomyllymaki:
Oh, I had only tested it with one of these videos: 

image

Made a fix which hopefully solves the problem!

CvX:
> Oh, I had only tested it with one of these videos:

Ha, I didn't even think about videos with a description below! 😃
It's great now! 🙂 

autoplay disabled, after changes

sindresorhus:
Thank you!

sindresorhus:
@timomyllymaki You need to submit the PR URL to https://issuehunt.io/r/sindresorhus/caprine/issues/393 to claim the bounty.

PR #739, opened 2019-01-20
Allow to change chat colors with HEX value

mikigal:
Allow to change chat colors with HEX value. Feature will be visible only for you (or maybe for other Caprine users in future).

TODO:
 - Sync colors with other Caprine users, like in original messenger (send "change color command" with HEX to second user)
 - Create color palette instead of manual write HEX value (currently I haven't idea how to code it)
 - Style prompt to dark theme
 - Find other color selectors (most are already found!)
 - Add warning, that you must switch to another chat and back to view new color
 - Fix 'like emoji' as chat button (sometimes does not change color, or left-margin disappear)

g1x0m9cpcc

### DIFF
--- a/colors.js
+++ b/colors.js
@@ -1,0 +1,87 @@
+'use strict';
+const {dialog} = require('electron');
+const config = require('./config');
+const {getWindow} = require('./util');
+
+module.exports = {
+	changeColor: (color) => {
+		if (!checkRgb(color)) {
+			return dialog.showMessageBox({
+				detail: 'Invalid HEX code!',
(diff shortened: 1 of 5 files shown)


sindresorhus:
We appreciate the PR, but this is not something we want. It's too much code to maintain for an unimportant feature.

Instead, I would recommend you send a feature request to Facebook Messenger. Go to https://www.messenger.com, login, click the cog icon, click "Report a Problem". If messenger.com implements it, Caprine also gets it.

PR #741, opened 2019-01-21
Add git pre-commit script

janosorcsik:
Sometimes contributors forgive to run the npm lint.
This PR add automatic checking before commit.


felixfbecker:
Yeah -1 on lint-staged. It doesn't work when staging files partially and you are left banging your head on the table why git silently commits the whole file instead of just what you staged.

PR #745, opened 2019-01-24
Move to TypeScript

### LINKED ISSUE #740: Move to TypeScript?
@CvX Sounded like you want this and I don't mind. I find TypeScript much easier to work with in larger codebases.

CvX:
[WIP]
This PR converts the codebase to [TypeScript](https://www.typescriptlang.org).

Remaining tasks:

- [x] Resolve all “TODO: [TS]” comments
- [ ] Add TS type definitions to electron-util, electron-debug, electron-dl, electron-context-menu, facebook-locales
- [x] Fix "build/files" in package.json, try out the build
- [x] Look into TS linting via ESLint/xo
- [x] Enable stricter TS rules in tsconfig.json, fix all errors

---

Fixes #740


sindresorhus:
> Look into TS linting via ESLint/xo

Use this: https://github.com/xojs/eslint-config-xo-typescript#use-with-xo

> Enable stricter TS rules in tsconfig.json, fix all errors

Use this: https://github.com/sindresorhus/tsconfig

---

Can you use name the `src` directory `source` instead?

src/emoji.ts
```diff
@@ -1,8 +1,7 @@
-'use strict';
-const path = require('path');
-const {nativeImage} = require('electron');
-const config = require('./config');
-const {showRestartDialog} = require('./util');
+import {join} from 'path';
```
sindresorhus: Don't import this as named import. `join` is too ambiguous without the `path.` prefix.

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+	"compilerOptions": {
+		"target": "ES2017",
```
sindresorhus: We can use `esnext` (lowercase) here. Electron supports it.

CvX: I don't think we can't change it after all. While Electron does support it, it requires all our dependencies to use ES modules.

sindresorhus:
> Any reason why source specifically? Isn't src a convention for TS projects?

That's what I settled on and use in other projects too. `src` goes back to the old Unix times where file paths were limited. `source` looks better and is more readable. No good reason to shorten it.

src/browser.ts
```diff
@@ -1,14 +1,15 @@
 import {ipcRenderer as ipc} from 'electron';
 import {api, is} from 'electron-util';
-import elementReady = require('element-ready');
 import config from './config';
 
+import elementReady = require('element-ready');
```
sindresorhus: I would prefer just normal require if `import` cannot be used.

```js
const elementReady = require('element-ready');
```

src/browser.ts
```diff
@@ -612,6 +617,7 @@ function showNotification({id, title, body, icon, silent}): void {
 	});
 }
 
-ipc.on('notification-callback', (_event, data) => {
+// TODO: [TS] event type
+ipc.on('notification-callback', (_event: Electron.IpcMessageEvent, data: any) => {
```
sindresorhus: Prefer `unknown` over `any` whenever possible

tsconfig.json
sindresorhus: No, that's not what it does. It just sets the target JS level. Nothing to do with modules. I think you're confusing it with the `module` config key.

sindresorhus:
> Update the @types/electron-util package, add it to the project, change the imports

That type package is very outdated. I think it would be easier to just start from scratch and submit it directly to the `electron-util` package instead. I accept bundled typings for my packages. See: https://github.com/sindresorhus/typescript-definition-style-guide

tsconfig.json
CvX: Ah, we were both right. 😉 `target` dictates the `module` (if not specified), and while it is specified in https://github.com/sindresorhus/tsconfig/blob/master/tsconfig.json, for whatever reason Atom/ide-typescript doesn't pick that up (so it has to be defined again in the project).

CvX:
> > Any reason why source specifically? Isn't src a convention for TS projects?
> 
> That's what I settled on and use in other projects too. `src` goes back to the old Unix times where file paths were limited. `source` looks better and is more readable. No good reason to shorten it.

Should we change `dist` too? 😉

Jokes aside, what about `dist-ts`? I've named it that for a lack of a better idea at the time.
Should we move that inside `dist`, e.g. `dist/js` or is it better to keep those two separate? If so, at least I'd replace "ts" in that name to "js" (since that's what's inside).

src/browser.ts
CvX: Hm, it seems that `import` is necessary to retrieve type information from given package.
In our case that means errors in config.ts (the `Store` type).

sindresorhus: But why are you not just using normal `import` syntax like the other imports?

sindresorhus: > Standard import
> Module '"(…)/node_modules/@types/element-ready/index"' has no default export.
Ideally we would use this format, but that requires dependencies to

That's what https://github.com/sindresorhus/tsconfig/blob/4a00939dd2f7a4478eec278e096b37668fd0bce3/tsconfig.json#L7-L8 is for. I tried using `import elementReady from 'element-ready';` locally and building (`$ tsc`) works for me.

sindresorhus:
Yeah, I usually would go for `distribution`, but since we already have `dist`, which we cannot change as it's controlled by `electron-builder`, I think we should go with `dist-js`.

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+	"extends": "@sindresorhus/tsconfig",
+	"compilerOptions": {
+		"target": "esnext",
+		"module": "commonjs",
+		"lib": [
+			"ESNext",
+			"DOM",
+			"DOM.Iterable"
+		],
+		"downlevelIteration": true,
```
sindresorhus: Why this?

CvX: No longer needed, removing (was required at some point to make `[...element.children]` working).

src/browser.ts
CvX: Ah, got it: it's a tooling error. Don't know where does Atom's `ide-typescript` take these errors from…
It seems to be all good with `atom-typescript` and in VSCode (which I usually work in, but I just moved back to Atom for a while to work on this PR after issues with xo linting).

I'm gonna take another look and change those imports.

source/emoji.ts
```diff
@@ -197,23 +196,37 @@ const excludedEmoji = new Set([
...
 		default:
-			return 't';
+			return EmojiStyleCode.Facebook30;
 	}
 }
 
-const menuIcons = new Map();
+const menuIcons = new Map<EmojiStyle, Electron.NativeImage>();
+
+function getEmojiIcon(style: EmojiStyle): Electron.NativeImage {
```
sindresorhus: Could we import the Electron types instead of using the global implicit type?

```typescript
import {NativeImage} from 'electron';

function getEmojiIcon(style: EmojiStyle): NativeImage {
```

sindresorhus:
@CvX Is there anything left to do? If not, I'll do a final review and merge.

CvX:
1. Have to merge the master to this branch to fix the conflicts. I'll do it shortly.
2. Ideally we would get `facebook-locales` definition out of the repo. I've submitted the PR (https://github.com/wix/facebook-locales/pull/6) but maybe it will go to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) instead. In any case, we could update that later, after this gets merged into the master.
3. I'm not sure about other .d.ts files (those that are Caprine-specific). They are just thrown out there, I feel like they could probably should be colocated with they code that uses these definitions, but that would have to wait for a bit of a refactoring (where we would group code of a single feature together, like we do with emoji now). Again, not a blocker IMO.

sindresorhus:
Wooooot!! Super nice work on this. It makes the codebase so much easier to work with. ❤️

PR #760, opened 2019-02-08
Respect "Do Not Disturb" on macOS

### LINKED ISSUE #496: Respect Do Not Disturb on macOS
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/caprine/issues/496)

With the Do Not Disturb mode turned on in Preferences, I'm still getting sound and visual (dock icon bouncing) notifications from Caprine. Would this be possible to fix, so as to Do Not Disturb works as expected with Caprine?

> There is a $30.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/caprine/issues/496).

whitecrownclown:
Fixes #496 

* Using https://github.com/felixrieseberg/electron-notification-state to get the DND status (includes both mac DND and windows quiet hours)
* Disables bounce and messenger sound notification
* Only works on init/refresh, if status changes while the app is on, it will not update. Any clue if and how can this be done?


sindresorhus:
> * Using [felixrieseberg/electron-notification-state](https://github.com/felixrieseberg/electron-notification-state) to get the DND status

We can't use this module as it uses native Node.js add-ons, which mean we'll no longer be able to build all the binaries on one system.

> * Disables bounce and messenger sound notification

Should it disable the Dock badge too?

> * Only works on init/refresh, if status changes while the app is on, it will not update. Any clue if and how can this be done?

You can listen to the `com.apple.notificationcenterui.dndprefs_changed` distributed Apple event with https://electronjs.org/docs/api/system-preferences#systempreferencessubscribenotificationevent-callback-macos

sindresorhus:
For macOS, you might not even need a module for this. I think you can just read the `doNotDisturb` key from the UserDefaults `com.apple.notificationcenterui`. (See Electron docs on how to read user defaults).

whitecrownclown:
@sindresorhus I've been trying to get the `doNotDisturb` user default using Electron's API but i'm always gettings `false`, which also seems to be the case for any key that doesn't exist. 

`systemPreferences.getUserDefault('doNotDisturb', 'boolean');`

Also my attempt to get the event didn't work, not getting any callback : 

`systemPreferences.subscribeNotification('com.apple.notificationcenterui.dndprefs_changed', callback)`

sindresorhus:
> I've been trying to get the `doNotDisturb` user default using Electron's API but i'm always gettings `false`, which also seems to be the case for any key that doesn't exist.

You're not specifying the suite name, which is not surprising, since it's not currently possible in Electron. I've opened an issue about it: https://github.com/electron/electron/issues/17031

> systemPreferences.subscribeNotification('com.apple.notificationcenterui.dndprefs_changed', callback)

I have no idea why this is not working though...

sindresorhus:
Seems it would be easiest to just go with https://github.com/sindresorhus/do-not-disturb and polling for now.

Ideally, the `do-not-disturb` package should have a `change` event though: https://github.com/sindresorhus/do-not-disturb/issues/6

felixfbecker:
> * Only works on init/refresh, if status changes while the app is on, it will not update. Any clue if and how can this be done?

If that's true I would rather not have this feature...

sindresorhus:
@felixfbecker Polling is an option, or implementing https://github.com/sindresorhus/do-not-disturb/issues/6.

whitecrownclown:
@sindresorhus I can take on the polling approach. I'm not sure how to start on https://github.com/sindresorhus/do-not-disturb/issues/6

whitecrownclown:
@sindresorhus 	I've implemented a polling solution. Would you mind having a look over it for any feedback? Thanks!

package.json
```diff
@@ -25,11 +25,13 @@
 		"publish-snap": "del dist && tsc && electron-builder --linux && snapcraft push --release=stable dist/*.snap"
 	},
 	"dependencies": {
+		"@sindresorhus/do-not-disturb": "^0.2.0",
 		"electron-context-menu": "^0.11.0",
 		"electron-debug": "^2.1.0",
 		"electron-dl": "^1.13.0",
 		"electron-is-dev": "^1.0.0",
 		"electron-log": "^2.0.2",
+		"event-emitter": "^0.3.5",
```
sindresorhus: Why are you not using the built-in `events` module?

source/browser.ts
```diff
@@ -7,6 +7,7 @@ const listSelector = 'div[role="navigation"] > div > ul';
 const conversationSelector = '._4u-c._1wfr > ._5f0v.uiScrollableArea';
 const selectedConversationSelector = '._5l-3._1ht1._1ht2';
 const preferencesSelector = '._10._4ebx.uiLayer._4-hy';
+const messengerSoundsSelector = `${preferencesSelector} ._374d input`;
```
sindresorhus: You might have to update the selector here.

source/do-not-disturb.ts
```diff
@@ -0,0 +1,25 @@
...
+			emitter.emit('dndprefs_changed', newValue);
+		}
+	}, pollingInterval);
+};
+
+if (is.macos) {
+	startPolling();
+}
+
+export {getDoNotDisturb};
```
sindresorhus: The intention was to implement this functionality in `@sindresorhus/do-not-disturb`.

whitecrownclown: I understand, however, i'm not familiar with swift and wouldn't know how to start on contributing to the project. I will have a shot at it this following week though. Thanks!

sindresorhus: I meant, you don't have to touch Swift. I'm ok with a naive polling approach for now, so just move this code to `@sindresorhus/do-not-disturb`.

whitecrownclown:
Waiting on https://github.com/sindresorhus/do-not-disturb/pull/8 to be merged.

sindresorhus:
https://github.com/sindresorhus/do-not-disturb/releases/tag/v1.1.0

sindresorhus:
Can you mention it in the readme highlights that it respects "Do Not Disturb" on macOS?

sindresorhus:
If I have "Sounds" disabled, then turn on "Do Not Disturb" and then turn it off again, "Sounds" is then enabled. It should instead correctly revert back to the state before "Do Not Disturb".

sindresorhus:
Dock badge shows even when "Do Not Disturb" is enabled.

source/browser.ts
```diff
@@ -174,7 +175,31 @@ ipc.on('toggle-mute-notifications', async (_event: ElectronEvent, defaultStatus:
...
+			soundsCheckbox.click();
+		}
+
+		if (shouldClosePreferences) {
+			closePreferences();
+		}
 	}
+);
+
+ipc.on('toggle-mute-notifications', async (_event: ElectronEvent, defaultStatus: boolean) => {
```
sindresorhus: Why this? macOS already prevents notifications when Do Not Disturb is enabled.

sindresorhus:
Thanks :)

PR #774, opened 2019-02-12
Add `Launch at Login` setting for macOS and Windows

### LINKED ISSUE #773: Add "Launch at Login" menu item
https://electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows
https://electronjs.org/docs/api/app#appgetloginitemsettingsoptions-macos-windows

whitecrownclown:
@sindresorhus 
Fixes #773 

Hello!
I've added the following : 

- Launch at login menu item
- Launch as hidden menu item (disabled if launch at login is false)
- Tested on macOS 10.14.3

Feedback is welcome, thanks!

### DIFF
--- a/source/menu.ts
+++ b/source/menu.ts
@@ -192,6 +192,7 @@ Press Command/Ctrl+R in Caprine to see your changes.
 		},
 		{
 			label: 'Launch at Login',
+			visible: is.macos || is.windows,
 			type: 'checkbox',
 			checked: app.getLoginItemSettings().openAtLogin,
 			click(item) {
--- a/source/menu.ts
+++ b/source/menu.ts
@@ -1,6 +1,6 @@
 import * as path from 'path';
 import {existsSync, writeFileSync} from 'fs';
-import {app, shell, Menu, MenuItemConstructorOptions, LoginItemSettings} from 'electron';
+import {app, shell, Menu, MenuItemConstructorOptions} from 'electron';
 import {
 	is,
 	appMenu,
@@ -12,14 +12,6 @@ import {
 import config from './config';
 import {sendAction, showRestartDialog} from './util';
 import {generateSubmenu as generateEmojiSubmenu} from './emoji';
-
-function setLoginSettings(settings: LoginItemSettings): void {
-	app.setLoginItemSettings(settings);
-}
-
-function getLoginSettings(): LoginItemSettings {
-	return app.getLoginItemSettings();
-}
 
 export default function updateMenu(): Menu {
 	const newConversationItem: MenuItemConstructorOptions = {
@@ -136,8 +128,6 @@ Press Command/Ctrl+R in Caprine to see your changes.
 		}
 	];
 
-	const settings: LoginItemSettings = getLoginSettings();
-
 	const preferencesSubmenu: MenuItemConstructorOptions[] = [
 		{
 			label: 'Bounce Dock on Message',
@@ -203,9 +193,9 @@ Press Command/Ctrl+R in Caprine to see your changes.
 		{
 			label: 'Launch at Login',
 			type: 'checkbox',
-			checked: settings.openAtLogin,
+			checked: app.getLoginItemSettings().openAtLogin,
 			click(item) {
-				setLoginSettings({...settings, openAtLogin: item.checked, openAsHidden: item.checked});
+				app.setLoginItemSettings({openAtLogin: item.checked, openAsHidden: item.checked});
 			}
 		},
 		{
--- a/source/menu.ts
+++ b/source/menu.ts
@@ -13,6 +13,20 @@ import config from './config';
 import {sendAction, showRestartDialog} from './util';
 import {generateSubmenu as generateEmojiSubmenu} from './emoji';
 
+function setLoginSettings(): void {
+	const appFolder = path.dirname(process.execPath);
+	const updateExe = path.resolve(appFolder, '..', 'Update.exe');
+
+	const openAtLogin = config.get('openAtLogin');
+	const openAsHidden = config.get('openAsHidden');
+
+	app.setLoginItemSettings({
+		openAtLogin,
+		openAsHidden,
+		path: updateExe
+	});
+}
+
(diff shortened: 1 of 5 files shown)


sindresorhus:
You shouldn’t store the states in the config store. Just fetch the actual states from the `getLogin...` method when needed. That way the settings don’t get out of sync.

source/menu.ts
```diff
@@ -13,6 +13,23 @@ import config from './config';
 import {sendAction, showRestartDialog} from './util';
 import {generateSubmenu as generateEmojiSubmenu} from './emoji';
 
+const appFolder = path.dirname(process.execPath);
+const updateExe = path.resolve(appFolder, '..', 'Update.exe');
```
sindresorhus: ?

source/menu.ts
```diff
@@ -190,6 +209,24 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				focusedWindow.setAlwaysOnTop(item.checked);
 			}
 		},
+		{
+			type: 'checkbox',
+			label: 'Launch at login',
```
sindresorhus: `label` should be first

source/menu.ts
whitecrownclown: I only moved those outside of the function definitions since they're not updating at all

sindresorhus: No, I mean. Can you explain why this is needed?

sindresorhus:
> * Launch as hidden menu item (disabled if launch at login is false)

I don't see the point of having a setting for this. We should just launch hidden.

source/menu.ts
```diff
@@ -13,6 +13,23 @@ import config from './config';
 import {sendAction, showRestartDialog} from './util';
 import {generateSubmenu as generateEmojiSubmenu} from './emoji';
 
+const appFolder = path.dirname(process.execPath);
+const updateExe = path.resolve(appFolder, '..', 'Update.exe');
+
+function setLoginSettings({openAtLogin, openAsHidden}: AppLoginSettings): void {
```
sindresorhus: Just use the type for `setLoginItemSettings` instead of defining your own.

whitecrownclown: Ok, i did look for it at first, didn't find it, i'll have a closer look.

source/menu.ts
whitecrownclown: Initially it was for : [To work with Electron's autoUpdater on Windows](https://electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows).

whitecrownclown: I removed it for now, as i cannot really test it and also don't have access to windows.

whitecrownclown:
@sindresorhus I've updated the PR again.

* Added the ts definition from electron
* Removed `Launch as Hidden`; defaults to the same value as `Launch at Login`

source/menu.ts
```diff
@@ -13,6 +13,14 @@ import config from './config';
 import {sendAction, showRestartDialog} from './util';
 import {generateSubmenu as generateEmojiSubmenu} from './emoji';
 
+function setLoginSettings(settings: LoginItemSettings): void {
+	app.setLoginItemSettings(settings);
+}
+
+function getLoginSettings(): LoginItemSettings {
+	return app.getLoginItemSettings();
+}
```
sindresorhus: These wrappers are now moot.

source/menu.ts
```diff
@@ -190,6 +200,14 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				focusedWindow.setAlwaysOnTop(item.checked);
 			}
 		},
+		{
+			label: 'Launch at Login',
+			type: 'checkbox',
+			checked: settings.openAtLogin,
+			click(item) {
+				setLoginSettings({...settings, openAtLogin: item.checked, openAsHidden: item.checked});
```
sindresorhus: I don't think you need `...settings`. And `openAsHidden: item.checked` needs to be updated.

whitecrownclown: I thought so as well, but i'm getting : `Type '{ openAtLogin: boolean; openAsHidden: boolean; }' is missing the following properties from type 'LoginItemSettings': wasOpenedAtLogin, wasOpenedAsHidden, restoreState`

source/menu.ts
whitecrownclown: Yes, will update

source/menu.ts
whitecrownclown: Disregard, it's no longer applicable since i've removed the wrappers.

whitecrownclown:
@sindresorhus Updated again, however i'm unsure what you mean by this:

>  And openAsHidden: item.checked needs to be updated.

Do you mean to set it to `true` always? It is being updated with the same value as `Launch at Login`. I've noticed if `Launch at Login` is `false` it will not set `Launch as Hidden` as `true`, but `false` as well.

sindresorhus:
> Do you mean to set it to `true` always?

Yes, setting it to `true` when `openAtLogin: false` should have no effect, so might as well make it simpler.

source/menu.ts
```diff
@@ -190,6 +190,14 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				focusedWindow.setAlwaysOnTop(item.checked);
 			}
 		},
+		{
+			label: 'Launch at Login',
```
sindresorhus: This feature only works on macOS and Windows, so you have to limit it to those system with the `visible` option.

https://electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows

whitecrownclown: Yep, I missed that, thanks!

sindresorhus:
Thanks :)

blargle-ragequit:
Why is this not on Linux?

CvX:
@blargle-ragequit Because the Electron API doesn’t support it: https://electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows

Here’s an open issue at the Electron repo: https://github.com/electron/electron/issues/15198

PR #775, opened 2019-02-12
Handle initial network connectivity

### LINKED ISSUE #421: Blank page after internet activated
if internet isn't activated/doesn't work when the app starts, the app stays blank after internet is back

---

Caprine 2.10.0
Electron 1.8.4
linux x64 4.14.30-1-MANJARO

whitecrownclown:
@sindresorhus ,

Hello!

This PR should fix the initial app launch issue with network connectivity and also allow for `Caprine` to be launched on system start/login without the need of a manual refresh. (Fixes #421)

Any feedback is welcome. Thanks!

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -288,7 +288,7 @@ function createMainWindow(): BrowserWindow {
 	tray.create(mainWindow);
 
 	// TODO: Reload mainWindow instead as soon as #712 is fixed
-	if (!(await isOnline()) && (await pWaitFor(() => isOnline()), {interval: 1000})) {
+	if (!(await isOnline()) && (await pWaitFor(isOnline), {interval: 1000})) {
 		app.relaunch();
 		app.quit();
 	}
--- a/source/index.ts
+++ b/source/index.ts
@@ -295,6 +295,7 @@ const networkConnectivity = async (): Promise<boolean> => {
 
 	const shouldReload = !(await isOnline());
 
+	// TODO: Reload mainWindow instead as soon as #712 is fixed
 	if (shouldReload && (await networkConnectivity())) {
 		app.quit();
 		app.relaunch();
--- a/source/index.ts
+++ b/source/index.ts
@@ -29,8 +29,8 @@ import config from './config';
 import tray from './tray';
 import {sendAction} from './util';
 import {process as processEmojiUrl} from './emoji';
+import ensureOnline from './ensure-online';
 import './touch-bar'; // eslint-disable-line import/no-unassigned-import
-import {ensureOnline} from './ensure-online';
 
 electronDebug({
 	enabled: true, // TODO: This is only enabled to allow `Command+R` because messenger sometimes gets stuck after computer waking up
@@ -278,7 +278,10 @@ function createMainWindow(): BrowserWindow {
 }
 
 (async () => {
-	await Promise.all([ensureOnline(), app.whenReady()]);
+	await Promise.all([
+		ensureOnline(),
+		app.whenReady()
+	]);
 
 	const trackingUrlPrefix = `https://l.${domain}/l.php`;
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -16,8 +16,6 @@ import {
 	RequestHeaders,
 	OnSendHeadersDetails
 } from 'electron';
-import isOnline from 'is-online';
-import pWaitFor from 'p-wait-for';
 import log from 'electron-log';
 import {autoUpdater} from 'electron-updater';
 import electronDl from 'electron-dl';
@@ -29,9 +27,10 @@ import {bestFacebookLocaleFor} from 'facebook-locales';
 import updateAppMenu from './menu';
 import config from './config';
 import tray from './tray';
-import {sendAction, showRetryDialog} from './util';
+import {sendAction} from './util';
 import {process as processEmojiUrl} from './emoji';
 import './touch-bar'; // eslint-disable-line import/no-unassigned-import
+import {ensureOnline} from './ensure-online';
 
 electronDebug({
 	enabled: true, // TODO: This is only enabled to allow `Command+R` because messenger sometimes gets stuck after computer waking up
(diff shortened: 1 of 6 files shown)


CvX:
I don’t think this is a correct way to approach connectivity problems.

1. Reloading the page is one of the triggers of the bug described in #712.
2. User could be interacting with the app when the reload happens which isn’t great user experience.
3. If someone’s connection keeps dropping the app would reload constantly. Not ideal either.

We have to find a different way to force Messenger to re-establish its connections.

whitecrownclown:
@CvX I know it's not a great way to do it, but this PR only handles initial connectivity, before the user has a change to interact with the app (for example, launching the app at login, when most of the time there's no internet connection and the app remains unresponsive/blank page). I'm not sure about item number 1, but definitely 2 and 3 would not apply here.

CvX:
Ah, my bad! 😶 In this case, you’re right, disregard 2 and 3.
And to handle 1 it might be better to do what we do when switching between standard messenger and work chat?

```js
app.relaunch();
app.quit();
```

whitecrownclown:
@CvX Yes, i could make that change, the reason why i didn't go for that at first is that to the user it will look more like a bug and it also takes a bit longer to load. Is there no way we could reload the mainWindow and find a fix for #712 ?

CvX:
That would be ideal, but I'm afraid it might take quite a while to fix #712, since it's possibly a bug in Chromium.

I'd suggest doing a full app relaunch for now and adding a TODO comment to remind us to change it to a simple reload as soon as #712 is fixed.

whitecrownclown:
@CvX Alright, i've done exactly that. Thanks!

source/index.ts
```diff
@@ -276,6 +277,13 @@ function createMainWindow(): BrowserWindow {
 	return win;
 }
 
+const networkConnectivity = async (): Promise<boolean> => {
```
CvX:
```suggestion
const networkConnectivity = async (): Promise<true> => {
```

source/index.ts
```diff
@@ -276,6 +277,13 @@ function createMainWindow(): BrowserWindow {
...
+	/* eslint-disable no-await-in-loop */
```
CvX:
```suggestion
	/* eslint-disable-next-line no-await-in-loop */
```

source/index.ts
```diff
@@ -276,6 +277,13 @@ function createMainWindow(): BrowserWindow {
...
+	while (!(await isOnline()));
+	/* eslint-enable no-await-in-loop */
```
CvX:
```suggestion
```

whitecrownclown:
@CvX I've updated the PR and also used `p-wait-for` instead of the `while` statement.

source/index.ts
```diff
@@ -285,6 +287,12 @@ function createMainWindow(): BrowserWindow {
 	mainWindow = createMainWindow();
 	tray.create(mainWindow);
 
+	// TODO: Reload mainWindow instead as soon as #712 is fixed
+	if (!(await isOnline()) && (await pWaitFor(isOnline), {interval: 1000})) {
```
CvX: 1. There's a bug – the options object with `interval` isn't actually passed into the `pWaitFor` 😉
2. It might be more readable to move the second await inside the block:

```suggestion
	if (!await isOnline()) {
		await pWaitFor(isOnline, {interval: 1000});
```

whitecrownclown: 1. I somehow missed that from too many parenthesis. 2. That's the reason i missed nr. 1

CvX: You can also get rid of the extra parens around `await isOnline()` (see the previous suggestion). 🙂

whitecrownclown: @CvX The linter won't really let me do that inside an if condition

CvX: Ah, my bad, some linter errors don't show up in my text editor due to a bug… 😓

sindresorhus:
This results in a weird experience if you're actually offline. Then the app is just on the Dock without anything happening or showing. I think we need a timeout after 20 (?) seconds where it shows an error message to the user that they're offline, and with a "try again" button.

source/index.ts
```diff
@@ -289,7 +289,11 @@ function createMainWindow(): BrowserWindow {
 
 	// TODO: Reload mainWindow instead as soon as #712 is fixed
 	if (!(await isOnline())) {
+		const connectivityTimeout = setTimeout(() => {
+			showRetryDialog(`It appears that you're offline`);
```
sindresorhus: > You appear to be offline. Caprine requires a working internet connection.

?

CvX:
I've tried the this PR on Windows, and a couple of times the relaunched app's window was blank. I didn't find out the root cause, but instead I decided to try a simpler solution: wait for the connection **before** creating the window. Let me know what you think @whitecrownclown, @sindresorhus.

whitecrownclown:
@CvX I think it's both simpler and better. No need to relaunch app/ reload window. I did notice that after this change triggering window 'activate' will throw and error so i added in a small check.

source/util.ts
```diff
@@ -33,3 +33,20 @@ export function showRestartDialog(message: string): void {
 		}
 	);
 }
+
+export function showRetryDialog(message: string): void {
```
sindresorhus: This is not really a generic utility. I would move this and https://github.com/sindresorhus/caprine/pull/775/files#diff-21e9ddd93162651bd36f6e5bbfca8460R282 into a new file to contain the logic.

So in `index.js`, it would just be:

```js
(async () => {
await ensureOnline();

await app.whenReady();

...
```

sindresorhus:
@CvX Good solution!

source/util.ts
CvX: btw. these awaits don't have to be sequential, so we can change that to, e.g.:

```js
await Promise.all([ensureOnline(), app.whenReady()]);
```

CvX:
I've renamed `showRetryDialog` to `showWaitDialog` and did some minor cleanup.

CvX:
@sindresorhus Prettier disagrees with you 😉

sindresorhus:
Good work everyone :)

source/util.ts
sindresorhus: @CvX There's not really any benefit in using `Promise.all` here as `app.whenReady()` doesn't do any work. It's just a shortcut for checking for the status of the app load. So the result of using `Promise.all` is exactly the same as before. But it's fine, we can keep it.

PR #781, opened 2019-02-13
Add a fallback for missing uiContextualLayerPositioner

### LINKED ISSUE #779: Message buttons not working on Mac
None of the bottom buttons respond on tap or show the emoji/sticker popup.

CvX:
Closes #779 (or at least workarounds some of the issues described there)

It seems that either Messenger is rolling out or A/B testing some document structure changes or we have a bug that affects only some Caprine users.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -21,17 +21,20 @@ async function withMenu(
 	menuButtonElement.click();
 
 	// Wait for the menu to close before removing the 'hide-dropdowns' class
-	const menuLayer = document.querySelector<HTMLElement>(
-		'.uiContextualLayerPositioner:not(.hidden_elem)'
-	)!;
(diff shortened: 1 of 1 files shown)


source/browser.ts
```diff
@@ -21,17 +21,20 @@ async function withMenu(
 	menuButtonElement.click();
 
 	// Wait for the menu to close before removing the 'hide-dropdowns' class
-	const menuLayer = document.querySelector<HTMLElement>(
-		'.uiContextualLayerPositioner:not(.hidden_elem)'
-	)!;
+	const menuLayer = document.querySelector('.uiContextualLayerPositioner:not(.hidden_elem)');
```
whitecrownclown: I was about to open a PR about this as well. I used `elementReady` when getting `menuLayer`. It's either not present in the UI at that moment or, as you said, messenger running some A/B tests.

PR #787, opened 2019-02-16
Add option to hide names and avatars

### LINKED ISSUE #768: Feature to hide name of current conversation (people, group)
Hi,
When I use Messenger on my laptop in public, I wish that it has feature to hide the name of current conversation I'm focus on (even the tiny profile picture beside every message).
I can think of a button or an option of the context menu when I right click anywhere in the conversation area.
Hope our community will make this feature come true.

alexandrepa:
Fixes #768 
Add a new option (Private Mode) that hides names and pictures.


source/menu.ts
```diff
@@ -288,6 +288,17 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				sendAction('set-dark-mode');
 			}
 		},
+		{
+			label: 'Private Mode',
+			id: 'privateMode',
+			type: 'checkbox',
+			checked: config.get('privateMode'),
+			accelerator: 'CommandOrControl+N',
```
sindresorhus: This shortcut is already taken. I'm not sure this needs a shortcut and if it does, it should be something obscure, so it doesn't conflict with any existing or future shortcuts.

css/private.css
```diff
@@ -0,0 +1,18 @@
...
+	display: none;
+}
+
+/* Person icon */
+html.private-mode ._1ht1 ._1qt3,
+html.private-mode ._1t_q ._4ld- img,
+html.private-mode ._4jzq._jf4 img,
+html.private-mode  ._5bli._2_a2.img {
+	display: none;
+}
```
sindresorhus: I don't think this needs a separate CSS file. Just put in `browser.css`.

sindresorhus:
@alexandrepa Still interested in finishing this?

alexandrepa:
Yes l'll try to do the modification this weekend !

alexandrepa:
@sindresorhus I've just made the modifications !

source/menu.ts
```diff
@@ -315,6 +315,17 @@ Press Command/Ctrl+R in Caprine to see your changes.
...
+			label: 'Private Mode',
+			id: 'privateMode',
+			type: 'checkbox',
+			checked: config.get('privateMode'),
+			accelerator: 'CommandOrControl+Shift+N',
+			click() {
+				config.set('privateMode', !config.get('privateMode'));
+				sendAction('set-private-mode');
+			}
+		},
```
sindresorhus: I think it should be added on line 337 with a separator after it. It doesn't really fit into the dark mode style group.

css/browser.css
```diff
@@ -320,3 +320,25 @@ body.show-message-buttons ._4_j4 ._39bj {
...
+}
+
+/* Person icon */
+html.private-mode ._1ht1 ._1qt3,
+html.private-mode ._1t_q ._4ld- img,
+html.private-mode ._4jzq._jf4 img,
+html.private-mode ._5bli._2_a2.img,
+html.private-mode ._5743._6y4y {
+	display: none;
+}
```
sindresorhus: The style that hides the navbar name and icon should use `visibility: hidden` so it doesn't move around unrelated UI components.

source/menu.ts
```diff
@@ -315,6 +315,17 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				sendAction('set-dark-mode');
 			}
 		},
+		{
+			label: 'Private Mode',
```
sindresorhus: Can we come up with a better name than `Private Mode`?

Some ideas:

- Hide Names and Avatars
- Hide Identifies
- ?

sindresorhus:
Can you mention it in the readme and fix the merge conflict?

source/menu.ts
marcinlesek: `HideMode`?

sindresorhus:
Bump :)

alexandrepa:
I'm gonna work on it soon !! :)

alexandrepa:
@sindresorhus Ready to be merged :) !

css/browser.css
```diff
@@ -366,3 +366,25 @@ body.show-message-buttons ._4_j4 ._7mkk {
...
 }
+
+/* Private Mode */
+/* Contact list: person name */
+html.private-mode ._1ht6 {
+	display: none;
```
CvX: It be better to have it `visibility: hidden`, so it doesn’t affect the layout.

css/browser.css
```diff
@@ -366,3 +366,25 @@ body.show-message-buttons ._4_j4 ._7mkk {
...
+}
+
+/* Messages list: person name */
+html.private-mode ._17w2 ._3oh-,
+html.private-mode ._41ud ._ih3 {
+	display: none;
```
CvX: Same here: It be better to have it `visibility: hidden`, so it doesn’t affect the layout.

css/browser.css
```diff
@@ -366,3 +366,25 @@ body.show-message-buttons ._4_j4 ._7mkk {
...
+}
+
+/* Person icon */
+html.private-mode ._1ht1 ._1qt3,
+html.private-mode ._1t_q ._4ld- img,
+html.private-mode ._4jzq._jf4 img,
+html.private-mode ._5bli._2_a2.img,
+html.private-mode ._5743._6y4y {
+	visibility: hidden;
```
CvX: Here, I’d rather have those images blurred, to be able to differentiate them, but that could be done later.

CvX: - [ ] Actually, blur as opposed to hiding is essential when using the compact mode.
So: profile pictures should have `filter: blur(Npx)` instead (where N depends on the size of the image, I’d say 5px for large and 1-2px for small?)
- [ ] I believe `._1ht1 ._1qt3` and `._4jzq._jf4 img` could be rolled into one: `._4ld- .img`
- [ ] Can you annotate, inline each of those selectors? (e.g. `/* Contacts list picture */`, `/* Message read indicator */`)

alexandrepa: I couldn't find a way to aggregate css selectors, `_4ld-` is only used by one type of picture.

CvX: > I couldn't find a way to aggregate css selectors, `_4ld-` is only used by one type of picture.

Okay, that might be an old UI/new UI difference.

But could you change the “Contacts list picture” selectors to match `` elements and not the elements higher in the DOM tree? This way the images are blurred but retain the sharp border:

Screen Shot 2019-07-02 at 14 19 35
Screen Shot 2019-07-02 at 14 19 08

CvX:
@alexandrepa I’ve added two minor tweaks. Please take a look to make sure that didn’t brake anything on your end. 🙂

sindresorhus:
I'm still seeing names in the conversation list (`Im` and `Vigdis`):

Screen Shot 2019-07-04 at 23 23 06
Screen Shot 2019-07-04 at 23 23 20

CvX:
```css
._1htf {
  visibility: hidden;
}

._1htf > span:first-child {
  visibility: visible;
}
```

Should do the trick, though Message previews are not left-aligned, and we lose the “You:” prefixes.

alexandrepa:
I just found one issue with selecting the "Contacts list picture" with `` element. When it's a group conversation and there is a custom group picture, the element is not an `` but a `<div>` with a background so the picture is not blurred.

CvX:
Hm, that definitely depends on the version of Messenger UI. In mine, those are composed of separate `` elements, so it works alright:

group conversation image

alexandrepa:
Are you sure even if you add a custom group photo ? Like in this video : https://www.youtube.com/watch?v=d-_J4v7fWr0

CvX:
Nope! 😄 I glanced over the mention of “custom” because I didn’t even know that was a thing! 🙂 

You’re right. `html.private-mode ._1ht1 ._1qt3 ._4ld- div` should do the trick?

alexandrepa:
Yes but if I do that it will affect the last changed I did in this [commit](https://github.com/sindresorhus/caprine/pull/787/commits/9085f761a50cb86b0335f7afddaf6ed580a04b0d) and the blur effect will be applied on the whole div.

CvX:
Those two selectors, when used together, should work.
One targets the usual profile pic, and the other – the innermost div of group chat image.

```css
/* Contacts list picture */
html.private-mode ._1ht1 ._1qt3 img,
/* Contacts list group chat custom image */
html.private-mode ._1ht1 ._1qt3 ._4ld- div,
/* Navbar information */
html.private-mode ._5743._6y4y {
	filter: blur(5px);
}
```

alexandrepa:
For me `._1ht1 ._1qt3 ._4ld- div` is not returning anything, the class `_4ld-` is not present in contact list pictures.

CvX:
Okay. In my version it’s:

```css
div._1qt3._5l-3 div div._4ldz div._4ld-[has border radius 50%] div[has bg image]
```

CvX:
Ohh, I like the `div[style*='background-image']`! 😃

alexandrepa:
Some news ?

CvX:
CC: [anyone with the new UI] 😉 

@sindresorhus @marcinlesek ? 🙂

sindresorhus:
@CvX What do you need me to do?

CvX:
@sindresorhus this PR looks good on my end. Can you double check with the new Messenger UI? 🙂

sindresorhus:
Thanks :)

PR #790, opened 2019-02-17
Fix "missed call" styling in dark mode

### LINKED ISSUE #789: Dark mode - Display Issue - Missed Call
image

Underneath the white line there is a dark line of text for the amount of time ago it was.

It needs styling to be clearer.

whitecrownclown:
Fixes #789


whitecrownclown:
> 👍
> 
> "Missed Call" dialog before/after
> screen shot 2019-02-18 at 15 53 03 screen shot 2019-02-18 at 15 56 55

I really like that you post screenshots with before and after, I should do that as well.

PR #794, opened 2019-02-20
Use only the memoize module from lodash

janosorcsik:


sindresorhus:
I think using Lodash is fine here as it makes it easy for us to additional utilities in the future. I would worry about it in reusable packages, but less so here.

felixfbecker:
Agree with @sindresorhus

PR #800, opened 2019-02-24
[Feature] Option to hide tray icon

### LINKED ISSUE #684: Optionally hide tray icon?
Is there a way to optionally hide/not use the tray icon at all? The 'Quit on window close' option does half of what i'd want in this feature suggestion.

Generally, I always have Caprine open, along side several other messenger apps, so the window is not even minimized most of the time and the icon simply adds clutter for me.

How complicated would it be to simply not use an icon and just live from the taskbar entry until closed?

whitecrownclown:
Fixes #684

@sindresorhus @CvX  Let me know what you think and if this is something you want added into `Caprine`. Thanks!


sindresorhus:
When `Quit on window close` is enabled, should we even show the tray icon at all? Since the window is not minimized to the tray then, it doesn't really make sense to have it in the tray, right?

whitecrownclown:
It makes sense if you minimize the window instead. Not showing the tray might also confuse the users who don’t toggle the specific option (unless we update the ui, i suppose).

sindresorhus:
> It makes sense if you minimize the window instead.

I'm confused. When you click the minimize button in the window frame, the window is minimized to the Dock/Taskbar, not to the tray.

zakius:
Maybe there should be separate options to `Minimize to Tray` and `Close to Tray` but it's hard to say

sindresorhus:
> I'm confused. When you click the minimize button in the window frame, the window is minimized to the Dock/Taskbar, not to the tray.

@whitecrownclown

whitecrownclown:
@sindresorhus I think we could go either way. Hitting close without `Quit on window close` minimizes to tray. When you have `Quit on window close` and minimize, it will go to dock/taskbar, however you can still access it from tray if you want. I don't know which way is better.

sindresorhus:
> Edit: I did make the change, please have a look. Thanks!

What change?

And I'm still confused what problem this solves.

Look at the diff. You have included a lot of unrelated stuff.

whitecrownclown:
> > Edit: I did make the change, please have a look. Thanks!
> 
> What change?
> 
> And I'm still confused what problem this solves.
> 
> Look at the diff. You have included a lot of unrelated stuff.

Sorry, the unrelated changes are from a merge commit, i should rebase instead. I'll update the PR.

sindresorhus:
There's no need to open a new PR. You can always fix it locally and then just force push. Creating needless PRs creates noise in the issue tracker.

PR #374, opened 2018-01-22
Add menu bar mode on macOS

vanshg:
Building off of @antonio-ramadas's work in https://github.com/sindresorhus/caprine/pull/231, this PR fixes the Icon Sizes, swaps the incorrect read/unread menu bar icon indicators, renames Continuity to Menu Bar Mode in the View Menu, updates the version number, and merges in all latest changes.

PR #804, opened 2019-03-01
Add menu bar mode for macOS

### LINKED ISSUE #206: Add menu bar icon on macOS

whitecrownclown:
@sindresorhus @CvX This PR continues the work from #374

If you could please have a look, I've fixed most/all of the issues @sindresorhus addressed.

Thanks!

Fixes #206

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#206: Add menu bar icon on macOS](https://issuehunt.io/repos/42574339/issues/206)
---

</details>


CvX:
Not a full review yet, just some thoughts after trying this out for the first time:

1. The menubar icon's shape should be based on the current app icon, i.e. not as wide. Also (this is quite subjective) I'd make the outline thicker, add the Messenger's "zig-zag" to both icon states and would make the unread one blue-ish (like e.g. Tweetbot: Tweetbot menu bar icon)
2. I'd replace any mentions of "darwin" with just "macos"
3. Would move as much of new code (e.g. `setupMenuBarMode`) to either `tray.ts` or a new file to keep it encapsulated (for an example, see `emoji.ts`).
4. "Disable Menu Bar Mode" closes the app

Other than that, looks cool! 🙂

whitecrownclown:
@CvX Hmm, I was almost sure I took care of 4). Thanks for the suggestions and obviously, your time, I'll update the PR.

source/tray.ts
```diff
@@ -59,10 +59,13 @@ export default {
 	},
 
 	destroy: () => {
-		if (tray) {
-			tray.destroy();
-			tray = null;
-		}
+		// Workaround for electron#14036
```
sindresorhus: Can you use a full link to the issue, so it's more easily clickable?

sindresorhus:
If I activate menu bar mode, there's no way to access the menu items. I think it should still show the menu when the window is focused.

sindresorhus:
> The menubar icon's shape should be based on the current app icon, i.e. not as wide. 

👍 

> Also (this is quite subjective) I'd make the outline thicker

👍 

> add the Messenger's "zig-zag" to both icon states and would make the unread one blue-ish (like e.g. Tweetbot: Tweetbot menu bar icon)

👎 The "zig-zag" is the indicator of a new message. It should stay. Menu bar icons should not be colored.

---

The menu bar icon is also slightly to large and not fully vertically centered.

Screen Shot 2019-03-10 at 23 17 13

whitecrownclown:
Thanks, i’ll update it this week!

whitecrownclown:
> If I activate menu bar mode, there's no way to access the menu items. I think it should still show the menu when the window is focused.

Apparently there's no way to do this at the moment. When calling `app.dock.hide()`, the menu items are also hidden. [issue](https://github.com/electron/electron/issues/592)

whitecrownclown:
@sindresorhus 	I've updated the icons. This is how they look now.

`read`
read

`unread`
unread

sindresorhus:
> Apparently there's no way to do this at the moment. When calling app.dock.hide(), the menu items are also hidden.

Hmm, that's too bad. It would mean there's no way to access the menus, and also no keyboard shortcuts would work as they are registered in the menu...

sindresorhus:
The menu bar icon is still too large and it should be updated to the style of the app icon, meaning softer lightening symbol (just copy-paste it from the icon), softer arrow under bottom-left of the circle, and the circle shape is different.

whitecrownclown:
@sindresorhus I've updated the icons, size is 16px (previously 18px). I have the vector assets now so if we need changes, it'll be easier. The icon shape should be basically the same as `caprine` icon. Couldn't find a vector for it so i traced the png in inkscape instead. Let me know what you think, thanks!

`read`
read

`unread`
unread

sindresorhus:
> Couldn't find a vector for it so i traced the png in inkscape instead.

It's right here: https://github.com/sindresorhus/caprine/blob/master/media/Icon.svg

sindresorhus:
The icon positions are not the same:

foo

whitecrownclown:
> > Couldn't find a vector for it so i traced the png in inkscape instead.
> 
> It's right here: [/media/Icon.svg@`master`](https://github.com/sindresorhus/caprine/blob/master/media/Icon.svg)

@sindresorhus 	Thanks, I've updated it based on the same app icon `media/Icon.svg`.

menubar

sindresorhus:
We still need to find a way to resolve: https://github.com/sindresorhus/caprine/pull/804#issuecomment-473781977

How about if "menu bar mode" doesn't imply hiding the Dock icon? When the user enables menu bar mode, we enable a menu item that allows hiding the Dock icon, and when the user selects that, we show a dialog letting them know that keyboard shortcuts will not work. We also let the user access the normal menu from the menu bar icon menu:

```
Disable Menu Bar Mode
--------------
Menu >
--------------
Quit
```

// @CvX Thoughts?

sindresorhus:
> Thanks, I've updated it based on the same app icon media/Icon.svg.

👍 Looks good.

CvX:
Yeah, splitting that into two separate options is common in macOS apps: "Show Menu Bar icon" and "Show Dock icon"

Here's how it appears in Fantastical preferences:
Fantastical options

I'm for it! Adding all app menus to the menu bar icon's context menu is also a great idea.

But it would be nice to make keyboard shortcuts work also without the menu bar.

sindresorhus:
> Yeah, splitting that into two separate options is common in macOS apps: "Show Menu Bar icon" and "Show Dock icon"

👍 

> But it would be nice to make keyboard shortcuts work also without the menu bar.

I'm open to ideas. I don't see how we can do that though... Maybe a `Electron.MenuItem` wrapper that also register accelerators using https://www.npmjs.com/package/electron-localshortcut when in menubar mode. Although, that's ugly. Hoping for a better solution than that. We can tackle that later on though.

sindresorhus:
@whitecrownclown This is so close to getting merged. Interested in finishing it up?

whitecrownclown:
Yes, will work on it.

whitecrownclown:
@sindresorhus I've added the menu into the tray when enabling `MenuBarMode` and the shortcuts seem to work just fine. Please have a look.

CvX:
I've tried it out and it's looking great! 😃🎉

We're still going through with splitting the "Menu bar mode" into two separate options ("Show Menu Bar icon" and "Show Dock icon"), right?

vanshg:
@whitecrownclown Was just about to pick back up where I left off on https://github.com/sindresorhus/caprine/pull/374, when I saw this. Thank you for continuing this work!

whitecrownclown:
@vanshg Hey, thanks for starting #374 

I've separated the two options. When toggling `Show Menu Bar Icon` it will enable `Show Dock Icon` and it works, but I have one issue : 

If user has `Show Menu Bar Icon` true and `Show Dock Icon` false, when restarting `Caprine`, the app will start with the dock icon and then it would need to be hidden using : `app.dock.hide()` and it doesn't look great. I couldn't find a way around except this https://github.com/electron-userland/electron-builder/issues/1456#issuecomment-293518692 but i don't think it applies to this use case.

@CvX @sindresorhus Thoughts? Also have a look at the changes. Thanks!

sindresorhus:
If we set https://github.com/electron-userland/electron-builder/issues/1456#issuecomment-293518692, can we then show the Dock icon? If so, we could just do the reverse of what we're doing.

readme.md
```diff
@@ -19,6 +19,7 @@ Caprine is an unofficial and privacy-focused Facebook Messenger app with many us
 - [Vibrant theme](#vibrancy)\*
 - [Privacy-focused](#privacy)
 - [Keyboard shortcuts](#keyboard-shortcuts)
+- [Menu Bar mode](#menu-bar-mode)
```
sindresorhus:
```suggestion
- [Menu Bar mode](#menu-bar-mode)\*
```

source/index.ts
```diff
@@ -86,13 +87,17 @@ app.on('second-instance', () => {
...
+}
+
 function updateBadge(conversations: Conversation[]): void {
 	// Ignore `Sindre messaged you` blinking
 	if (!Array.isArray(conversations)) {
 		return;
 	}
 
-	const messageCount = conversations.filter(({unread}) => unread).length;
+	const messageCount: number = getMessageCount(conversations);
```
```suggestion
	const messageCount = getMessageCount(conversations);
```

source/index.ts
```diff
@@ -31,6 +31,7 @@ import {sendAction, sendBackgroundAction} from './util';
 import {process as processEmojiUrl} from './emoji';
 import ensureOnline from './ensure-online';
 import './touch-bar'; // eslint-disable-line import/no-unassigned-import
+import {setupMenuBarMode} from './menu-bar-mode';
```
```suggestion
import {setUpMenuBarMode} from './menu-bar-mode';
```

source/menu-bar-mode.ts
```diff
@@ -0,0 +1,46 @@
+import {app, globalShortcut, BrowserWindow, Menu} from 'electron';
+import {is} from 'electron-util';
+
+import config from './config';
+import tray from './tray';
+
+const menuBarShortcut = 'CommandOrControl+Shift+Y';
```
sindresorhus: This is only for macOS, so we don't need `Control`.

sindresorhus: This needs to be documented in the keyboard and menu bar mode section in the readme.

source/tray.ts
```diff
@@ -1,44 +1,84 @@
 import * as path from 'path';
 import {app, Menu, Tray, BrowserWindow} from 'electron';
 import {is} from 'electron-util';
+import config from './config';
+import {toggleMenuBarMode} from './menu-bar-mode';
 
 let tray: Tray | null = null;
+let previousMessageCount = 0;
+
+let contextMenu: any;
```
sindresorhus: Don't use `any`.

source/tray.ts
```diff
@@ -1,44 +1,84 @@
...
-			} else {
-				win.show();
-			}
-		};
-
-		const contextMenu = Menu.buildFromTemplate([
+		contextMenu = Menu.buildFromTemplate([
 			{
-				label: 'Toggle',
+				label: 'Disable Menu Bar Mode',
```
sindresorhus: What about Linux and Windows?

readme.md
```diff
@@ -142,6 +143,11 @@ Note that you can change the behavior of Caprine so that the app closes when the
 
 Advanced users can modify the colors/styles of Caprine. Click the menu item `Caprine`/`File` → `Caprine Settings` → `Advanced` → `Custom Styles` and a CSS file will open up in your default editor.
 
+### Menu Bar Mode *(macOS only)*
+
+You can remove the app from the dock, create a menu bar icon and have a global shortcut to toggle the window. Caprine will show above any app, even if they are full screen, and it'll hide when is not focused.
+You can enable this in the menu bar options and disable in the options of the menu bar icon (right click on it).
```
sindresorhus: This should better explain where to find those two options and should also mention that keyboard shortcuts will not work when the Dock icon is hidden. It should also explain where to find the menu then.

Would also be good with a screenshot of the menu bar icon.

sindresorhus:
In the menu bar icon menu, the "Menu" item should only be visible if the Dock icon is hidden.

---

When the user unchecks the "Show Dock Icon" preference, we should show a dialog explaining to the user that keyboard shortcuts will no longer work, and that they can access the menu by right-clicking the menu bar icon.

---

Would the "Hide Dock Icon" maybe be better placed in the menu bar icon menu? Since it only really applies when the menu bar icon is shown.

whitecrownclown:
I agree with 1&3
2) Shortcuts work even if we hide the Dock Icon since we have the Menu
registered in the Menu bar.

On Mon, 17 Jun 2019 at 14:24, Sindre Sorhus <notifications@github.com>
wrote:

> In the menu bar icon menu, the "Menu" item should only be visible if the
> Dock icon is hidden.
>
------------------------------

sindresorhus:
> 2) Shortcuts work even if we hide the Dock Icon since we have the Menu
registered in the Menu bar.

Oh wow. You're right. I didn't think that would work. Great!

whitecrownclown:
> If we set [electron-userland/electron-builder#1456 (comment)](https://github.com/electron-userland/electron-builder/issues/1456#issuecomment-293518692), can we then show the Dock icon? If so, we could just do the reverse of what we're doing.

I think it might work. I'll give it a shot.

css/dark-mode.css
```diff
@@ -267,9 +267,6 @@ html.dark-mode ._1ht6 {
 }
 
 /* Contact list: person (selected) window inactive */
-html.dark-mode.is-window-inactive ._1ht1._1ht2 {
-	background: var(--base-twenty) !important;
-}
```
sindresorhus: ?

readme.md
```diff
@@ -19,6 +19,7 @@ Caprine is an unofficial and privacy-focused Facebook Messenger app with many us
 - [Vibrant theme](#vibrancy)\*
 - [Privacy-focused](#privacy)
 - [Keyboard shortcuts](#keyboard-shortcuts)
+- [Menu Bar mode](#menu-bar-mode)\*
```
sindresorhus:
```suggestion
- [Menu bar mode](#menu-bar-mode)\*
```

readme.md
```diff
@@ -142,6 +143,14 @@ Note that you can change the behavior of Caprine so that the app closes when the
 
 Advanced users can modify the colors/styles of Caprine. Click the menu item `Caprine`/`File` → `Caprine Settings` → `Advanced` → `Custom Styles` and a CSS file will open up in your default editor.
 
+### Menu Bar Mode *(macOS only)* <img src="media/screenshot-menu-bar-mode.png" width="40">
+
+You can create a menu bar icon and have a global shortcut to toggle the window (<kbd>Cmd</kbd> <kbd>Shift</kbd> <kbd>y</kbd>).You can also remove the app from the dock by toggling the menu item `Hide Dock Icon` from the menu bar icon.
```
sindresorhus: You should explain where to find the setting for enabling the menu bar icon. And I think the text could be better written.
```suggestion
You can enable a menu bar icon and have a global shortcut to toggle the window (<kbd>Cmd</kbd> <kbd>Shift</kbd> <kbd>y</kbd>). You can also remove the app from the dock by toggling the menu item `Hide Dock Icon` from the menu bar icon.
```

readme.md
```diff
@@ -142,6 +143,14 @@ Note that you can change the behavior of Caprine so that the app closes when the
...
+This will also make the app `Menu` available as a menu item inside the menu bar icon.
+
+<div align="center">
+<img src="media/screenshot-menu-bar-menu.png" width="200"></div>
```
sindresorhus:
```suggestion
	<img src="media/screenshot-menu-bar-menu.png" width="200">
</div>
```

source/index.ts
```diff
@@ -289,7 +296,9 @@ function createMainWindow(): BrowserWindow {
 
 	await updateAppMenu();
 	mainWindow = createMainWindow();
-	tray.create(mainWindow);
+
+	// Start in Menu Bar mode if enabled, otherwise start normally
```
```suggestion
	// Start in menu bar mode if enabled, otherwise start normally
```

sindresorhus:
Almost there! Just some minor nitpick left :)

sindresorhus:
Have to tested the app on Linux or Windows too? To ensure the tray functionality still works there.

whitecrownclown:
@sindresorhus 	I did look on Windows, but can't check it on Linux. I can have another look on Windows, just to make sure. I'll also update with what you reviewed. Thanks!

css/dark-mode.css
whitecrownclown: Hm, this is how it looks for me:
Screenshot 2019-06-20 at 22 48 41

sindresorhus:
Yay! Great to finally have this in. Nice work, @whitecrownclown 🙌

PR #811, opened 2019-03-05
Fix error on quit caused by Electron race issue

### LINKED ISSUE #809: Error on quit
capture d ecran 2019-03-05 a 16 00 11
when I minimize the app and right click on the icon and click on quit this error show up

whitecrownclown:
Fixes #809

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -457,7 +457,9 @@ app.on('activate', () => {
 
 app.on('before-quit', () => {
 	isQuitting = true;
-	config.set('lastWindowState', mainWindow.getNormalBounds());
+	if (mainWindow) {
+		config.set('lastWindowState', mainWindow.getNormalBounds());
+	}
 });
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
@whitecrownclown Are you able to reproduce the issue?

sindresorhus:
The issue makes no sense as the windows are not supposed to be quit until after the `before-quit` event is called:

> Emitted before the application starts closing its windows. Calling event.preventDefault() will prevent the default behaviour, which is terminating the application.
> \- https://electronjs.org/docs/api/app#event-before-quit

whitecrownclown:
@sindresorhus No, I couldn't reproduce this issue and I agree, it doesn't make any sense. I thought there may be an electron issue which could cause it but couldn't find anything related.

sindresorhus:
Merging this as more people are experiencing the issue...

Issue #568, opened 2018-08-21
Caprine cannot detect built-in webcam


trmaphi:
Tested webcam with cheese

System info
-------------- 
OS: Ubuntu 18.04.1 LTS x86_64 
Host: 42361S9 ThinkPad T420 
Kernel: 4.15.0-32-generic 
Uptime: 3 hours, 28 mins 
Packages: 1760 
Shell: bash 4.4.19 
Resolution: 1920x1080 
DE: GNOME 3.28.2 
WM: GNOME Shell 
WM Theme: Adwaita 
Theme: Materia-light [GTK2/3] 
Icons: Flat-Remix-Light [GTK2/3] 
Terminal: gnome-terminal 
CPU: Intel i7-2620M (4) @ 3.400GHz 
GPU: Intel 2nd Generation Core Proce 
Memory: 3737MiB / 7749MiB

dbisk:
I was also on Ubuntu 18.04.1 and it worked fine. Does it work if you use facebook messenger from a browser?

trmaphi:
Facebook messenger works normaly with the webcam.

trmaphi:
@dbisk How do you install the caprine? I installed it via snap.
I did build from the source the camera work normal!

dbisk:
I installed it using snap. Mine seems to work fine though, so I'm not sure if the reason is snap...

trmaphi:
@dbisk  I had a workaround, that's put `--classic`. The reason is security confinement of [snap](https://docs.snapcraft.io/reference/snap-command#command-options-1)
```
snap install --classic caprine
```

dbisk:
@trmaphi Yes, you're correct. I forgot I installed with the `--classic` option.

PR #813, opened 2019-03-07
Add `camera` plug to Snap build config

jstevans:
Camera doesn't work for me when installing from the snap store, and it looks like the reason is because it doesn't have permissions. The [old workaround of `snap install --classic`](https://github.com/sindresorhus/caprine/issues/568) no longer works as of `snapd@2.36.2` ([relevant PR](https://github.com/snapcore/snapd/pull/6039)). 

Reading the relevant docs for [electron-builder](https://www.electron.build/configuration/snap) and [snap](https://docs.snapcraft.io/interface-management/6154), it seems like the right solution is to use the default confinement model and to explicitly specify a `snap.plugs` that includes `camera` in the `electron-builder` config.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,12 @@
 			"description": "Caprine is an unofficial and privacy focused Facebook Messenger app with many useful features.",
 			"category": "Network;Chat"
 		},
+		"snap": {
+			"plugs": [
+				"default",
+				"camera"
+			]
+		}
 		"win": {
 			"verifyUpdateCodeSignature": false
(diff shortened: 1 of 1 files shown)


sindresorhus:
Any other plugs we might need? https://docs.snapcraft.io/supported-interfaces/7744

jstevans:
Not too sure -- I'm afraid I'm a recent user and haven't read much of the code here yet. Nothing jumped out at me, though 🙂

PR #816, opened 2019-03-14
Allow navigation to alternate messenger domains

dhharris:
I was having an issue where navigating to something like `prod.messenger.com` would open in my browser. I'd like to be able to do that in the Caprine app, so I changed the navigation rules to allow any hostname that contains `messenger.com`.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -394,7 +394,7 @@ function createMainWindow(): BrowserWindow {
 	webContents.on('will-navigate', (event, url) => {
 		const isMessengerDotCom = (url: string): boolean => {
 			const {hostname} = new URL(url);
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think it would be better to use `String#endsWith`.

dhharris:
Sounds good to me. I'll edit the diff.

CvX:
Hm, this change makes it match URLs like e.g. "https://my.classmessenger.com/login" 😉
It should probably check for the dot in the hostname too, i.e. `hostname.endsWith('.messenger.com')`.

dhharris:
Sorry for the delay, I was on vacation :)

PR #822, opened 2019-03-20
Add `License Agreement` screen on Windows installer

klaussinani:
### Description

The PR enables the `nsis` installer to display the MIT project license to the user, through a `License Agreement` screen/step, before proceeding on with the installation.
The main purpose of this one is to provide the installation process with a more Windows-native feel.

More info regarding the `license` option: https://www.electron.build/configuration/nsis

This can be considered a fairly subjective addition, thus feel free to reject it if it doesn't look appealing enough : )

### Preview

<p align="center"></p>

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,9 @@
 			"description": "Caprine is an unofficial and privacy focused Facebook Messenger app with many useful features.",
 			"category": "Network;Chat"
 		},
+		"nsis": {
+			"license": "license"
+		},
 		"snap": {
 			"plugs": [
 				"default",


sindresorhus:
> provide the installation process with a more Windows-native feel.

I don't think this is a good thing.

sindresorhus:
I'm going to pass on this as I would prefer to have as much of a streamlined experience as possible. Just because Windows users are used to a bad UX of having to click through a multi-page wizard when installing, doesn't mean we can't do better.

sindresorhus:
Thanks for the suggestion though :)

Issue #584, opened 2018-09-16
Preferences doesn't launch


DenyDarko:
Well.. My issue is pretty much described on the title.

Browsing Edit -> Preferences does nothing. Nothing appears

Tried as elevated user too just in case, same result.

Thoughs?😊

---

Caprine 2.18.0
Electron 2.0.4
win32 x64 10.0.17134

dusansimic:
Fixed by #841

PR #841, opened 2019-04-04
Fix preferences not launching on small window sizes

whitecrownclown:
#584 I managed to reproduce this issue by resizing the app.


CvX:
It isn't a correct solution as it reverts fixes done in #732 (see https://github.com/sindresorhus/caprine/pull/732#discussion_r249214072)

whitecrownclown:
@CvX Thanks for letting me know. I wouldn't have known about #732. Thankfully, I've found the specific issue that blocked preferences from opening.

PR #860, opened 2019-04-15
Support the workplace.com domain

### LINKED ISSUE #845: Workplace domain change
Workplace has changed domain from `companyname.facebook.com` to `companyname.workplace.com`. So far the change has to be approved by an admin, but later it will happen for everybody I think.

After this change I can no longer login, I just get redirected to `companyname.workplace.com/chat` in Chrome. I'm guessing it's because the domain is hardcoded in some places (index file for example). Let me know if you need more info to fix this issue.

Jonher937:
This seems to fix #845 for me.
Built and tested on Linux.

Redirects are as follows:
* work.facebook.com (initial page loaded by caprine)
* work.workplace.com 
* companyname.workplace.com

More info on the domain switch: 
* https://www.facebook.com/help/work/1116975388401705?helpref=hc_fnav
* https://www.facebook.com/workplace/resources/deepen/prepare-community-workplace-domain

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -407,7 +407,8 @@ function createMainWindow(): BrowserWindow {
 		const isWorkChat = (url: string): boolean => {
 			const {hostname, pathname} = new URL(url);
 
-			if (hostname === 'work.facebook.com') {
+			// Facebook is now redirecting to the workplace domain while tenants are transitioning to the new domain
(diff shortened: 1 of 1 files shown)


Jonher937:
Sorry for the mess of commits, now that I'm done struggling with the syntax and lint it should work.
Thanks

Issue #851, opened 2019-04-10
Image preview sizes are too big


gklka:
<img width="839" alt="Screenshot_2019-04-10_at_16_55_42" src="https://user-images.githubusercontent.com/866694/55889942-27b52500-5bb2-11e9-8011-57f6b6980b52.png">

I think the image preview size should be maximized to the 60-90% of the available space and never should be cropped by the viewport. Facebook clearly optimizes these for full screen windows, but macOS users tend to use apps in smaller window sizes.

I am not sure, but it seems to me that happens only on own images/gifs, and not for the other participants in the chat.

CvX:
> I am not sure, but it seems to me that happens only on own images/gifs, and not for the other participants in the chat.

I've tested this, and it seems that it's just that some gifs are actually quite large ("cigar homer" being one of the biggest) and Messenger doesn't doesn't scale these images at all. So, they are shown in their original size, regardless of who sent them.

Anyway, I agree that it would be best to ensure any images are scaled down beyond some threshold. I'll add that to my TODO list. 😉

gklka:
Thank you for testing it out. My experience is that this happens more than once a day, and not only with GIFs but images dragged into Messenger or captured with my iPhone (and sent from there) too.

sadmanf:
Hi there! I submitted a PR for this issue over at #861 

I set the image/GIF sizes to be be contained within the boundaries of the screen. I also included before and after pictures. Feel free to take a look and let me know what you think.

gklka:
Perfect, thank you!

c0mrx:
Just wanted to add that a quick fix for this problem is to open up custom styles and use this rule:

Go to Caprine > Caprine Preferences > Advanced > Custom Styles and paste the following CSS rule at the bottom of the file.

```
div[role="presentation"] * {
  max-width: 100% !important;
}
```

Restart the app and it should be fixed! Note that I've tested this briefly, so there may be some unexpected behaviour. If so, remove the rule and wait for the PR to be merged.

Cheers!

PR #861, opened 2019-04-15
Resize image previews to fit in screen

sadmanf:
Referring to issue #851 

Small fix to make sure that image previews are resized based on the window size.

### DIFF
--- a/css/browser.css
+++ b/css/browser.css
@@ -264,13 +264,12 @@ a,
 	border-radius: 8px;
 }
 
-/* Contact list: person (selected) window inactive */
-._1ht1._1ht2.inactive {
-	background: #D2D2D2 !important;
-	color: #000000 !important;
+/* Contact list: person (selected) */
+._1ht1._1ht2 {
+	background: var(--selected-conversation-background) !important;
 }
-._1ht1._1ht2.inactive * {
-	color: rgba(0, 0, 0, 0.8);
+._1ht1._1ht2 * {
+	color: rgba(255, 255, 255, 0.8);
 }
 
 /* Remove top Facebook cookie banner */
--- a/css/browser.css
+++ b/css/browser.css
@@ -314,3 +314,8 @@ body.show-message-buttons ._4_j4 ._39bj {
 	max-height: 80vh;
 	overflow-y: auto;
 }
+
+/* GIF messages */
+._3zvs {
+	width: 100%; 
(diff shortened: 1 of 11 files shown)


sadmanf:
BEFORE:
image

AFTER:
image

whitecrownclown:
I'd use `max-width` instead. If you send an image you'll see the `Forward/Reply` actions on the left. Using `width` will automatically push these to the left-side of the chat (which means you can't really click them) instead of the immediate left of the image itself.

maxwidth
width

rachelyhe:
Made the quick change to `max-width: 100%`, should solve any alignment issues.

rachelyhe:
Sorry for the random updates. Was working on the wrong branch.

CvX:
Hmm, doesn't seem to work in all cases:

a gif in a conversation

The message element (`.3zvs`) has a reduced width but the actual image inside it exceeds its bounds.

rachelyhe:
It looks like the div containing the react button, "GIPHY", etc. has the class `_3zvs` which was resized in the changes above, but there is a _**second**_ div which has a custom-generated width and height. Within this hard value div is an img with `class="_mxz img"` and `max-width: 100%; width: 100%;`. 

Looks like this might be a much bigger change, selecting all img with `class="_mxz img"` and changing their parents to `width=100%; height=100%` so that it fits within the bounds of `_3zvs`.

sadmanf:
I looked through the codebase to see where an appropriate place would be for adding this change. I first thought to look through the message callbacks and perform this operation every time a gif was sent. But doing a check every time a message is sent seems to be a needlessly expensive operation. I don't think an interval-based check would be quite appropriate for this either.

Setting the parent div to `display: flex` correctly resizes the image, but there is sometimes unwanted white space below the image that I can't seem to get rid of (pictured below).
image


Any thoughts on how to proceed from here?

sindresorhus:
@whitecrownclown ⬆️

whitecrownclown:
That's because the container has values for both `width` and `height`.

This works alright for me:

```css
._3zvs {
	max-width: 100%;
}

/* Only match for elements that have style attributes `width` and `height` */
._41ud [style*="height"][style*="width"] {
	width: auto !important;
	height: auto !important;
}
```

before
after

sindresorhus:
Ping @sadmanf

PR #865, opened 2019-04-21
Add contact list styling for inactive window

### LINKED ISSUE #702: Inactive windows should have gray sidebar selection colors
Inactive windows lack their prominent highlight color to let the user navigate more easy between his/her apps and find the active one. The prominent blue highlight color in the sidebar should be changed to gray when the window is inactive. See the attached examples in the other ticket: #701

rachelyhe:
Fixes #702, when the window is inactive / blurred, the open chat changes to gray in the contact list. When the window is refocused, the open chat returns to blue in the contact list.

The selected state when the window is active (also the original "inactive" window state):
image
 
Inactive window for light mode:
light mode inactive

On dark mode:
dark mode inactive

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -573,13 +573,13 @@ window.addEventListener('load', () => {
 // Toggles the sidebar color to gray when window is inactive
 window.addEventListener('blur', () => {
 	const selected = document.querySelector('._1ht1._1ht2');
-	if(selected !== null) {
+	if (selected !== null) {
 		selected.classList.add('inactive');
 	}
 });
 window.addEventListener('focus', () => {
 	const selected = document.querySelector('._1ht1._1ht2');
-	if(selected !== null) {
+	if (selected !== null) {
 		selected.classList.remove('inactive');
 	}
 });
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -570,6 +570,16 @@ window.addEventListener('load', () => {
 	}
 });
 
+// Toggles the sidebar color to gray when window is inactive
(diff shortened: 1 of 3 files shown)


source/browser.ts
```diff
@@ -570,6 +570,20 @@ window.addEventListener('load', () => {
 	}
 });
 
+// Toggles the sidebar color to gray when window is inactive
+window.addEventListener('blur', () => {
+	const selected = document.querySelector('._1ht1._1ht2');
```
sindresorhus: Instead of targeting a specific selector, it would be better to add/remove a `is-window-inactive` class to `document.documentElement`. That way we can use it for any CSS selector.

rachelyhe: Updated this! The listener now adds / removes the `inactive` class to `document.documentElement`.

sindresorhus:
We should do this with all the blue buttons too, to closer imitate how native apps work.

rachelyhe:
If you check out Messages on Mac, most blue buttons / items aren't grayed out. Blue (sent) chat bubbles, header buttons, and unread notification buttons stay blue when the window is inactive. But it would be nice to gray out vibrancy, desaturate black text, etc.

CvX:
Keep in mind that with the new design (see #878):
* the "new message" button is already grayscale instead of blue
* buttons in the conversation view were already colored to match the "content" (own message bubbles), but with the new gradient options the notion that these buttons are part of the user content is even more pronounced (and could be argued that these shouldn't be changed when inactive, just like bubbles, avatars, and any shared media)

PR #872, opened 2019-04-25
Electron 5-related updates

CvX:
The only important one is the emoji tweak, fixes cropped native emojis.

### DIFF
--- a/source/menu.ts
+++ b/source/menu.ts
@@ -550,7 +550,8 @@ ${debugInfo()}`;
 			...switchItems
 		]),
 		{
-			label: 'File',
+			// @ts-ignore Buggy Electron types
+			role: 'fileMenu',
 			submenu: [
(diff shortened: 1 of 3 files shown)


source/menu.ts
```diff
@@ -550,7 +550,8 @@ ${debugInfo()}`;
 			...switchItems
 		]),
 		{
-			label: 'File',
+			// @ts-ignore Buggy Electron types
```
sindresorhus: If you have time, we should open an issue about this on https://github.com/electron/typescript-definitions

CvX: It looks like it has been already fixed some time ago (see: https://github.com/electron/electron/pull/16820/files#diff-27aca9003e7bfe3e2c635e67d04fc917L17) but just hasn't made it to the latest release?

I've added a comment in that PR.

PR #922, opened 2019-05-30
Fixes initial chat list dark background

pkuczynski:
I believe it's a bug: when I start Caprine my chat list has a dark gray background and when I hit CMD+R it reloads nicely to white one. This PR fixes this behavior and the list is white from the start.

Before:

Screenshot 2019-05-30 at 21 26 41

After:

Screenshot 2019-05-30 at 21 29 55

### DIFF
--- a/css/vibrancy.css
+++ b/css/vibrancy.css
@@ -1,12 +1,5 @@
 /* -- BLOCK START: sidebar-only vibrancy -- */
 
-html.sidebar-vibrancy body,
-html.sidebar-vibrancy ._4sp8,
-html.sidebar-vibrancy.dark-mode body,
-html.sidebar-vibrancy.dark-mode ._4sp8 {
-	background: transparent !important;
-}
-
 /* Login screen */
 html.sidebar-vibrancy ._3v_o {
 	background-color: #fff;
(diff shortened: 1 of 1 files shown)


CvX:
That's the desired effect ("vibrancy", the semi-transparent background, for the sidebar).
You can control that via "View -> Vibrancy" menu. The look you most likely want is the "No vibrancy".

And the white background after a refresh is actually a bug (tracked in #712)

pkuczynski:
Hmm, I don't really perceive this as semi-transparent. It looks grayish. It was also my default setting. Not something I turned on purpose. So I expect it would affect many people?

I also expect that vibrancy mode would rather make it more vibrant, and not less vibrant :) So I am a bit confused here?

CvX:
Well, it depends on what's behind the window. 🙂 Here's how it looks over a bright, colorful background:

caprine vibrancy screenshot caprine dark vibrancy screenshot

That's the default macOS look. And yeah, it's more grayish when the background is mostly monochrome.

pkuczynski:
Got it! So it supposed to be translucent? I would never guess that! :) Fact is, that I often have Webstorm open in a dark mode behind Caprine window :) Maybe if it would have more translucency it would be more clear what it does? :)

sindresorhus:
> So it supposed to be translucent?

Yes, to adhere to the macOS Human Interface Guidelines. Look at other apps like Messages, which has that too. You can also disable it in the "View" menu.

PR #923, opened 2019-05-31
Upgrade element-ready package

pkuczynski:
Required some changes to satisfy ts compiler. Tested locally and seems to work fine.


sindresorhus:
Please be careful about this and ensure none of the uses depend on that:

> It now stops looking when the DOM is ready by default. - https://github.com/sindresorhus/element-ready/releases/tag/v4.0.0

pkuczynski:
Well, I tested on local. Not sure what else shall I do?

sindresorhus:
> Well, I tested on local. Not sure what else shall I do?

I'm just making sure you tested all the places it's used. For example, the "Settings" menu seems to no longer show.

sindresorhus:
I just meant that there might be something that is loaded after DOM ready and might not work anymore.

sindresorhus:

pkuczynski:
Sure I can! Just had no chance to test it yet so not sure if this upgrade does not break anything. I am not aware of the changes in element-ready and reasoning behind them?

sindresorhus:
You need to convert the `import elementReady from 'element-ready';` statements to `import elementReady = require('element-ready');`

sindresorhus:
> I am not aware of the changes in element-ready and reasoning behind them?

The reasoning behind the changes are discussed in https://github.com/sindresorhus/element-ready/pull/17

pkuczynski:
I applied requested changes. Unfortunately the Settings or remove conversation dialog appear only after second time after launching the app (and every time afterwards). Not sure what could be the problem. Any advise?

source/browser-call.ts
```diff
@@ -1,6 +1,9 @@
-import elementReady from 'element-ready';
+import elementReady = require('element-ready');
 
 (async () => {
 	const startCallButton = await elementReady<HTMLElement>('._3quh._30yy._2t_');
-	startCallButton.click();
+
+	if (startCallButton) {
```
CvX: Those method calls should be correct without wrapping them with `if` statements. I’ll update element-ready definitions, then we can revert those changes here.

CvX: You can see my mental wrestling here: https://github.com/sindresorhus/element-ready/pull/23 😉
Bottom line: handling the undefined values may be unavoidable.

But in that case I think I’d rather have the `!` operator after the `elementReady()` (that we already use all over the place) rather than `if` statements.

source/browser.ts
```diff
@@ -43,7 +43,11 @@ async function withMenu(
 }
 
 async function withSettingsMenu(callback: () => Promise<void> | void): Promise<void> {
-	await withMenu(await elementReady<HTMLElement>('._30yy._6ymd._2agf,._30yy._2fug._p'), callback);
+	const settingsMenu = await elementReady<HTMLElement>('._30yy._6ymd._2agf,._30yy._2fug._p');
```
CvX: You’ll have to add `{stopOnDomReady: false}` here to fix the Settings/overlays issue.

Basically we’d like that option on all `elementReady()` calls…
I don’t like that `true` is the default. 😛

source/browser-call.ts
pkuczynski: Thing is that even having `!` at the end did not solved the problem and I was still seeing error in console log. That’s why I wrapped it with `if`.

source/browser.ts
pkuczynski: Yeah, I don’t like the default `true` either! I thought I added it everywhere, but maybe I am wrong. Will dbl check again in the afternoon...

source/browser-call.ts
CvX: A runtime error or a TypeScript error?

If runtime, that’s most likely related to missing `stopOnDomReady: false`.

If TS, you either have to wrap the whole await expression in parens, e.g. `let settingsMenu = (await elementReady<HTMLElement>('.test'))!;` or add the `!` where the value is used, e.g. `await withMenu(settingsMenu!, callback);`.

pkuczynski: The error was happing in runtime not in TS. For example I see

```
VM250 browser.js:43 Uncaught (in promise) TypeError: Cannot read property 'click' of null
    at selectMenuItem (VM250 browser.js:43)
    at VM250 browser.js:322
    at withMenu (VM250 browser.js:33)
    at withSettingsMenu (VM250 browser.js:38)
    at async openPreferences (VM250 browser.js:321)
    at async EventEmitter.<anonymous> (VM250 browser.js:128)
```

which is on the second line of compiled code:

```js
function selectMenuItem(itemNumber) {
    const selector = document.querySelector(`.uiLayer:not(.hidden_elem) ._54nq._2i-c._558b._2n_z li:nth-child(${itemNumber}) a`);
    selector.click();
}
```

despite the TS code looks as:

```js
function selectMenuItem(itemNumber: number): void {
	const selector = document.querySelector<HTMLElement>(
		`.uiLayer:not(.hidden_elem) ._54nq._2i-c._558b._2n_z li:nth-child(${itemNumber}) a`
	)!;
	selector.click();
}
```

So seems that `!` is not doing it's job I am afraid...

source/browser.ts
pkuczynski: You were right. It's now applied everywhere and seems to work fine!

source/browser-call.ts
pkuczynski: I was able to apply `!` in all `(await elementReady(...))!` but not for `document.querySelector`. Seems that `document.querySelector(...)!` does not do the job...

CvX: What’s the whole stacktrace of the error you get?

sindresorhus: @pkuczynski ⬆️

pkuczynski: Holidays, sorry :)

The issue disappeared and after removing the `if` I can not see any error in the console, so I reverted these changes.

sindresorhus:

pkuczynski:
Sure! Done. Is this ready to merge?

sindresorhus:
Yup. Thanks :)

PR #939, opened 2019-06-19
[Feature] Add `Hide Notification Content` menu option

### LINKED ISSUE #937: Ability to hide notification content
Hello,

firstly, great app - thanks for developing and maintaining. 👏 

I'm writing to ask, if `Caprine` actually have feature to hide content of message in desktop notification? Something like Slack way of not showing what someone sent to us, but only notification that this person send something (without message text). 🤔 

Best regards,
Marcin

whitecrownclown:
@sindresorhus 

Fixes #937 

Unsure about the name of the option. This is how it looks when it's enabled:
Screenshot 2019-06-20 at 01 52 13


sindresorhus:
How about calling it `Show Message Preview in Notifications`?

---

Can you document it in the readme with a before and after screenshot? (Take the screenshot with the built app, so it gets the Caprine icon, and take the screenshot with transparency using Cmd+Shift+4, press Space, and click the notification)

source/config.ts
```diff
@@ -26,7 +26,8 @@ const defaults = {
 	block: {
 		chatSeen: false,
 		typingIndicator: false,
-		deliveryReceipt: false
+		deliveryReceipt: false,
+		hideNotificationContent: false
```
sindresorhus: `notificationMessagePreview: true`?

sindresorhus: I don't think it should be under `block`.

sindresorhus:
@marcinlesek Any opinions?

whitecrownclown:
@sindresorhus Maybe we could add it as part of the config for https://github.com/sindresorhus/caprine/pull/787 ?

sindresorhus:
@whitecrownclown I think they are separate features. Private mode is much more "intrusive". I would enable this, but not private mode.

marcinlesek:
@sindresorhus `Show Message Preview in Notifications` and screenshots of notification before and after with transparent screenshot are great ideas. Totally agree 👍

PR #942, opened 2019-06-20
WIP: Reload mainWindow on system resume

### LINKED ISSUE #103: App stops updating on resume after suspend mode

whitecrownclown:
Fixes #103 

@sindresorhus 

Right now it only checks for connection on `resume` and reloads the mainWindow.
What other behaviour would we like? I checked and `Messenger` already keeps the current message & focus. Keeping it `WIP` for now.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -8,6 +8,7 @@ import {
 	screen as electronScreen,
 	session,
 	shell,
+	powerMonitor,
 	BrowserWindow,
 	Menu,
 	Notification,
@@ -331,6 +332,12 @@ function createMainWindow(): BrowserWindow {
 	enableHiresResources();
 
 	const {webContents} = mainWindow;
+
+	powerMonitor.on('resume', async () => {
+		await ensureOnline();
+
+		mainWindow.reload();
+	});
 
 	webContents.on('dom-ready', async () => {
 		await updateAppMenu();


sindresorhus:
Can you check if we need to add anything [here](https://github.com/sindresorhus/caprine/blob/ed720bdb931dc005567e71b125f695e14a9b4941/package.json#L139) for this?

sindresorhus:
This is not really a fix for #103. It's just an easy workaround. I would prefer to try to find a proper fix first. There must be a reason it stops updating. Maybe we haven't given enough permissions. Maybe we're blocking some connection. I think this needs more research.

reganlawton:
@sindresorhus its not a fix but I've been using the same workout for like 2 month might be nice to give non-dev built users a workaround till we can solve the root issue.

CvX:
This workaround gets a no-go from me.

I rarely have the issue where Messenger stop updating (maybe once a couple of months), but since I have full vibrancy enabled, this change would trigger the vibrancy bug every time I wake my machine up (and would have to restart the app).

reganlawton:
That’s not a workable solution then. I personally don’t use that feature so I didn’t notice the bug.

sindresorhus:
Like @CvX said, this workaround is not feasible until Electron fixes the vibrancy issue, and even then, I would prefer a proper fix.

sindresorhus:
Closing as I don't see us merging this.

Issue #904, opened 2019-05-23
Add copy image to context menu


pcktm:
I never really save images, I just copy them directly from my browser to some other software. Caprine sadly doesn't feature that option.

PR #943, opened 2019-06-20
WIP: Enable copy image context menu item

whitecrownclown:
Related to #904 

@sindresorhus 	Anything else we might add to the context menu?

Thanks!

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -39,7 +39,9 @@ electronDebug({
 });
 
 electronDl();
-electronContextMenu();
+electronContextMenu({
+	showCopyImageAddress: true
+});
 
 const domain = config.get('useWorkChat') ? 'facebook.com' : 'messenger.com';
 


sindresorhus:
https://github.com/sindresorhus/caprine/issues/904 is about copying the image itself, not its URL.

sindresorhus:
`Related to #904` => `Fixes #904` so the issue is auto-closed when the PR is merged.

whitecrownclown:
@sindresorhus Oh, my bad. I misunderstood. I'll update the PR.

sindresorhus:
Would be best to do a PR to Electron-context-menu to add it.

FreeOoI:
The important thing is to be able to copy (also with a keyboard shortcut command) from a tab an img/vids like the browser behaviour. Atm, the only solution is to save the files and manually transfer it or copy from your OS to paste it somewhere.

sindresorhus:
@whitecrownclown Can you finish the PR? Re: https://github.com/sindresorhus/electron-context-menu/releases/tag/v0.15.0

sindresorhus:
Actually, it's `true` by default, so I can just upgrade the dependencies.

PR #946, opened 2019-06-21
Add setting to hide the tray icon

### LINKED ISSUE #684: Optionally hide tray icon?

whitecrownclown:
Fixes #684 

@sindresorhus 

I've closed the previous PR, had some issue with those commits, #800 

To summarize:

- Added `Show Tray Icon` menu item, Defaults to `true`
- If `quitOnWindowClose` is `true` or `showTrayIcon` is `false`, don't create tray or `destroy()` already created tray

Thanks!

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -18,8 +18,8 @@ const defaults = {
 		y: undefined as number | undefined
 	},
 	menuBarMode: false,
-	hideDockIcon: false,
-	hideTrayIcon: false,
+	showDockIcon: true,
+	showTrayIcon: true,
 	alwaysOnTop: false,
 	bounceDockOnMessage: false,
 	showUnreadBadge: true,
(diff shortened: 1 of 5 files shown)


sindresorhus:
@CvX Thoughts on this?

source/config.ts
```diff
@@ -19,6 +19,7 @@ const defaults = {
 	},
 	menuBarMode: false,
 	hideDockIcon: false,
+	hideTrayIcon: false,
```
CvX: Can we straighten out both negations in the config? 😃

hideDockIcon -> showDockIcon
hideTrayIcon -> showTrayIcon

whitecrownclown: Done 👍

CvX: Alright, lgtm! 🙂

PR #956, opened 2019-06-24
Fix "Show Message Buttons" for new Messenger UI

### LINKED ISSUE #950: Show Message Buttons Not Working
When I click show message buttons, nothing changes.  Where as when my wife does it, removes the buttons.  Also for some reason the text bar is right in the middle of the buttons.

Caprine 2.32.1
Windows 7

whitecrownclown:
Fixes #950


sindresorhus:
Thanks for all these great fixes 🙌

PR #974, opened 2019-06-29
Fix Facebook logo placement

pkuczynski:
This PR supposed to fix the issue when FB needs to re-authenticate you due changed location. That happened to me today. Unfortunately I am not able to force FB to do it again, so can't test it after restoring the app.

Screenshot 2019-06-29 at 22 20 05


CvX:
Clearing site data in DevTools does the job:

PR #975, opened 2019-06-29
Fix new conversation icon on narrow window size

pkuczynski:
Before:

Screenshot 2019-06-29 at 23 34 02

After:

Screenshot 2019-06-29 at 23 58 04

### DIFF
--- a/css/browser.css
+++ b/css/browser.css
@@ -99,6 +99,23 @@ html.hide-dropdowns .uiContextualLayerPositioner {
 	height: 0 !important;
 }
 
+/* Narrow view */
+@media (max-width: 900px) {
+	/* New conversation and settings icons */
+	a._6-xq,
+	div._6ymu {
+		display: flex !important;
+		margin-top: 28px !important;
+		height: 26px !important;
+		width: 26px !important;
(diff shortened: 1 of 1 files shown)


pkuczynski:
I was thinking about it, but decided to go for smaller icons as otherwise the alignment of the left panel against header and conversation area is broken. I felt we should keep it as in original (full width) layout. Don’t you think?

And why would we remove the settings icon? It feels quite useful on the narrow view. I know messenger is hiding the icon but find it quite strange, don’t you think?

CvX:
Caprine hides the Settings button in both narrow and wide views (because we aim to provide all its functionality via native app menu and keyboard shortcuts).

I believe the fact it isn’t hidden in the new Messenger UI is an omission.

pkuczynski:
Changed to as per @CvX request.

pkuczynski:
By the way, current version of Caprine with new FB layout does not hide settings icon anymore. Should I fix it?

sindresorhus:
Yes

pkuczynski:
Here you go @sindresorhus #979

PR #976, opened 2019-06-30
Remove the white background from the thumbs-up icon

mikigal:
# Without fix:
Zrzut ekranu 2019-06-30 o 02 02 14

# With fix
Zrzut ekranu 2019-06-30 o 01 52 59


mikigal:
Zrzut ekranu 2019-06-30 o 01 53 08

Now I looked that with vibrancy there's dark (very similar to background) rectangle, I haven't idea how to fix it. Background as transparent broke emoji, but I think it's better than white background

CvX:
Could you share that SVG? 🙂

mikigal:
@CvX It's emoji HTML code: https://pastebin.com/qgtJWfzt
Rect must be non-transparent. When it's transparent there's purple (chat color) rectangle

CvX:
Uhh, what a backward way to do it… I know why they did that, but I absolutely hate it. 😬

Anyway, I’ve come up with a hack and added it to this PR. Basically I took the svg path of the new thumbs-up icon, encoded it in base64, and applied it as a `-webkit-mask`.

--- 

I’d need you to confirm if the `._7sli` is a good way to check for the new Messenger design. It should match the conversation pane (with other class on that element being `._1t2u`, in the old UI that’s the only class of that element).

mikigal:
Yep, conversation panel has classes . _1t2u and ._7sli
Your method works nice, good job :) 

Zrzut ekranu 2019-06-30 o 16 09 10

PR #986, opened 2019-07-04
Fix missing tray context menu on Linux

### LINKED ISSUE #965: 2.33.0 - Exit tray menu item is gone, only Caprine menu item available
Debian 9.9 MATE.

Caprine current version 2.33.0, previous version 2.27.

There has been some regressions lately regarding the tray icon. First the possibility to double click on the icon to open the program from the tray was removed.

Now with the upgrade from 2.27 to 2.33.0 is **the Exit menu item on the tray gone**. I only have Caprine in the menu. To quit the program I now need to open it and go to File, Quit.

whitecrownclown:
@sindresorhus 

Fixes #965

### DIFF
--- a/source/tray.ts
+++ b/source/tray.ts
@@ -80,6 +80,8 @@ export default {
 
 		tray = new Tray(getIconPath(false));
 
+		tray.setContextMenu(contextMenu);
+
 		updateToolTip(0);
 
 		const trayClickHandler = (): void => {


alekksander:
few builds back we (kde users)  had LMB to show/hide app, and RMB gave menu (toggle, quit). it only lasted for one or two builds. i hoped this would revert to this behavior, but it doesn't. currently both MB act alike. is it worth tracking back these working builds to fix what i mentioned?

PR #995, opened 2019-07-12
Fix hiding own profile picture in sidebar

pkuczynski:
Additionally, to the fix, I grouped all sidebar related styles into one "section", so it's easier to maintain them. @sindresorhus if you like this approach, I could group the other ones as well...


sindresorhus:
> if you like this approach, I could group the other ones as well...

👍

PR #998, opened 2019-07-15
Use JSON schema for the config

### LINKED ISSUE #870: Upgrade `electron-store` and use JSON Schema
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/caprine/issues/870)

TypeScript will help us at compile-time, but would be nice to have runtime validation too.

https://github.com/sindresorhus/electron-store#schema

The latest version of `electron-store` also ships with its own TS definition, so we can drop the `@types/electron-store` dependency then.

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/caprine/issues/870).

whitecrownclown:
Fixes #870


source/config.ts
```diff
@@ -1,50 +1,146 @@
 import Store = require('electron-store');
 import {is} from 'electron-util';
-import {EmojiStyle} from './emoji';
+import {JSONSchema} from 'json-schema-typed';
```
sindresorhus: I think we should re-export this type from `conf` and then `electron-store` instead.

sindresorhus:
Travis is failing

source/config.ts
whitecrownclown: You mean adding and exporting `json-schema-typed` to `conf` and then get it into `electron-store` through it?

sindresorhus: Yes, it's weird having to import an extra package just to get the type. `conf` and `electron-store` should provide the necessary types.

sindresorhus: https://github.com/sindresorhus/conf/issues/84

Issue #1006, opened 2019-07-20
Snap does not run from Wayland Session


zigford:
**Describe the bug**
Snap package immediately exits if run from within a wayland session. Additionally, if you have had a wayland session running, then switch to Xorg, the same problem exists as described [here](https://forum.snapcraft.io/t/electron-apps-and-wayland/12315)

**To Reproduce**
1. Install the snap
2. Ensure your logged into a Wayland session over Xorg
3. Attempt to run snap
4. Error is shown:

`Gtk-WARNING **: cannot open display: :0`

**Expected behavior**
The app opens

**Desktop (please complete the following information):**
 - OS: Gentoo linux
 - Installation package snap
 - Version [2.34.0]

**Additional context**

**Workaround**
A workaround is to set env variable `DISABLE_WAYLAND=1 snap run caprine`
This forces the electron app to fall back to XWayland

I will create a PR for the fix.

sindresorhus:
https://github.com/sindresorhus/caprine/pull/1007#issuecomment-514854648

PR #1007, opened 2019-07-20
Force XWayland otherwise electron apps fail

zigford:
Fix for #1006 
Manually built snap on my machine to test and confirm.


sindresorhus:
This will be fixed in the next update when I upgrade electron-builder, which I do before each update. https://github.com/electron-userland/electron-builder/commit/2278c2a

PR #1013, opened 2019-07-26
Fix keyboard shortcut for "Insert GIF"

aude:
Seems Messenger has been updated.

"Insert Emoji" and "Insert GIF" seem to have broken.

Here are fixes :raised_hands:

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -119,7 +119,15 @@ ipc.on('search', () => {
 });
 
 ipc.on('insert-gif', () => {
-	document.querySelector<HTMLElement>('._yht')!.click();
+	const gifElement =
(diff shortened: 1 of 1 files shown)


sindresorhus:
@aude Bump :)

aude:
Ah, right. Let's get it done.

aude:
I've updated the PR.

- I've removed the fix for "Insert Emoji", that was fixed in #1023
- I've updated the fix for "Insert GIF"
  - Include compatibility for old Messenger UI
  - Use style recommendations from @CvX in #1015 

Not super happy with how it looks, but it works. Feel free to change it if you want to.

PR #1014, opened 2019-07-26
Add `Attach files` menu item and keyboard shortcut

aude:
`Ctrl+U` seems to be used, could not make it work. Went for `Ctrl+T` instead.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -154,7 +154,9 @@ ipc.on('insert-sticker', () => {
 });
 
 ipc.on('insert-files', () => {
-	document.querySelector<HTMLElement>('._7oam input[type="file"]')!.click();
+	document
+		.querySelector<HTMLElement>('._5vn8 + input[type="file"], ._7oam input[type="file"]')!
+		.click();
 });
 
 ipc.on('focus-text-input', () => {
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -153,7 +153,7 @@ ipc.on('insert-sticker', () => {
 	stickerElement!.click();
 });
 
-ipc.on('insert-files', () => {
+ipc.on('attach-files', () => {
 	document
 		.querySelector<HTMLElement>('._5vn8 + input[type="file"], ._7oam input[type="file"]')!
 		.click();
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -175,11 +175,11 @@ ipc.on('delete-conversation', async () => {
 	await deleteSelectedConversation();
 });
 
-ipc.on('archive-conversation', async () => {
+ipc.on('hide-conversation', async () => {
 	const index = selectedConversationIndex();
 
 	if (index !== -1) {
-		await archiveSelectedConversation();
+		await hideSelectedConversation();
 
 		const key = index + 1;
 		await jumpToConversation(key);
@@ -261,7 +261,7 @@ ipc.on('show-message-requests-view', async () => {
 	await selectOtherListViews(4);
 });
 
-ipc.on('show-archived-threads-view', async () => {
+ipc.on('show-hidden-threads-view', async () => {
 	await selectOtherListViews(5);
 });
 
@@ -476,7 +476,7 @@ async function openMuteModal(): Promise<void> {
 	});
 }
 
-async function archiveSelectedConversation(): Promise<void> {
+async function hideSelectedConversation(): Promise<void> {
 	const groupConversationProfilePicture = document.querySelector<HTMLElement>(
 		`${selectedConversationSelector} ._55lu`
 	);
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -153,6 +153,10 @@ ipc.on('insert-sticker', () => {
 	stickerElement!.click();
 });
 
+ipc.on('insert-files', () => {
+	document.querySelector<HTMLElement>('._7oam input[type="file"]')!.click();
+});
+
 ipc.on('focus-text-input', () => {
 	document.querySelector<HTMLElement>('._7kpg ._5rpu')!.focus();
 });
--- a/source/menu.ts
+++ b/source/menu.ts
@@ -508,10 +508,10 @@ Press Command/Ctrl+R in Caprine to see your changes.
 			}
 		},
 		{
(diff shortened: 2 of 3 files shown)


aude:
Earlier discussion on the same subject: #357

source/browser.ts
```diff
@@ -123,6 +123,10 @@ ipc.on('insert-emoji', async () => {
 	emojiElement.click();
 });
 
+ipc.on('insert-files', () => {
+	document.querySelector<HTMLElement>('._7oam input[type="file"]')!.click();
```
CvX: `. _7oam` doesn't exist in the old messenger UI. There, the path would probably be:

```css
._5vn8 + input[type="file"]
```

source/menu.ts
```diff
@@ -486,6 +486,13 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				sendAction('insert-emoji');
 			}
 		},
+		{
+			label: 'Insert Files',
+			accelerator: 'CommandOrControl+T',
```
CvX: I'm not sure about `cmd+T`. That's the default "New Tab" shortcut in other apps.

Mail and Notes use `cmd+shift+A` for "Attach Files…", so maybe that would be a better choice? (at least on macOS)

CvX: Ah, `cmd+shift+A` is already taken by "Archive Conversation". 😕

Maybe it would be better to re-arrange some keyboard shortcuts then? Not sure…

source/browser.ts
aude: Oh, I wasn't aware that the old Messenger UI still exists.

Is there any way I can access the old UI? So I can test that as well.

CvX: Some people have reported that using a different browser makes the Messenger load the other UI version for them, but there's no reliable way to choose the UI version.

Anyway, `._5vn8 + input[type="file"]` does work in my (old UI) version (`.5vn8` is the "Add files" icon, and the input element directly follows it).

aude: Okay, that's fine, we can support both the old version and the new version.

What do you think about something like this?
```suggestion
	// old UI
	document.querySelector<HTMLElement>('._5vn8 + input[type="file"]')!.click();
	// new UI
	document.querySelector<HTMLElement>('._7oam input[type="file"]')!.click();
```

CvX: I'd go with a single query:

```suggestion
	document.querySelector<HTMLElement>('._5vn8 + input[type="file"], ._7oam input[type="file"]')!.click();
```

source/menu.ts
aude: Best would be to find out why `Ctrl+U` doesn't work IMO. That shortcut is muscle memory for, and probably many others.

Do you know if any shortcuts are impossible to override in Electron?

CvX: Hm, it seems that `CommandOrControl+U`, as well as `CommandOrControl+B` and `CommandOrControl+I`, are reserved for the rich text manipulation (underline, bold, and italic). I don't know if there's a way to avoid that.

If there is, instead it might be a good idea to rather add shortcuts for those actions, since Messenger supports bold/italic (see: https://www.facebook.com/help/147348452522644)

Anyway, why the `Ctrl+U` in the first place? You mentioned muscle memory, what apps use that shortcut?

As I noted, macOS Mail and Notes use `cmd+shift+A` ("Attach Files…"), and Messages uses `opt+cmd+F` ("Send file…").

aude: Slack uses `Ctrl+U`, that's where I have it from. Discord uses `Ctrl+Shift+U`. I think it makes sense because of the word **u**pload.

You mention several other apps that don't use `Ctrl+U`, so I'm open to alternatives.
 
I'm starting to like `Ctrl+Shift+A`. It's taken, like you mention. If we want to work on that, we could consider whether these hotkeys are needed. Could we remove hotkeys for these three?
- Mute Conversation `Ctrl+Shift+M`
- Archive Conversation `Ctrl+Shift+A`
- Delete Conversation `Ctrl+Shift+D`

CvX: We should keep existing keyboard shortcuts (they're quite useful, especially when doing conversations cleanup, e.g. archiving a bunch of conversations in a row).

Currently "Insert GIF" and "Insert Emoji" have shortcuts without the `shift` key modifier, so it would be preferable to have this new shortcut match that.

But… neither of those are available:

* `ctrl/cmd + F` ("Files")
* `ctrl/cmd + A` ("Attach"/"Add")
* `ctrl/cmd + U` ("Upload")

Maybe `ctrl/cmd + S` ("Send file")?

It's a stretch, but doesn't seem to collide with anything, and the usual meaning of "Save" doesn't appear in the context of Messenger.

---

(side note: "Insert Text" should probably renamed to "Focus message input" or similar, because it doesn't "insert" any text)

CvX: I've noticed that on iOS Messenger, there's no "Archive" action. It's "Hide Chat" now. So it wouldn't be unreasonable to rename that option in Caprine and to change the shortcut to `ctrl/cmd shift H` 🤔 

Then "Attach File" could take its spot.

sindresorhus: ⬆️ Let's do what @CvX suggests.

sindresorhus:
@aude Can you fix the merge conflict?

source/menu.ts
aude: I've implemented this in https://github.com/sindresorhus/caprine/pull/1014/commits/510f3ec44a5e9134ab60191404895e3b47b326b3 and https://github.com/sindresorhus/caprine/pull/1014/commits/11d1f7ca2cecdf184b480e82bf64b0b79a41a3be.

aude:
Yep.

Sorry for the delay, these PRs dragged out too long. Will make an effort to be more efficient in the future.

I think everything we've discussed is resolved now.

source/browser.ts
aude: This is done in https://github.com/sindresorhus/caprine/pull/1014/commits/b34aaf33d1157abb281595b6907bafc49e069e3f.

sindresorhus:

aude:
Yep, done.

PR #941, opened 2019-06-20
Fix emoji picker

whitecrownclown:
Fixes #831 

@sindresorhus 	@CvX 

Issue is pretty much explained [here](https://github.com/sindresorhus/caprine/issues/831#issuecomment-498070981) by @CvX 

Explanation: Triggering `mouseover` fetches the popover module. When switching conversations, the emoji component unmounts and needs `mouseover` in order to function correctly with the keyboard shortcut.`Mutation Observer` is clunky and heavy on performance since we can't listen directly on the node we require. Polling is also unreliable.

I think this one is consistent & the most performant. Please have a look.Thanks!

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#831: Bug with emoji selector](https://issuehunt.io/repos/42574339/issues/831)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

PR #1015, opened 2019-07-26
Add insert sticker shortcut

aude:


source/browser.ts
```diff
@@ -123,6 +123,14 @@ ipc.on('insert-emoji', async () => {
 	emojiElement.click();
 });
 
+ipc.on('insert-sticker', () => {
+	document.querySelectorAll<HTMLElement>('._7oam').forEach(node => {
```
CvX: Same as #1014: `.7oam` doesn't exist in the old Messenger UI. There the sticker button selector is `. _4rv6`.

Also, why the `forEach`? Is this code expected to `click()` on multiple elements?

aude: The code uses `forEach` because there are many `._7oam` elements, and I want the one that has a `svg path[d^="M22.5,18.5 L27.998,18.5"]` child.

So I want the `.7oam` element of the selector `.7oam svg path[d^="M22.5,18.5 L27.998,18.5"]`. If you know a pure CSS method, let's use that.

CvX: It might be better to use `find()` to better convey that you're looking for a single element:

```ts
[...document.querySelectorAll<HTMLELement>('.7oam')].find((element) =>
  element.querySelector<HTMLElement>('svg path[d^="M22.5,18.5 L27.998,18.5"]')
);
```

aude: Great idea, thanks.

I'll do this PR after #1014.

aude: Now I'm planning to do this after #1013, and then #1014 last.

sindresorhus:
> Now I'm planning to do this after #1013

It's merged.

source/browser.ts
aude: I have updated my PR.

Both old and new UI is now supported, same as #1013.

aude:
Thanks.

PR updated.

CvX:
On the old UI the first triggering of the shortcut results with just a flash of the sticker selector. Further shortcut invocations seem to work:

sticker selector flash

Unfortunately the #941 workaround for emoji doesn't work for this issue.
TBH I don't think it's a problem, I'd merge it in anyway.

aude:
Thanks for checking.

I remember that being a problem, it resembles the emoji problem for me.

As a side note, it makes a little weird trigger in the new UI as well. It might be because the code is clicking the `<a>`, but if you do it with the mouse you're actually clicking the SVG (or elements inside). I tried simulating clicking the SVG using [this method](https://gomakethings.com/how-to-simulate-a-click-event-with-javascript/), and it actually worked better.

If you want, @CvX, you could check whether that method works better in the old UI as well. But I'm open to merging this PR like it is.

PR #1023, opened 2019-08-05
Fix the emoji picker

whitecrownclown:
@sindresorhus 	@CvX 

Either I have access to a new messenger UI iteration or the selector changed. I added this additional selector for the emoji picker.

### DIFF
--- a/css/browser.css
+++ b/css/browser.css
@@ -347,7 +347,8 @@ body.show-message-buttons ._4_j4 ._7mkk {
 	}
 }
 
-._5s2p {
+._5s2p,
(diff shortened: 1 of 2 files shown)


CvX:
@aude Could you check if this change fixes what you were going for in #1013?

whitecrownclown:
@CvX The workaround doesn't seem to be necessary on the new UI. I think the picker is not lazy loaded.

CvX:
@whitecrownclown so we'll be able to remove the workaround in a year or two 😉

Issue #1010, opened 2019-07-23
Caprine window app won't move + Dark Mode issue


Andro:
When I try moving the window in macOS 10.14.5 and grab the area above the "Search Messenger" field, the app won't move. It only moves if I grab the area to the right of my interlocutor.

Also, Follow System Appearance is on, but the app won't change the dark mode when dark mode is selected...

---

Caprine 2.34.0
Electron 5.0.6
darwin 18.6.0
Locale: en-US

PR #1024, opened 2019-08-05
Fix draggable region

whitecrownclown:
@sindresorhus 	@CvX
 
Fixes draggable issue mentioned in #1010

### DIFF
--- a/css/browser.css
+++ b/css/browser.css
@@ -30,6 +30,7 @@ body {
 /* Bind the toolbar as the window's draggable region */
 ._36ic._5l-3,
 ._5742,
+._6-xk,
 ._673w {
 	-webkit-app-region: drag;
 }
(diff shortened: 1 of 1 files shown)


css/browser.css
```diff
@@ -297,7 +298,8 @@ html.is-window-inactive ._1ht1._1ht2 * {
 
 /* Message composer buttons */
 a._4rv6, /* "Choose a sticker" */
-a._4ce_ { /* "Play a game" */
+a._4ce_ {
+	/* "Play a game" */
```
CvX: Can you move this comment before the current block? Now it looks like it's describing either the `opacity` below it, or the whole block, instead of `a._4ce_` selector.

whitecrownclown: Yeah, that shouldn't have been part of the changes 😃, but I'll do it.

PR #1025, opened 2019-08-05
Fix conversation display on small window sizes

### LINKED ISSUE #1020: Opening the info sidebar results in super-wrapped text if window is thin
I like to use Caprine as a sidebar. It occupies maybe the left 20% of my screen, and whatever else I'm doing has the rest. It works fine for most applications. If I open someone's info (circled "i" button), however, it looks... funny.

image

It would be nice if this could be improved. By adding putting that part in a pop-up when the window is too thin, for instance.

whitecrownclown:
@CvX 

Fixes #1020

### DIFF
--- a/css/browser.css
+++ b/css/browser.css
@@ -416,3 +416,18 @@ html.private-mode ._1htf > span:first-child {
 	visibility: visible;
 }
 /* -- BLOCK END: Private Mode -- */
+
+/* -- BLOCK START: Small conversations window -- */
+
+@media all and (max-width: 500px) {
+	._20bp {
+		flex-direction: column !important;
+	}
+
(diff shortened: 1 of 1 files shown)


css/browser.css
```diff
@@ -416,3 +416,18 @@ html.private-mode ._1htf > span:first-child {
 	visibility: visible;
 }
 /* -- BLOCK END: Private Mode -- */
+
+/* -- BLOCK START: Small conversations window -- */
+
+@media all and (max-width: 500px) {
```
CvX: Hooray for Media Queries Level 4 😉

```suggestion
@media (max-width: 500px) {
```

PR #1101, opened 2019-10-08
Work around bug where the screen is black when closing window in full screen on macOS

nautatva:
Closing the app window when on full screen leaves a black screen. added logic to exit fullscreen before closing


source/index.ts
```diff
@@ -364,6 +364,18 @@ function createMainWindow(): BrowserWindow {
 			});
 			app.dock.setMenu(Menu.buildFromTemplate([firstItem, {type: 'separator'}, ...items]));
 		});
+		// Closing the app window when on full screen leaves a black screen
+		// Exit fullscreen before closing
+		electronLocalshortcut.register(mainWindow, 'Command+W', () => {
```
sindresorhus: You should listen to the window close event instead.

sindresorhus:
Can you open an issue on https://github.com/electron/electron? This is definitely an Electron bug.

nautatva:
The issue is already open at https://github.com/electron/electron/issues/20263. Since it's been almost a month and Electron-Version-7 also has this bug, i suspect this isn't the priority and hence the work-around.

sindresorhus:
I agree we should work around it. Can you resolve https://github.com/sindresorhus/caprine/pull/1101/files#r333316940 ?

nautatva:
Missed it, resolved in this commit

PR #1117, opened 2019-10-16
Add private mode confirmation prompt

### LINKED ISSUE #1112: No profil picture/Blur profil picture
Hi,

Since few days the profile pictures on each conversation are blurry, and the only way to see which conversation is with who is to select it and to look on the right panel.

This happened since the last Caprine's update.

Cheers,

alexandrepa:
Fixes #1112 
Feel free to change the dialog text, it's not my mother tongue 😄

### DIFF
--- a/source/util.ts
+++ b/source/util.ts
@@ -40,7 +40,7 @@ export function confirmPrivateModeDialog(): boolean {
 		dialog.showMessageBoxSync({
 			message: 'Are you sure you want to hide names and avatars ?',
 			detail: 'You have triggered the function using Ctrl+Shift+N.',
-			buttons: ['Yes', 'No'],
+			buttons: ['Hide', 'Don\'t Hide'],
 			defaultId: 0,
 			cancelId: 1
 		}) === 0
--- a/source/util.ts
+++ b/source/util.ts
@@ -38,7 +38,7 @@ export function showRestartDialog(message: string): void {
 export function confirmPrivateModeDialog(): boolean {
 	return (
 		dialog.showMessageBoxSync({
-			message: 'Are you sure you want to hide names and avatars ?',
+			message: 'Are you sure you want to hide names and avatars?',
 			detail: 'You have triggered the function using Ctrl+Shift+N.',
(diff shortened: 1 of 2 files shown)


source/util.ts
```diff
@@ -34,3 +34,15 @@ export function showRestartDialog(message: string): void {
...
+	return (
+		dialog.showMessageBoxSync({
+			message: 'Are you sure you want to hide names and avatars ?',
+			detail: 'You have triggered the function using Ctrl+Shift+N.',
+			buttons: ['Yes', 'No'],
+			defaultId: 0,
+			cancelId: 1
+		}) === 0
+	);
+}
```
sindresorhus: This should not be in `util.js`. This file is for reusable things.

source/util.ts
```diff
@@ -34,3 +34,15 @@ export function showRestartDialog(message: string): void {
 		app.quit();
 	}
 }
+
+export function confirmPrivateModeDialog(): boolean {
+	return (
+		dialog.showMessageBoxSync({
+			message: 'Are you sure you want to hide names and avatars ?',
+			detail: 'You have triggered the function using Ctrl+Shift+N.',
+			buttons: ['Yes', 'No'],
```
```suggestion
			buttons: ['Hide', 'Don\'t Hide'],
```

source/util.ts
```diff
@@ -34,3 +34,15 @@ export function showRestartDialog(message: string): void {
 		app.quit();
 	}
 }
+
+export function confirmPrivateModeDialog(): boolean {
+	return (
+		dialog.showMessageBoxSync({
+			message: 'Are you sure you want to hide names and avatars ?',
```
```suggestion
			message: 'Are you sure you want to hide names and avatars?',
```

source/util.ts
```diff
@@ -34,3 +34,15 @@ export function showRestartDialog(message: string): void {
...
+			detail: 'You have triggered the function using Ctrl+Shift+N.',
```
sindresorhus: This is not correct. It could have been Command instead of Control.

alexandrepa:
I think there is a conflict between eslint and prettier with single quote.

PR #1118, opened 2019-10-16
Fix "Insert Text" keyboard shortcut

aude:
The selector for the "Insert Text" keyboard is not specific enough, it focuses the wrong textbox some times (for example when having the "search" feature open in a group chat).

* I've made the selector more specific.
* I also renamed "Insert Text" to "Focus Textbox" https://github.com/sindresorhus/caprine/pull/1014#discussion_r310539228
  * I chose "Focus Textbox" over "Focus Text Input" for brevity. Can rename if wanted.


sindresorhus:
I would prefer `Focus Text Input`.

Issue #1122, opened 2019-10-17
Showing menu bar after auto hide in Windows is not discoverable


hiteda:
I had some difficulty learning how to restore the menu bar after turning on "Auto Hide Menu Bar". I was about to file an issue when I found a few preexisting issues (#115, #208) addressing the same thing. The solution is to press the Alt key.

While this is consistent with other Windows applications, it doesn't seem to be something a lot of users would try right away.

I would suggest to find a way to make it more discoverable to users. Even adding a shortcut key (e.g. Ctrl + Shift + H) so the users could see in the menu how to toggle it would greatly improve the user experience.

Any other ideas?

sindresorhus:
The Alt shortcut is documented in the readme: https://github.com/sindresorhus/caprine#keyboard-shortcuts We could potentially show a prompt about it when you enable "Auto Hide Menu Bar", but honestly, it's expected knowledge when you use Windows.

hiteda:
I think the prompt is a great idea. We can also allow them to dismiss the message so it doesn't appear every time. While using the Alt key is a standard feature of Windows, I wouldn't assume that most users really know about it or use it often. Discoverability makes a big difference in user experience.

PR #1123, opened 2019-10-17
Added Ctrl+Shift+H shortcut for auto hide menu bar to improve user...

hiteda:
…experience
Issue #1122


sindresorhus:
This will toggle the option to auto-hide the menu bar, not actually toggle the menu bar, which is just done as a side-effect of toggling option. I think this is more confusing than helpful, honestly.

PR #1125, opened 2019-10-18
Show tip after enabling `Auto Hide Menu Bar`

### LINKED ISSUE #1122: Showing menu bar after auto hide in Windows is not discoverable

hiteda:
Added an info alert when the user toggles on Auto Hide Menu Bar.
Fixes #1122

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -33,6 +33,7 @@ type StoreType = {
 	useWorkChat: boolean;
 	sidebarHidden: boolean;
 	autoHideMenuBar: boolean;
+	showAutoHideAlert: boolean;
 	notificationsMuted: boolean;
 	hardwareAcceleration: boolean;
 	quitOnWindowClose: boolean;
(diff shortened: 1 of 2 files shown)


sindresorhus:
I don't think we need the checkbox. The "Auto Hide Menu Bar" is a setting you set once and then use Alt to toggle it. It's not like you would constantly change it.

hiteda:
Sounds good. I'll remove the checkbox.

PR #1138, opened 2019-10-24
Make the Touch Bar respect private mode

jyoo980:
👋Hi!

First-time contributor to a project on GitHub (or any open source, really). Thanks for all the work you've done on Caprine. I noticed that when I selected the `Hide Names and Avatars` option in the menu bar, it worked great for the app itself, but the touch bar still showed information. 

This is a PR which would add behaviour to make the touch bar look like this instead when `Hide Names and Avatars` is selected

image

Thanks for taking a look!


source/touch-bar.ts
```diff
@@ -15,9 +22,16 @@ ipc.on('conversations', (_event: ElectronEvent, conversations: Conversation[]) =
...
 
-	const touchBar = new TouchBar({items});
-	const win = getWindow();
-
-	win.setTouchBar(touchBar);
+ipc.on('hide-touchbar-labels', (_event: ElectronEvent) => {
+	const privateModeLabel = new TouchBarButton({
+		label: 'Private mode enabled',
+		backgroundColor: undefined,
+		icon: nativeImage.createFromPath(path.join(__dirname, '..', 'static', 'Icon.png')),
```
sindresorhus: I noticed we use the icon path in multiple places. Can you move the icon path to a `constants.js` file?

```
source/menu.ts
573:				icon: path.join(__dirname, '..', 'static', 'Icon.png'),

source/touch-bar.ts
32:		icon: nativeImage.createFromPath(path.join(__dirname, '..', 'static', 'Icon.png')),

source/index.ts
261:		icon: is.linux ? path.join(__dirname, '..', 'static', 'Icon.png') : undefined,
```

sindresorhus:
Thanks :)

PR #1139, opened 2019-10-24
Fix "Attach Files" keyboard shortcut

aude:
Was happy to see v2.40.0 released, with the new "Attach Files" keyboard shortcut :smile: Good work :beers: 

Was eager to try "Attach Files", and it works, though not well.

`CommandOrControl+Shift+A` requires two presses before activating at my system. I'm guessing that the key combination conflicts with something.

I'm proposing to change it to `CommandOrControl+T` to make it work in one key press. Any other combination is fine as well, as long as it works.


aude:
It would be good to know what `CommandOrControl+U` and `CommandOrControl+Shift+A` are conflicting with. Is it core Electron? Or is it the Messenger website?

[This](https://github.com/electron/electron/issues/1338) suggests it's possible to override even core Electron keyboard shortcuts, but maybe [electron-localshortcut](https://www.npmjs.com/package/electron-localshortcut) is not able to?

sindresorhus:
You need to update the readme too.

aude:
Thanks. Updated.

PR #1150, opened 2019-10-28
Fix `Always on Top` after tray toggling on KDE Plasma

MHekert:
It fixes bug on linux with KDE Plasma that after toggling to and from tray 'Always on Top' setting stops working despite it is still checked.


sindresorhus:
Can you report this to https://github.com/electron/electron and add a comment above the fix with a link to the issue?

alekksander:
this is still present in freshly build caprine

Issue #1091, opened 2019-09-28
Does not switch between dark/light modes without restart


stackptr:
It's unclear to me if this is as intended (i.e., not implemented) or if I am experiencing a bug.

Reading this [guide](https://electronjs.org/docs/tutorial/mojave-dark-mode-guide) on implementing dark mode in Electron, it seems like the check on the current "mode" is via `api.systemPreferences.isDarkMode` which I see in `source/browser.ts: setDarkMode()`. Yet that guide also mentions the `AppleInterfaceThemeChangedNotification` which I cannot grep for.

I can work on adding this if it is indeed missing but I wanted to check here first

Version: `2.38.0 (2.38.0.1617)`

stackptr:
The issue I'm experiencing around the "Follow system appearance" option is referenced but not elaborated on in these other issues: #1010 #1046

iandrewt:
Supposedly this requires electron 7.0.0 to fix properly (at least in Catalina), or an Info.plist file with `NSRequiresAquaSystemAppearance`.

Electron 7 isn't out yet (it's in beta), so we might need to wait for that before this can be fixed. According to [Electron's release timeline](https://electronjs.org/docs/tutorial/electron-timelines), that should be October 22nd.

Issue #1153, opened 2019-10-30
openExternalSync is deprecated (and removed in Electron 7)


AidanGG:
https://github.com/sindresorhus/caprine/blob/d27f5af0be14baeb02a602736e8ada44d54a8e07/source/index.ts#L456-L460
and
https://github.com/sindresorhus/caprine/blob/d27f5af0be14baeb02a602736e8ada44d54a8e07/source/index.ts#L500-L503

This function is not available on Electron 7 (which will presumably be supported in the future, and is the only way to use Caprine on Arch Linux right now). I'm not asking to switch to Electron 7, but could we perhaps switch to using the asynchronous `shell.openExternal` instead, which returns `Promise<void>` since the Boolean result of `openExternalSync` is unused? This should restore link opening on Arch Linux.

Saboteur777:
:+1:

sindresorhus:
Should be fixed by https://github.com/sindresorhus/caprine/releases/tag/v2.41.0

PR #1157, opened 2019-10-31
Update to Electron 7

stackptr:
I was complaining about a system vibrancy problem in #1091 and it was indicated that to fix the issue properly on Catalina it was necessary to wait for the next version of Electron to be released. It was indeed released a week ago.

This PR updates the version and then fixes the errors that I saw running `npm install && npm start`.

(It does not appear to fix my original complaint, any additional work will go in another branch)

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
 		"@typescript-eslint/eslint-plugin": "^2.3.3",
 		"@typescript-eslint/parser": "^2.3.3",
 		"del-cli": "^3.0.0",
-		"electron": "^6.0.11",
+		"electron": "^7.0.0",
 		"electron-builder": "^21.2.0",
 		"eslint-config-xo-typescript": "^0.19.0",
 		"husky": "^3.0.8",
--- a/source/emoji.ts
+++ b/source/emoji.ts
@@ -12,16 +12,6 @@ import {memoize} from 'lodash';
 import config from './config';
 import {showRestartDialog, getWindow, sendBackgroundAction} from './util';
 
-// TODO: removeListener is typed incorrectly in v7.0.0. Remove on later version
-// Issue: https://github.com/electron/electron/issues/20705
-// PR: https://github.com/electron/electron/pull/20712
-declare module 'electron' {
-	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	class IpcMain {
-		removeListener(event: string, listener: (...args: any[]) => void): this;
-	}
-}
-
 // The list of emojis that aren't supported by older emoji (facebook-2-2, messenger-1-0)
 // Based on https://emojipedia.org/facebook/3.0/new/
 const excludedEmoji = new Set([
--- a/source/emoji.ts
+++ b/source/emoji.ts
@@ -16,6 +16,7 @@ import {showRestartDialog, getWindow, sendBackgroundAction} from './util';
 // Issue: https://github.com/electron/electron/issues/20705
 // PR: https://github.com/electron/electron/pull/20712
 declare module 'electron' {
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	class IpcMain {
 		removeListener(event: string, listener: (...args: any[]) => void): this;
 	}
--- a/source/emoji.ts
+++ b/source/emoji.ts
@@ -16,11 +16,10 @@ import {showRestartDialog, getWindow, sendBackgroundAction} from './util';
 // Issue: https://github.com/electron/electron/issues/20705
 // PR: https://github.com/electron/electron/pull/20712
 declare module 'electron' {
-  class IpcMain {
-    removeListener(event: string, listener: (...args: any[]) => void): this;
-  }
+	class IpcMain {
+		removeListener(event: string, listener: (...args: any[]) => void): this;
+	}
 }
-
 
 // The list of emojis that aren't supported by older emoji (facebook-2-2, messenger-1-0)
 // Based on https://emojipedia.org/facebook/3.0/new/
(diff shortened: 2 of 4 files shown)


stackptr:
There are some non-blocking warnings that I will also try to address:

```
> tsc && electron .

(electron) 'getName function' is deprecated and will be removed. Please use 'name property' instead.
(electron) 'systemPreferences.isDarkMode()' is deprecated and will be removed. Please use 'nativeTheme.shouldUseDarkColors' instead.
(electron) 'setAppLevelAppearance function' is deprecated and will be removed. Please use 'appLevelAppearance property' instead.
(electron) 'setBadgeCount function' is deprecated and will be removed. Please use 'badgeCount property' instead.
```

AidanGG:
This fixes my issue that I opened up yesterday too https://github.com/sindresorhus/caprine/issues/1153 :smile:.

stackptr:
It's unclear to me how to resolve the lint error on `@typescript-eslint/no-unused-vars` for the `IpcMain` class, but since it's a hack that seems to be resolved in electron/electron#20712 maybe the thing to do is to wait for the minor version that fixes it? I'm not quite sure in what version that would be expected to land.

sindresorhus:
Just use a `// @ts-ignore` comment to work around it.

stackptr:
Dark mode following system appearance doesn't have to go in this PR, but pretty much all the existing logic for it seems to have been deprecated. I've been trying to get this to use the new `nativeTheme` api and I've finally determined that it is not importable in `source/browser.ts` where `setDarkMode()` lives. 

The import statement `import {nativeTheme} from 'electron'` does not yield type errors it just turns out undefined when trying to use it. After some digging, I see that the different "processes" have a module listing in the main electron repo, `lib/${PROCESS}/api/module_list.ts`. So, for `browser.ts` the process is `renderer`, which lacks any mention of the relevant `nativeTheme` api. It is however accessible in `source/index.ts`, which could be the `browser` process. (Which seems to be referred to as Main in the electron docs).

Anyways, it appears some code needs to be shuffled around here. I can send `nativeTheme` in the `set-dark-mode` event in the `webContents` `EventEmitter` but then there is the `setDarkMode` call in the `DOMContentLoaded` event that would also need it. I'll poke around further but any ideas regarding this architecture would be helpful!

sindresorhus:
You could use https://github.com/sindresorhus/electron-util#api-1

source/index.ts
```diff
@@ -156,7 +154,18 @@ ipcMain.on('update-tray-icon', updateTrayIcon);
...
+interface OnSendHeadersDetails {
+	id: number;
+	url: string;
+	method: string;
+	webContentsId?: number;
+	resourceType: string;
+	referrer: string;
+	timestamp: number;
+	requestHeaders: Record<string, string>;
 }
```
sindresorhus: Why do we have to manually type this now?

source/emoji.ts
```diff
@@ -12,6 +12,16 @@ import {memoize} from 'lodash';
...
 
+// TODO: removeListener is typed incorrectly in v7.0.0. Remove on later version
+// Issue: https://github.com/electron/electron/issues/20705
+// PR: https://github.com/electron/electron/pull/20712
+declare module 'electron' {
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	class IpcMain {
+		removeListener(event: string, listener: (...args: any[]) => void): this;
+	}
+}
```
sindresorhus: For reference, this is always why I wait a while after a major Electron release before upgrading. There are always silly bugs like this.

source/index.ts
```diff
@@ -444,28 +453,31 @@ function createMainWindow(): BrowserWindow {
...
 		event.preventDefault();
 
 		if (url === 'about:blank') {
 			if (frameName !== 'about:blank') {
 				// Voice/video call popup
 				options.show = true;
 				options.titleBarStyle = 'default';
-				options.webPreferences.nodeIntegration = false;
-				options.webPreferences.preload = path.join(__dirname, 'browser-call.js');
+				if (options.webPreferences !== undefined) {
```
sindresorhus: Why would it be `undefined`?

And even if it is, the correct way would be to do `options.webPreferences = options.webPreferences || {}`.

stackptr: It's marked as optional in the docs: https://electronjs.org/docs/api/browser-window#new-browserwindowoptions

and was throwing a type error without this change. But I think I misunderstood what this code was doing, it appears as though an `options` object is being built so I'll revert and see a better way of doing this

source/index.ts
stackptr: It doesn't seem to be present in the `electron` library. I'm not sure why

package.json
```diff
@@ -48,7 +48,7 @@
 		"@typescript-eslint/eslint-plugin": "^2.3.3",
 		"@typescript-eslint/parser": "^2.3.3",
 		"del-cli": "^3.0.0",
-		"electron": "^6.0.11",
+		"electron": "^7.0.0",
```
sindresorhus:
```suggestion
		"electron": "^7.1.1",
```

Is the latest now.

source/emoji.ts
```diff
@@ -12,6 +12,16 @@ import {memoize} from 'lodash';
 import config from './config';
 import {showRestartDialog, getWindow, sendBackgroundAction} from './util';
 
+// TODO: removeListener is typed incorrectly in v7.0.0. Remove on later version
```
sindresorhus: I think this is fixed in the latest Electron version.

stackptr:
Addressed merge conflicts, resolved some questions and using later Electron version to avoid that type hack.

sindresorhus:
Thanks 🙌

PR #1158, opened 2019-11-01
Restore maximized state of the window

tymmesyde:
Restore the maximized state of the window at launch if it was before exiting

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -324,15 +324,15 @@ function createMainWindow(): BrowserWindow {
 	});
 
 	win.on('resize', () => {
-		config.set('lastWindowState', win.getNormalBounds());
+		config.set('lastWindowState', win.getNormalBounds() as any);
 	});
 
 	win.on('maximize', () => {
-		config.set('maximizedState', true);
+		config.set('lastWindowState.isMaximized' as any, true);
 	});
 
 	win.on('unmaximize', () => {
-		config.set('maximizedState', false);
+		config.set('lastWindowState.isMaximized' as any, false);
 	});
 
 	return win;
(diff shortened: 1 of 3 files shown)


sindresorhus:
Is it normal on Windows and Linux to restore maximized state?

source/index.ts
```diff
@@ -327,6 +327,14 @@ function createMainWindow(): BrowserWindow {
 		config.set('lastWindowState', win.getNormalBounds());
 	});
 
+	win.on('maximize', () => {
+		config.set('maximizedState', true);
```
```suggestion
		config.set('lastWindowState.isMaximized', true);
```

source/tray.ts
```diff
@@ -19,7 +19,11 @@ export default {
 			if (win.isVisible()) {
 				win.hide();
 			} else {
-				win.show();
+				if (config.get('maximizedState')) {
+					win.maximize();
```
sindresorhus: Are you sure `.maximize()` totally replaces `.show()`?

source/index.ts
sindresorhus: And add a default value of false here: https://github.com/sindresorhus/caprine/blob/30d83cc3b22c279e05501ef0199d3e188885d372/source/config.ts#L86

And add it here: https://github.com/sindresorhus/caprine/blob/30d83cc3b22c279e05501ef0199d3e188885d372/source/config.ts#L14 and

tymmesyde: > Is it normal on Windows and Linux to restore maximized state?

Yes, I think so.
On Windows, Spotify and other apps are doing it.

source/tray.ts
tymmesyde: So I checked the doc and `maximize()` show the window but not focusing.

sindresorhus:
I'm getting an error when Command+Q quitting the app:

Screenshot 2019-11-11 at 23 29 40

sindresorhus:
Bump

sindresorhus:
Bump

sindresorhus:
Closing in favor of https://github.com/sindresorhus/caprine/issues/1268.

PR #1169, opened 2019-11-07
Remove Facebook tracking from copied links

### LINKED ISSUE #248: Block Facebook link tracking when "Copy link" chosen
PR #226 fixed #224 when clicking on a link. If you right-click a link and choose "Copy link" the Facebook-wrapped tracking link is still copied, though.

### LINKED ISSUE #286: Rewrite urls to remove Facebook's redirect
For example instead of "copy link" showing `https://google.com` it shows this link.

```
https://l.messenger.com/l.php?u=https%3A%2F%2Fgoogle.com%2F&h=ATM85iuo8Kuz-3WhZMFqASMviGKKm6On0v8CdBSD2rO9Vaf6IbC78KK9E_erXrWK_oTnOoxA5FUPDVuAHtuNrWQevJHoKJp7RUOnDBGJk75wuuGaP5395YBvxBMOUwKIG6QieMxBETPOsahb-NI
```

jaflo:
This fixes #248 and fixes #286. The actual code to clean the URL is copied from PR #226 (same functionality for clicked links). As mentioned in the linked issues, this makes use of code changes in sindresorhus/electron-context-menu#44.

Ideally the menu item would be replaced in a neater way, but that requires upstream changes. For now, this code change takes advantage of the fact that the default menu items can be exposed through the `prepend` function and then the `transform` functionality is used to change the URL as needed.

I considered two ways to change the functionality: use of the `menu` option or some new functionality in the library that replaces default menu items. The first approach cannot be done because we cannot read the default list of menu items, so we would need to duplicate code. Even if that is done, the type definition for `Actions` has no `copyLink`. The second approach is cleaner: some new functionality that allows for replacement of default items. I think that this should be possible once sindresorhus/electron-context-menu#70 is addressed.

A summary of these issues has been put in place while this hacky solution is used.


source/index.ts
```diff
@@ -43,7 +43,27 @@ electronDebug({
...
 
 electronDl();
-electronContextMenu();
+electronContextMenu({
+	prepend: defaultActions => {
+		/*
+		 * TODO: Use menu option or use replacement of options (https://github.com/sindresorhus/electron-context-menu/issues/70)
+		 * This is a bit hacky, but we cannot import menuTemplate directly to populate the menu.
+		 * If we use prepend and return a MenuItem, we will get a duplicate menu item.
+		 * The cast below is because the type definition of Actions has no copyLink.
```
sindresorhus: Fixed: https://github.com/sindresorhus/electron-context-menu/releases/tag/v0.15.1

source/index.ts
```diff
@@ -43,7 +43,27 @@ electronDebug({
...
+		 */
+		(defaultActions as any).copyLink({
+			transform: (content: any) => {
+				if (content.startsWith('https://l.messenger.com/l.php')) {
+					content = new URL(content).searchParams.get('u');
+				}
```
sindresorhus: Instead of copying the logic, move it to `util.ts` and use it in both places.

jaflo: Thanks! I've addressed all comments in ceed63d and toggled the branch from and to master to get an updated diff including the new changes merged in from upstream.

source/util.ts
```diff
@@ -37,3 +38,14 @@ export function showRestartDialog(message: string): void {
 		app.quit();
 	}
 }
+
+export const domain = config.get('useWorkChat') ? 'facebook.com' : 'messenger.com';
```
```suggestion
export const messengerDomain = config.get('useWorkChat') ? 'facebook.com' : 'messenger.com';
```

sindresorhus: Is this export actually used anywhere?

source/util.ts
```diff
@@ -37,3 +38,14 @@ export function showRestartDialog(message: string): void {
...
+
+export function getUntrackedURL(url: string): string {
```
```suggestion
export function stripTrackingFromUrl(url: string): string {
```

source/util.ts
jaflo: It is in index.ts: https://github.com/sindresorhus/caprine/blob/ceed63debda47b039d800e92d08fe84d8da4d88b/source/index.ts#L209-L214

I'll make the changes to rename the variable though.

PR #1170, opened 2019-11-07
Add ability to force sidebar width

### LINKED ISSUE #297: Compact contact sidebar even on larger sizes
I honestly would enjoy the option to have the contact sidebar be compact even when the window is at a larger size, instead of compact as well. This way the window doesn't have to be extra big when you want a normal line length. This would save a little bit less than 20% of the total size. :)

jaflo:
Adds the ability to change between different sidebar widths (fixes #297). Let's you choose between:

* adaptive (default, changes with window size)
* hidden (same as before, hides sidebar completely, fixed to move group text title out of the way of MacOS traffic lights)
* narrow (force narrow-screen sidebar on all window sizes)
* wide (force wide-screen sidebar on all window sizes)

image

I wasn't sure on how to keep the toggle sidebar keyboard shortcut, so I made the menu item to hide the sidebar act as toggle. This might be unconventional, let me know if there is a better way.

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -30,7 +30,7 @@ type StoreType = {
 	};
 	emojiStyle: 'native' | 'facebook-3-0' | 'messenger-1-0' | 'facebook-2-2';
 	useWorkChat: boolean;
-	sidebar: 'default' | 'hidden' | 'slim' | 'wide';
+	sidebar: 'default' | 'hidden' | 'narrow' | 'wide';
 	autoHideMenuBar: boolean;
 	notificationsMuted: boolean;
 	hardwareAcceleration: boolean;
@@ -155,7 +155,7 @@ const schema: {[Key in keyof StoreType]: Store.Schema} = {
 	},
 	sidebar: {
 		type: 'string',
-		enum: ['default', 'hidden', 'slim', 'wide'],
+		enum: ['default', 'hidden', 'narrow', 'wide'],
 		default: 'default'
 	},
 	autoHideMenuBar: {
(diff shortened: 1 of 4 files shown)


source/menu.ts
```diff
@@ -86,6 +86,51 @@ export default async function updateMenu(): Promise<Menu> {
...
+			accelerator: 'CommandOrControl+Shift+S',
+			async click() {
+				// Toggle between default and hidden
+				config.set('sidebar', config.get('sidebar') === 'hidden' ? 'default' : 'hidden');
+				sendAction('update-sidebar');
+				await updateMenu();
+			}
+		},
+		{
+			label: 'Slim Sidebar',
```
sindresorhus: `Narrow` would probably be a better word for it?

source/config.ts
```diff
@@ -153,9 +153,10 @@ const schema: {[Key in keyof StoreType]: Store.Schema} = {
 		type: 'boolean',
 		default: false
 	},
-	sidebarHidden: {
-		type: 'boolean',
```
sindresorhus: You need to migrate the old `boolean` version. See the bottom of this file.

source/menu.ts
jaflo: That does sound better! Fixed

source/config.ts
jaflo: Good catch, I didn't know of this. Added!

jaflo:
@sindresorhus @CvX is there any update on this?

Issue #1198, opened 2019-11-22
Can't display 360 degree photos


huylv:
Latest version 2.41.1 can't display 360 photos. The loading icon shows up for a few seconds then disappears, breaks chat layout afterwards.

PR #1199, opened 2019-11-26
Fix hidden 360 photos

panda7789:
After #1167 360 photos don't have minimal width and height.
So I removed 360 class from being set width|height = auto !important.


panda7789:
Resolves #1198 forget to mention.

PR #1202, opened 2019-11-27
Fix “Hide conversation” menu item

### LINKED ISSUE #1027: Archive does nothing right now
It just doesn't do anything on my win10 pc. EDIT: At least from the conversation menu. Haven't tried other ways as I haven't got any new messages since then. 😅 

---

Caprine 2.34.0
Electron 5.0.6
win32 10.0.18945
Locale: en-US

yatharth:
Fixes #1027 

Right now, the “Hide conversation” menu item (AKA the archive button) does not work. This was because of some CSS selectors seem to have become outdated. I have replaced them as follows:

1. `.uiLayer:not(.hidden_elem) ._54nq._2i-c._558b._2n_z` → `.uiLayer:not(.hidden_elem) [role=menu]`
1. `${selectedConversationSelector} ._5blh._4-0h` → `${selectedConversationSelector} [aria-haspopup=true] [role=button]`

Note that the new selectors do not rely on minified class names like `_5blh` but rather on descriptive attributes like `[role=button]`. Hopefully, this means that these new selectors will be less likely to break with changes made to the Messenger website by Facebook.

Note also that I have tried to avoid using any localized attributes, so hopefully this works for all messenger.com localizations.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -508,7 +508,7 @@ In other words, you should only use this function within a callback
  makes sure to have the conversation menu open before executing the
  callback and closes the conversation menu afterward.
 */
-function isSelectedConversationGroup(): Boolean {
+function isSelectedConversationGroup(): boolean {
 	const separator = document.querySelector<HTMLElement>(
 		`${conversationMenuSelector} > li:nth-child(6)[role=separator]`
 	);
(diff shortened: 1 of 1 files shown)


sindresorhus:
👍

PR #1215, opened 2019-12-12
Fix the "Show Tray Icon" preference

JBudny:
System: Windows 7
Fixed problem: File->Caprine Settings->Show Tray Icon option wasn't working.


sindresorhus:
Can you move the toggle logic into a function and also apply it here: https://github.com/sindresorhus/caprine/blob/a6c914ebde7390afb230de5f9daa8b55635fd3d4/source/menu.ts#L348 ?

sindresorhus:
Also run `$ npm test` locally and ensure the linting passes.

source/config.ts
```diff
@@ -1,7 +1,7 @@
 import Store = require('electron-store');
 import {is} from 'electron-util';
 
-type StoreType = {
+export type StoreType = {
```
sindresorhus: Why are you exporting this? It's not used.

source/util.ts
```diff
@@ -49,3 +50,34 @@ export function stripTrackingFromUrl(url: string): string {
 
 	return url;
 }
+
+export const toggleTrayIcon = (): void => {
+	config.set('showTrayIcon', !config.get('showTrayIcon'));
+	config.get('showTrayIcon') ? tray.create(getWindow()) : tray.destroy();
```
sindresorhus: Save `config.get('showTrayIcon')` in a variable above and ten reuse it here. This should also use an if statement and not a ternary. Ternaries are not meant for function calls.

source/util.ts
```diff
@@ -49,3 +50,34 @@ export function stripTrackingFromUrl(url: string): string {
...
+	if (config.get('launchMinimized')) {
+		if (!config.get('showTrayIcon')) {
+			toggleTrayIcon();
+		}
+
+		disableMenuItem(showTrayIconItem, true);
+
+		dialog.showMessageBox({
+			type: 'info',
+			message: 'Show Tray Icon option has been locked on enabled due to enabled Launch Minimized option.',
```
```suggestion
			message: 'The “Show Tray Icon” setting is force-enabled while the “Launch Minimized” setting is enabled.',
```

source/config.ts
JBudny: I've cherry-picked this from another branch I closed and improved it a little. Somehow I forget to delete this export. Thanks for catching it.

source/util.ts
JBudny: Thank you for pointing me out that misuse of ternary operators. I wasn't aware of it. And sorry you had to wait so long for those changes.

sindresorhus:
Thanks :)

PR #1216, opened 2019-12-12
Prevent user from hiding tray icon when the Launch Minimized option i…

JBudny:
…s enabled.

System: Windows 7

Fixed problem:
I've noticed that Show Tray Icon and Launch Minimized options can be mutually exclusive. 
To reproduce the problem, set:
-File->Caprine Settings->Show Tray Icon on false
-File->Caprine Settings->Launch Minimized on the true
then close application and open it again. The process is running but you have no way to maximize the app window.

I've resolved this problem by enforcing Show Tray Icon on enabled when Launch Minimized is enabled.

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -1,7 +1,7 @@
 import Store = require('electron-store');
 import {is} from 'electron-util';
 
-type StoreType = {
+export type StoreType = {
 	followSystemAppearance: boolean;
 	darkMode: boolean;
 	privateMode: boolean;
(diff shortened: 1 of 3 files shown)


sindresorhus:
Why are you closing?

JBudny:
I realized it could be a part of #1215 and started to work on this there. Maybe I closed it too hastily. Tell me if it's too much for just one PR and I reopen this PR.

PR #1232, opened 2020-01-06
Show all conversations in touchbar and add emojis to touchbar

gohlinka2:
Nested touchbar conversation selector into a popover and added left and right arrow buttons to allow navigation between more conversation pages (not show only 3 conversations, like before). Added feature to touchbar that allows the user to input emoji from the touchbar to the message input.
Fixed selector for inputField in `sendReply`.


sindresorhus:
I think it would be better to use https://electronjs.org/docs/api/touch-bar-scrubber instead of manual previous/next buttons.

sindresorhus:
I think we could better optimize how it works.

- When a text input is not focused, only show the list of conversations.
- When a text input is focused, show a button to show conversations in a popover, and use the rest of the space for an emoji scrubber.

sindresorhus:
Would also be nice to show the users recent and favorite emoji. Which can be read from `~/Library/Preferences/com.apple.CharacterPicker.plist`.

gohlinka2:
@sindresorhus 
Sorry for a long delay with my response. 

> I think it would be better to use https://electronjs.org/docs/api/touch-bar-scrubber instead of manual previous/next buttons.

I initially wanted to use the Scrubber, but i couldn't get it to work correctly. It couldn't differentiate between a scroll gesture and a click well, it always intercepts a click event at the end of a scroll. Maybe the API for it is not finished, in the docs for `TouchBar` it says that it is experimental. I think that, maybe, the previous solution was showing just three conversations because they couldn't get the scrubber to work too. If you know anything more about it, please let me know.

> When a text input is focused, show a button to show conversations in a popover, and use the rest of the space for an emoji scrubber.

Can't be done unless the scrubber works.

> Would also be nice to show the users recent and favorite emoji. Which can be read from `~/Library/Preferences/com.apple.CharacterPicker.plist`.

I'll have a look.

Also, regarding the text focusing and inputting, is there any way to know if the input is focused in the project? Something like `ipc.on("input-focused")...`. I searched through the project but couldn't find anything. Same for the text inputting, is there any function to insert text into the message field. I also found some bugs while using my solution - when the emoji gets inserted into the field, sometimes it isn't possible to delete inputted text, and also in a group chat, when you tag someone and then enter an emoji, it stops working. So the current solution is buggy and I'll have to rework it. Any clues?

sindresorhus:
> It couldn't differentiate between a scroll gesture and a click well

That sounds strange. The whole point of the scrubber is to pick items.

> Maybe the API for it is not finished, in the docs for TouchBar it says that it is experimental.

All the Touch Bar APIs are marked as experimental. Nothing special with this one. I would recommend trying to create a small test case for this and then report the bug to the Electron project.

> is there any way to know if the input is focused in the project?

Nothing built-in here. You can compare `document.activeElement` against the input to see if it's focused.

MaticConradi:
Are those merge conflicts going to be resolved? Touch bar emoji picker is the only thing that's missing for me as far as I can tell 👌

gohlinka2:
@MaticConradi I am unfortunately quite busy now 😕 But I'm glad someone would like this feature, so I'll try to find time to finish it soon 🙂

sindresorhus:
Bump

gohlinka2:
> Bump
I must apologize, but I can't find time to finish this, so I'll close this PR 😕If anybody wants to give it a shot, feel free to reopen and finish it.

PR #1234, opened 2020-01-10
Fix an issue where emoji & messenger preferences menus don't open (macOS)

### LINKED ISSUE #1185: App goes blank after clicking emoji button
(I have the app set to dark mode)
When I try to click on the emoji button in the chat window, your messenger app goes black. If I want to fix it I need to quit the app and relaunch it. But the same thing happens again when I click on the emoji button, so it is really easy to reproduce.

---

Caprine 2.41.1
Electron 7.1.1
darwin 18.7.0
Locale: nb

n-nikol98:
System: macOS 10.15.2

Conversation -> Insert Emoji & the corresponding chat button would black / white screen the app and make it unusable.
Caprine -> Messenger Preferences... would not open under most* circumstances.

* At certain rare times this function was operational, however I have been unable to reproduce conditions under which it worked.

Fixes #1185

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -4,7 +4,7 @@ import elementReady = require('element-ready');
 import selectors from './browser/selectors';
 import config from './config';
 import {toggleVideoAutoplay} from './autoplay';
-import {createConversationList} from './browser/conversation-list';
+import './browser/conversation-list';
 
 const selectedConversationSelector = '._5l-3._1ht1._1ht2';
 const preferencesSelector = '._10._4ebx.uiLayer._4-hy';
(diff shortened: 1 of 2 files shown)


source/browser.ts
```diff
@@ -276,16 +276,11 @@ function setDarkMode(): void {
...
+function setPrivateMode(): void {
 	document.documentElement.classList.toggle('private-mode', config.get('privateMode'));
 
-	if (is.macos) {
-		if (config.get('privateMode')) {
-			ipc.send('hide-touchbar-labels');
-		} else {
-			const conversationsToRender: Conversation[] = await createConversationList();
-			ipc.send('conversations', conversationsToRender);
-		}
```
CvX: Removing this means that conversations no longer show up in the TouchBar, and the private mode info doesn't go away after it's disabled.

Issue #1249, opened 2020-01-21
Reloading with vibrancy enabled makes the background white


dapetcu21:
On macOS, when you reload the app for any reason (for example Cmd+R) and vibrancy enabled, the background gets replaced with pure white (even in dark mode).

Possibly related to #737, but has nothing to do with switching chats.

PR #1250, opened 2020-01-22
[Fix] Vibrancy issue on app refresh

whitecrownclown:
Fixes: #1249


whitecrownclown:
I was surprised as well.. tried a bunch of things until i found this. 😄

dapetcu21:
Cool. Thanks! That magic value though. 😅

PR #1255, opened 2020-01-27
[Fix] Emoji picker crashes app

### LINKED ISSUE #1185: App goes blank after clicking emoji button

whitecrownclown:
Fixes #1185 
Closes #1234

Please have a look, thanks! 
cc @CvX @sindresorhus

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -277,11 +277,11 @@ function setDarkMode(): void {
 	updateVibrancy();
 }
 
-async function setPrivateMode(): Promise<void> {
+function setPrivateMode(): void {
 	document.documentElement.classList.toggle('private-mode', config.get('privateMode'));
 
 	if (is.macos) {
-		await sendConversationList();
+		sendConversationList();
 	}
 }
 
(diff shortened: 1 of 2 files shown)


source/browser.ts
```diff
@@ -281,12 +281,7 @@ async function setPrivateMode(): Promise<void> {
 	document.documentElement.classList.toggle('private-mode', config.get('privateMode'));
 
 	if (is.macos) {
-		if (config.get('privateMode')) {
-			ipc.send('hide-touchbar-labels');
-		} else {
-			const conversationsToRender: Conversation[] = await createConversationList();
-			ipc.send('conversations', conversationsToRender);
-		}
+		await sendConversationList();
```
CvX: If we no longer await for `setPrivateMode()` maybe it doesn't have to be async anymore?

whitecrownclown: We need it to be async since we’re doing sendConversationList

CvX: But we're don't need to await for `sendConversationList()` anymore either, since we don't do anything after it is called, right?

whitecrownclown:
I think you’re right. I’ll update the PR.

Issue #848, opened 2019-04-09
Use `electron-better-ipc` to simplify IPC code


sindresorhus:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/caprine/issues/848/badge.svg" />](https://issuehunt.io/r/sindresorhus/caprine/issues/848)

We do a lot of back and forth IPC. Using [`electron-better-ipc`](https://github.com/sindresorhus/electron-better-ipc) would simplify a lot of code dealing with IPC.

This issue is blocked on: https://github.com/sindresorhus/electron-better-ipc/issues/7

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/caprine/issues/848).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/caprine/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

IssuehuntBot:
@issuehunt has funded $60.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/42574339/issues/848) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/42574339/issues/848).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

PR #1266, opened 2020-02-04
Use `electron-better-ipc` instead of the built-in IPC

dusansimic:
As @sindresorhus mentioned in an issue https://github.com/sindresorhus/caprine/issues/848, Caprine should use [`electron-better-ipc`](https://github.com/sindresorhus/electron-better-ipc) so code would be more simple and readable.

Not a lot of code was removed but some unnecessary event listeners were removed because of `electron-better-ipc` features for sending and receiving data in a single emit.

An important thing to check would be the refactoring of the [`renderEmoji()`](https://github.com/sindresorhus/caprine/compare/master...dusansimic:add-electron-better-ipc?expand=1#diff-42884126f5a9ea114868102427dd7a72R234) function in `emoji.ts`. A lot of code is removed there since all the checks done by that function are actually implemented in `electron-better-ipc`. By that I mean that the if statement used to check if the correct emoji is received as a response because electron ipc doesn't handle unique requests on the same channel but rather all the response events are handled by the first listener that catches them, is no longer needed since `electron-better-ipc` handles that problem with ids for each ipc call.

I'd like to thank @felixfbecker who helped me out with the `renderEmoji()` refactoring by explaining the behavior of generic ipc.

Closes: https://github.com/sindresorhus/caprine/issues/848

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
 	},
 	"dependencies": {
 		"@sindresorhus/do-not-disturb": "^1.1.0",
+		"electron-better-ipc": "^0.8.0",
 		"electron-context-menu": "^0.15.1",
 		"electron-debug": "^3.0.1",
 		"electron-dl": "^1.14.0",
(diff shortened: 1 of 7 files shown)


dusansimic:
Awesome!

sindresorhus:
Excellent work 👌

PR #1270, opened 2020-02-06
Persist maximized state and handle title bar double-click

### LINKED ISSUE #1268: Restore maximized state of the window
See: https://github.com/sindresorhus/caprine/pull/1158

whitecrownclown:
Closes #1268 

Also added maximize toggle when double-clicking the title bar. I think it's expected behaviour.

Thoughts?
cc @CvX @sindresorhus

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -584,11 +584,13 @@ document.addEventListener('DOMContentLoaded', async () => {
 // Toggle maximized state when double clicking the titleBar
 window.addEventListener('dblclick', (event: Event) => {
 	const target = event.target as HTMLElement;
-	const dblClickTitleBar = target.closest('._36ic._5l-3,._5742,._6-xk,._673w');
+	const titleBar = target.closest('._36ic._5l-3,._5742,._6-xk,._673w');
 
-	if (dblClickTitleBar) {
-		ipc.callMain('titlebar-dblclick');
-	}
+	if (!titleBar) {
+		return
+	};
+
+	ipc.callMain('titlebar-doubleclick');
 }, {
 	passive: true
 });
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -1,4 +1,3 @@
-import {remote} from 'electron';
 import {ipcRenderer as ipc} from 'electron-better-ipc';
 import {api, is} from 'electron-util';
 import elementReady = require('element-ready');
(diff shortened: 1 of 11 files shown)


source/browser.ts
```diff
@@ -567,6 +567,15 @@ document.addEventListener('DOMContentLoaded', async () => {
 	// Activate Private Mode if it was set before quitting
 	setPrivateMode();
 
+	// Toggle maximized state when double clicking the titleBar
+	window.addEventListener('dblclick', (event: Event) => {
```
sindresorhus: This needs to respect macOS settings: https://github.com/electron/electron/issues/16385

source/index.ts
```diff
@@ -344,7 +344,18 @@ function createMainWindow(): BrowserWindow {
...
+	});
+
+	win.on('maximize', () => {
+		const lastWindowState = config.get('lastWindowState');
+		config.set('lastWindowState', {...lastWindowState, isMaximized: true});
+	});
+
+	win.on('unmaximize', () => {
+		const lastWindowState = config.get('lastWindowState');
+		config.set('lastWindowState', {...lastWindowState, isMaximized: false});
```
```suggestion
		config.set('lastWindowState.isMaximized', false);
```

source/browser.ts
whitecrownclown: I didn't know about this. I've updated the PR.

source/index.ts
whitecrownclown: The issue here was that TS kept saying it didn't match anything from the store definition. I've added // @ts-ignore to work around this.

source/browser.ts
```diff
@@ -565,6 +566,26 @@ document.addEventListener('DOMContentLoaded', async () => {
...
+		const target = event.target as HTMLElement;
+
+		if (target.closest('._6-xk') || target.closest('._673w')) {
```
sindresorhus: `target.closest('._6-xk') || target.closest('._673w')` should be moved out into a variable with a descriptive name.

source/browser.ts
```diff
@@ -565,6 +566,26 @@ document.addEventListener('DOMContentLoaded', async () => {
...
+			if (is.macos) {
+				const doubleClickAction = remote.systemPreferences.getUserDefault('AppleActionOnDoubleClick', 'string');
```
sindresorhus: I think it would be better to have all this logic in the main process as we already have to call main anyway, and in future Electron versions `systemPreferences` will not be available in the browser process.

source/browser.ts
```diff
@@ -581,6 +581,18 @@ document.addEventListener('DOMContentLoaded', async () => {
 	toggleVideoAutoplay();
 });
 
+// Toggle maximized state when double clicking the titleBar
+window.addEventListener('dblclick', (event: Event) => {
+	const target = event.target as HTMLElement;
+	const dblClickTitleBar = target.closest('._36ic._5l-3,._5742,._6-xk,._673w');
```
sindresorhus:
```suggestion
	const titleBar = target.closest('._36ic._5l-3,._5742,._6-xk,._673w');
```

source/browser.ts
```diff
@@ -581,6 +581,18 @@ document.addEventListener('DOMContentLoaded', async () => {
...
+
+	if (dblClickTitleBar) {
```
sindresorhus: Do an early return.

source/browser.ts
```diff
@@ -581,6 +581,18 @@ document.addEventListener('DOMContentLoaded', async () => {
...
+		ipc.callMain('titlebar-dblclick');
```
sindresorhus:
```suggestion
		ipc.callMain('titlebar-doubleclick');
```

sindresorhus:
Thanks for fixing 👍

Issue #209, opened 2017-05-16
Call sound should also be disabled


fregante:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2450%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/caprine/issues/209)

<img width="554" alt="Settings: sounds disabled" src="https://cloud.githubusercontent.com/assets/1402241/26092468/9f9c5e68-3a43-11e7-8947-56c5bc0f2067.png">

The Sound is disabled in the settings, but I still hear the sound when I receive a call.

Calls also seem to be bypass the Desktop Notification setting.

Edit: This is not a Caprine bug but a Messenger "feature". Perhaps Caprine can fix it.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $50.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($50.00)

### Submitted pull Requests
- [#1273 Call ringtone muted](https://issuehunt.io/r/sindresorhus/caprine/pull/1273)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/caprine/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

piomar123:
What's your actual use case of this? Do you want to disable all the sounds or allow to have notification sounds but mute ringing only?

piomar123:
My current idea is to use `setAudioMuted(true)`. It seems to work but if you want to actually answer or make a voice call, you need to disable the mute (it would be nice to detect it automatically).
You can find my draft here: https://github.com/piomar123/caprine/tree/209-disable-call-sound

IssuehuntBot:
@issuehunt has funded $50.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/42574339/issues/209) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/42574339/issues/209).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
[...]

PR #1273, opened 2020-02-07
Add preference to mute call ringtone

LuKks:
fix https://github.com/sindresorhus/caprine/issues/209

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#209: Call sound should also be disabled](https://issuehunt.io/repos/42574339/issues/209)
---

</details>

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -241,8 +241,9 @@ function initRequestsFiltering(): void {
 			return callback({});
 		}
 
+		const callRingtoneHash = '2NAu/QVqg211BbktgY5GkA==';
 		callback({
-			cancel: responseHeaders['content-md5'][0] === '2NAu/QVqg211BbktgY5GkA=='
+			cancel: responseHeaders['content-md5'][0] === callRingtoneHash
 		});
 	});
 }
--- a/source/index.ts
+++ b/source/index.ts
@@ -234,12 +234,17 @@ function initRequestsFiltering(): void {
 		}
 	});
 
-	session.defaultSession.webRequest.onHeadersReceived({ urls: [`*://static.xx.fbcdn.net/rsrc.php/*`] }, ({responseHeaders}, callback) => {
+	session.defaultSession.webRequest.onHeadersReceived({
+		urls: ['*://static.xx.fbcdn.net/rsrc.php/*']
+	}, ({responseHeaders}, callback) => {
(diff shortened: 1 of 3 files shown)


source/index.ts
```diff
@@ -233,6 +233,18 @@ function initRequestsFiltering(): void {
...
+
+	session.defaultSession.webRequest.onHeadersReceived({
+		urls: ['*://static.xx.fbcdn.net/rsrc.php/*']
+	}, ({responseHeaders}, callback) => {
+		if (!config.get('callRingtoneMuted') || !responseHeaders) {
+			return callback({});
+		}
+
+		callback({
+			cancel: responseHeaders['content-md5'][0] === '2NAu/QVqg211BbktgY5GkA=='
```
sindresorhus: Can you move `'2NAu/QVqg211BbktgY5GkA=='` to a variable with a name that indicates what it is?

source/menu.ts
```diff
@@ -247,6 +247,14 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				sendAction('toggle-mute-notifications');
 			}
 		},
+		{
+			label: 'Mute Ringtone Call',
```
sindresorhus:
```suggestion
			label: 'Mute Call Ringtone',
```

source/index.ts
```diff
@@ -233,6 +233,18 @@ function initRequestsFiltering(): void {
 			callback({cancel: config.get('block.deliveryReceipt' as any)});
 		}
 	});
+
+	session.defaultSession.webRequest.onHeadersReceived({
+		urls: ['*://static.xx.fbcdn.net/rsrc.php/*']
+	}, ({responseHeaders}, callback) => {
+		if (!config.get('callRingtoneMuted') || !responseHeaders) {
+			return callback({});
```
```suggestion
		callback({});
		return;
```

sindresorhus:
Have you manually tested that this actually works, by receiving a call and ensuring it doesn't make a sound?

LuKks:
Yes. This is what I commented in the issue discussion some time ago:
> Due HTTP cache, for example, if the ringtone plays and now you disable it, the sound is already cached so you will need to restart the app.

I hope that is ok and I also tested it again, it works fine. By the way, really thank you for the corrections in the code.

PR #1280, opened 2020-02-18
Automatically reset window position when disconnecting secondary screen

tomioe:
Added a few features that I find immensely useful in my everyday use of Caprine (on Windows 10).

1. When going from multi-screen to single-screen (i.e. when undocking laptop) the main window of Caprine will sometimes be stuck off-screen (cannot even using Win+Arrow Keys to move it). Having a 'Reset Position' in the tray menu helps this problem.

2. I try to minimize my mouse movement as much as possible, but using Alt to toggle the menu and going into Exit simply isn't fast enough. I believe Command+Q is automatically bound in macOS to make an application quit, but not so on Windows, so I have added the Ctrl+Q keybinding to quit Caprine.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -573,7 +573,6 @@ document.addEventListener('DOMContentLoaded', async () => {
 	// Prevent flash of white on startup when in dark mode
 	// TODO: find a CSS-only solution
 	if (!is.macos && config.get('darkMode')) {
-		// eslint-disable-next-line require-atomic-updates
 		document.documentElement.style.backgroundColor = '#1e1e1e';
 	}
 
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -573,6 +573,7 @@ document.addEventListener('DOMContentLoaded', async () => {
 	// Prevent flash of white on startup when in dark mode
 	// TODO: find a CSS-only solution
 	if (!is.macos && config.get('darkMode')) {
+		/* eslint-disable-next-line require-atomic-updates */
 		document.documentElement.style.backgroundColor = '#1e1e1e';
 	}
 
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -573,7 +573,7 @@ document.addEventListener('DOMContentLoaded', async () => {
 	// Prevent flash of white on startup when in dark mode
 	// TODO: find a CSS-only solution
 	if (!is.macos && config.get('darkMode')) {
-		/* eslint-disable-next-line require-atomic-updates */
+		// eslint-disable-next-line require-atomic-updates
 		document.documentElement.style.backgroundColor = '#1e1e1e';
 	}
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -92,7 +92,7 @@ app.on('second-instance', () => {
 	}
 });
 
-// Main window should move to the same position as it had on another screen
+// Preserves the window position when a display is removed and Caprine is moved to a different screen.
 app.on('ready', () => {
 	electronScreen.on('display-removed', () => {
 		const [x, y] = mainWindow.getPosition();
--- a/source/index.ts
+++ b/source/index.ts
@@ -95,7 +95,7 @@ app.on('second-instance', () => {
 // Main window should move to the same position as when it was another screen
 app.on('ready', () => {
 	electronScreen.on('display-removed', () => {
-		const oldPosition = mainWindow.getPosition();
+		const [x, y] = mainWindow.getPosition();
 		mainWindow.setPosition(oldPosition[0], oldPosition[1]);
 	});
 });
--- a/source/index.ts
+++ b/source/index.ts
@@ -92,11 +92,11 @@ app.on('second-instance', () => {
 	}
 });
 
-// Main window should move to the same position as when it was another screen
+// Main window should move to the same position as it had on another screen
 app.on('ready', () => {
 	electronScreen.on('display-removed', () => {
 		const [x, y] = mainWindow.getPosition();
-		mainWindow.setPosition(oldPosition[0], oldPosition[1]);
+		mainWindow.setPosition(x, y);
 	});
 });
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -92,6 +92,14 @@ app.on('second-instance', () => {
 	}
 });
 
+// Main window should move to the same position as when it was another screen
+app.on('ready', () => {
+	electronScreen.on('display-removed', () => {
+		const oldPosition = mainWindow.getPosition();
+		mainWindow.setPosition(oldPosition[0], oldPosition[1]);
+	});
+});
+
 function getMessageCount(conversations: Conversation[]): number {
 	return conversations.filter(({unread}) => unread).length;
 }
--- a/source/menu.ts
+++ b/source/menu.ts
@@ -720,6 +720,7 @@ ${debugInfo()}`;
 	const linuxWindowsTemplate: MenuItemConstructorOptions[] = [
 		{
 			role: 'fileMenu',
+			label: '&File',
 			submenu: [
 				newConversationItem,
 				{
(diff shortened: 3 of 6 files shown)


tomioe:
For some reason I am getting an `require-atomic-updates rule is disabled but never reported` when building on `source\browser.ts:581:31` and `source\browser\conversation-list.ts:90:32`.

sindresorhus:
> When going from multi-screen to single-screen (i.e. when undocking laptop) the main window of Caprine will sometimes be stuck off-screen (cannot even using Win+Arrow Keys to move it). Having a 'Reset Position' in the tray menu helps this problem.

I guess this is the cause: https://github.com/electron/electron/issues/10862

sindresorhus:
> For some reason I am getting an require-atomic-updates rule is disabled but never reported when building on source\browser.ts:581:31 and source\browser\conversation-list.ts:90:32

That means you can just remove those disable comments.

sindresorhus:
Can you try to do this automatically? You could listen to this event https://www.electronjs.org/docs/api/screen#event-display-removed and then reset the position. I wonder if after resetting the position, we could set the saved position again and it would be correct. Worth a shot.

tomioe:
> 
> 
> > For some reason I am getting an require-atomic-updates rule is disabled but never reported when building on source\browser.ts:581:31 and source\browser\conversation-list.ts:90:32
> 
> That means you can just remove those disable comments.

[Removing those disable comments](https://github.com/sindresorhus/caprine/pull/1280/commits/fbdadf3aed5c24f83ae81cffcd7b4925bd572bd9) makes my build pass locally, but fails when [building in Travis](https://travis-ci.org/sindresorhus/caprine/builds/652109923). For some reason, I don't get the `Possible race condition...` error when building locally.

tomioe:
> 
> 
> Can you try to do this automatically? You could listen to this event https://www.electronjs.org/docs/api/screen#event-display-removed and then reset the position. I wonder if after resetting the position, we could set the saved position again and it would be correct. Worth a shot.

Will give this a shot, would definitely be smoother to have it happen automatically.

tomioe:
> 
> 
> Can you try to do this automatically? You could listen to this event https://www.electronjs.org/docs/api/screen#event-display-removed and then reset the position. I wonder if after resetting the position, we could set the saved position again and it would be correct. Worth a shot.

I've tried an implementing this in [9f6e1ad](https://github.com/sindresorhus/caprine/pull/1280/commits/9f6e1adb3b92e705f7881bd325ee076edb3d9b05). It can be tested in Windows (if using two screens), by moving the Caprine main window to the secondary screen and setting the "Project" (Win+P) setting to "PC Screen Only". The Caprine window position should move to the same position on the primary screen as it had on the secondary screen.

source/index.ts
```diff
@@ -92,6 +92,14 @@ app.on('second-instance', () => {
 	}
 });
 
+// Main window should move to the same position as when it was another screen
+app.on('ready', () => {
+	electronScreen.on('display-removed', () => {
+		const oldPosition = mainWindow.getPosition();
```
```suggestion
		const [x, y] = mainWindow.getPosition();
```

source/index.ts
```diff
@@ -92,6 +92,14 @@ app.on('second-instance', () => {
 	}
 });
 
+// Main window should move to the same position as when it was another screen
```
sindresorhus: This sentence is not grammatically correct.

source/tray.ts
```diff
@@ -73,6 +73,14 @@ export default {
...
 				}
 			},
+			{
+				label: 'Reset Position',
+				visible: !is.macos,
+				click() {
+					win.setPosition(0, 0);
+				}
+
+			},
```
sindresorhus: This is no longer needed.

source/browser.ts
```diff
@@ -573,7 +573,7 @@ document.addEventListener('DOMContentLoaded', async () => {
 	// Prevent flash of white on startup when in dark mode
 	// TODO: find a CSS-only solution
 	if (!is.macos && config.get('darkMode')) {
-		// eslint-disable-next-line require-atomic-updates
+		/* eslint-disable-next-line require-atomic-updates */
```
sindresorhus: Don't do unrelated changes.

sindresorhus:
> I believe Command+Q is automatically bound in macOS to make an application quit, but not so on Windows, so I have added the Ctrl+Q keybinding to quit Caprine.

Is that really a Windows convention? Do other Windows app support that. Because the menu item is named "Exit" on windows, so "Q" doesn't really make sense.

tomioe:
> 
> 
> > I believe Command+Q is automatically bound in macOS to make an application quit, but not so on Windows, so I have added the Ctrl+Q keybinding to quit Caprine.
> 
> Is that really a Windows convention? Do other Windows app support that. Because the menu item is named "Exit" on windows, so "Q" doesn't really make sense.

Huh, I must've picked up that bad habit over the years... I just assumed it was a conventional shortcut, due to how closely it is to Cmd+Q, but I was wrong: https://support.microsoft.com/en-us/help/12445/windows-keyboard-shortcuts

It could be argued, that although uncommon, it is still implemented in some applications (i.e. Firefox uses `Ctrl+Shift+Q`), while others seem to favor the more idiomatic `Alt+F, X`:https://www.ghacks.net/2018/10/22/google-retires-ctrl-shift-q-in-chrome-to-exit-web-browser/

Currently `Alt+F, X` does not work in Carpine, since underlined letters for keyboard navigation have not been implemented. Perhaps this is a better suited goal of this PR?

sindresorhus:
> Currently Alt+F, X does not work in Carpine, since underlined letters for keyboard navigation have not been implemented. Perhaps this is a better suited goal of this PR?

Yes, but that's really out of the scope of this PR and should be a separate PR.

source/menu.ts
```diff
@@ -744,23 +745,27 @@ ${debugInfo()}`;
...
 				{
+					label: 'E&xit',
 					role: 'quit'
 				}
 			]
 		},
 		{
-			role: 'editMenu'
+			role: 'editMenu',
+			label: '&Edit'
```
sindresorhus: I think you should rather open an issue on Electron about this. I'm already using the `role`, which should give the correct automatic behavior. If `&Edit` (and others) is indeed correct on Windows, it should be handled by Electron.

source/index.ts
```diff
@@ -92,6 +92,14 @@ app.on('second-instance', () => {
 	}
 });
 
+// Main window should move to the same position as it had on another screen
```
```suggestion
// Preserves the window position when a display is removed and Caprine is moved to a different screen.
```

sindresorhus:
The PR title needs to be updated.

source/menu.ts
tomioe: It seems like the default roles do not generate these accelerators: https://github.com/electron/electron/blob/a5f87cee6b79e2ed38fbfe7b8af26a9de0649658/lib/browser/api/menu-item-roles.js#L187-L193

The documentation also calls for using '&', which I believe is because it does not want to be too presumptuous about application keyboard shortcuts: https://www.electronjs.org/docs/api/menu#menusetapplicationmenumenu

But as you said, this is out of scope for this PR, I will create a separate PR for this.

source/index.ts
tomioe: Appreciated, I was struggling with wording on this one ...

source/menu.ts
sindresorhus: I still think it's worth opening an Electron issue. If it doesn't make sense to add default keyboard shortcuts for the defined role menus, I'm not sure why apps should do it at all.

source/menu.ts
```diff
@@ -756,7 +756,6 @@ ${debugInfo()}`;
 			submenu: viewSubmenu
 		},
 		{
-			label: 'Conversation',
```
sindresorhus: ?

tomioe: Absolute blunder on my part, apologies...

PR #1289, opened 2020-02-29
Add remember option checkbox in private mode dialog

### LINKED ISSUE #1186: feature: remove blur for `active now` and last seen status
remove blur for user active and last seen status below name in private mode. and also that popup is annoying on Ctrl+Shift+N (active private mode) it will be good if there is a checkbox for `Don't ask again`. btw love this project <3

panda7789:
Resolves #1186 

I have already looked at this problem before but Electron had a bug with the checkbox. electron/electron#21159

I had to use async function because sync doesn't return information if the check was checked or not.
Also when you once check the check you have to 'Debug->Delete settings' if you want to see dialog again.

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -5,6 +5,7 @@ type StoreType = {
 	followSystemAppearance: boolean;
 	darkMode: boolean;
 	privateMode: boolean;
+	showPrivateModeModal: boolean;
 	vibrancy: 'none' | 'sidebar' | 'full';
 	zoomFactor: number;
 	lastWindowState: {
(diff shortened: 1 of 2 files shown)


source/config.ts
```diff
@@ -5,6 +5,7 @@ type StoreType = {
 	followSystemAppearance: boolean;
 	darkMode: boolean;
 	privateMode: boolean;
+	showPrivateModeModal: boolean;
```
sindresorhus:
```suggestion
	showPrivateModePrompt: boolean;
```

source/menu.ts
```diff
@@ -424,27 +424,31 @@ Press Command/Ctrl+R in Caprine to see your changes.
...
-						cancelId: 1
-					}) === 0;
-
-					if (!confirmPrivateMode) {
-						menuItem.checked = false;
-						return;
-					}
+						cancelId: 1,
+						checkboxLabel: 'Don\'t ask me again'
+					}).then(res => {
```
sindresorhus: Don't use `.then`. You can just use `await`.

sindresorhus:
```suggestion
					}).then(result => {
```

PR #1306, opened 2020-03-17
Fix the sidebar config migration

### LINKED ISSUE #1305: MacOs , ver. 2.43.0 Caprine Sidebar width setting doesn't persist after quitting the app
I Like to have wide sidebar all the time, so i'm using View -> Sidebar -> Wide Sidebar option checked, but after quitting the app this setting is always set to Auto, and I have to manually set it back to Wide

whitecrownclown:
Fixes #1305

Summary:

- The issue was that the config would always be overwritten for those who didn't have the `sidebarHidden` key to always `default`.
- After setting the new migration key, we should also delete the old one so we don't reset it to `hidden` every time.
- I've also deleted a IPC call that looks out of place.

cc @sindresorhus @CvX

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -319,8 +319,6 @@ function updateSidebar(): void {
 			break;
 		default:
 	}
-
-	ipc.callMain('set-sidebar');
 }
 
 async function updateDoNotDisturb(): Promise<void> {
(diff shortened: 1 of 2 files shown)


CvX:
Great work! 😊

PR #1325, opened 2020-04-06
Fix making and receiving calls

### LINKED ISSUE #1189: Calls not working
It shows only a blank white window. Can't receive call either. It works on Chrome.

wereii:
Workaround electron's location handling on new windows [window.open()]

Fixes #1189

I also described the full solution in my [comment](https://github.com/sindresorhus/caprine/issues/1189#issuecomment-609765491) on the issue. 

> - Doing newWindow.location = "/path" fails because at it seems, by default electron is trying to join the new-window's location (that window's window.location) with the new location (/path), but the new-window doesn't have valid location.href because it wasn't opened with one.
>     `w1 = window.open() // w1.location.href === "about:blank"`
>     In normal browser, as long as your new location is not valid URL on it's own (?) it would join the new path with current's window (the window from where you are opening the new-window) location, not the new-window's one.
> 
> **Solution**
> 
> - When open of Group Calls window is detected, additionally pre-set the location to `https://www.messenger.com`.
>

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -307,6 +307,7 @@ function createMainWindow(): BrowserWindow {
 		darkTheme: isDarkMode, // GTK+3
 		webPreferences: {
 			preload: path.join(__dirname, 'browser.js'),
+			nativeWindowOpen: true,
 			contextIsolation: true,
 			plugins: true
 		}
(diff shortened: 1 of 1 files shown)


wereii:
Ah, after a bit more testing this (simple) solution is not reliable.
I was testing it by the "Join call" in a group, or after the call ends, there is a button to call others again.

Tried direct calling to someone or someone calling me and that isn't working, still fails with blank window and errors like
`LocationProxy: failed to parse string  TypeError: Failed to construct 'URL': Invalid URL`. 

Shifting back to `nativeWindowOpen: true`.

CvX:
Tried it, and I could successfully call someone, end the call and redial within the call window, and was able to answer a call. Great work! 😊 🚀

CvX:
Released in 2.45.0

PR #1347, opened 2020-04-28
add webframe zoom in and out

abbychau:
(code for toggling conversation sidebar is inside.


sindresorhus:
What problem is this solving?

Also note from the top of the readme:

> Caprine is feature complete.

abbychau:
we cannot zoom out and zoom in the whole app. but the conversation text only.
zooming out the whole app is very useful for me.

sindresorhus:
Sorry, this is not something I want in Caprine.

PR #1369, opened 2020-05-31
Add option to disable spell checker

fnesveda:
The spell checker can be quite annoying at times, so I've added an option to Caprine's preferences to configure whether it is enabled.
I couldn't find a good way to update the BrowserWindow webPreferences option during runtime, so unfortunately a restart is needed to configure the spell checker.

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -40,6 +40,7 @@ type StoreType = {
 	quitOnWindowClose: boolean;
 	keepMeSignedIn: boolean;
 	autoplayVideos: boolean;
+	enableSpellChecker: boolean;
 	spellCheckerLanguages: string[];
 };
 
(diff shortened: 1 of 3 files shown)


source/config.ts
```diff
@@ -40,6 +40,7 @@ type StoreType = {
 	quitOnWindowClose: boolean;
 	keepMeSignedIn: boolean;
 	autoplayVideos: boolean;
+	enableSpellChecker: boolean;
```
sindresorhus, line 43:
```suggestion
	isSpellCheckerEnabled: boolean;
```

source/menu.ts
```diff
@@ -264,6 +264,15 @@ Press Command/Ctrl+R in Caprine to see your changes.
 				config.set('showUnreadBadge', !config.get('showUnreadBadge'));
 			}
 		},
+		{
+			label: 'Enable Spell Checker',
```
sindresorhus, line 268:
```suggestion
			label: 'Spell Checker',
```

fnesveda:
Thanks! ☺

PR #1372, opened 2020-06-01
Invert "Create new room" button

itsmingjie:
Before
image

After
image

PR #1373, opened 2020-06-01
Fix "Create new room" button in dark mode

itsmingjie:
Redo of #1372, sorry for the inconvenience.


sindresorhus:
Have you manually tested that the PR here fixes the problem?

itsmingjie:
Yup -- see the screenshot above. I used the same CSS to test in devtools during the last PR, but got lazy when editing the repo.

Issue #1278, opened 2020-02-14
Messages don't scroll to the bottom


wilg:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2450%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/caprine/issues/1278)

Most of the time, when new messages are received or you switch conversations, the scroll position does not move to the bottom and you have to manually scroll to see conversations.

This is such a frustrating bug, it must only happen in certain scenarios that don't happen to the authors, because the app is so well thought through otherwise. I am not sure what is unusual about what I am doing though.

---

Caprine 2.44.0
Electron 8.0.0
darwin 19.3.0
Locale: en-US

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $50.00)

- [<img src='https://avatars1.githubusercontent.com/u/54353?v=4' alt='wilg' width=24 height=24> wilg](https://issuehunt.io/u/wilg) ($50.00)

### Submitted pull Requests
- [#1295 Autoscroll (#1278) ](https://issuehunt.io/r/sindresorhus/caprine/pull/1295)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/caprine/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

mquevill:
I have this problem too. It doesn't seem to be remembering previous places or anything. It seems to just randomly go up a ways, requiring the user to scroll down to the bottom.

Caprine 2.44.0
Linux Mint 19.3 (installed via .deb)

mmirkov:
+1 

Caprine 2.44.0
Windows 10 (1909)

mquevill:
Actually, this may be an issue inherent in the Messenger interface. Even when switching between chats on messenger.com, this issue occurs. Furthermore, it may be difficult to implement in Electron (see https://github.com/electron/electron/issues/5884). So if Caprine itself wants to force a scroll to the bottom to address this issue, a workaround might be necessary.

NightMean:
[...]

PR #1295, opened 2020-03-06
Added Autoscroll on new messages - FIX issue 1278

fbianconi:
Hi there, I've been working in the autoscroll issue (fixes #1278) that I saw in IssueHunt. Let me know if this makes the cut, or if you need anything else.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#1278: Messages don't scroll to the bottom](https://issuehunt.io/repos/42574339/issues/1278)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

PR #1382, opened 2020-06-20
Fix autoscroll on new messages

dusansimic:
This completes the requested changes in https://github.com/sindresorhus/caprine/pull/1295. It's basically the same PR and most of the code is from @fbianconi. I've made this since there hasn't been a lot of activity on this issue for some time and I wanted to get it fixed.

 - [x] Fix variable names
 - [x] Scroll on conversation change

Closes: https://github.com/sindresorhus/caprine/issues/1278

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -548,12 +548,25 @@ async function observeAutoscroll(): Promise<void> {
 		return;
 	}
 
+	const scrollToBottom = (): void => {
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for finishing this :)

ViRb3:
We need a way to disable this - it is impossible to read history in an active conversation, as it will force your view to the bottom on every new message.

Issue #1321, opened 2020-04-04
How does it compare to Messenger Desktop?


ad31mar:
Now that [it's out](https://www.messenger.com/desktop), how does Caprine compare?

sindresorhus:
Caprine is open-source, has privacy features, is more customizable, and also supports Linux.

1nikolas:
Here is [my opinion](https://github.com/sindresorhus/caprine/issues/1189#issuecomment-607371563) on this. I get the privacy stuff but its running faster and better than Caprine (cause its official)

carlinmack:
I would like someone who is knowledgeable on UWP to chime in on how private the official app must be. Afaik UWP doesn't allow apps to know anything outside of it's container

centraleffects:
> Here is [my opinion](https://github.com/sindresorhus/caprine/issues/1189#issuecomment-607371563) on this. I get the privacy stuff but its running faster and better than Caprine (cause its official)

Does it support Linux? I can't find a way to install it.

1nikolas:
> > Here is [my opinion](https://github.com/sindresorhus/caprine/issues/1189#issuecomment-607371563) on this. I get the privacy stuff but its running faster and better than Caprine (cause its official)
> 
> Does it support Linux? I can't find a way to install it.

No it doesn't.. For linux Caprine is the only solution

johan456789:
> Here is [my opinion](https://github.com/sindresorhus/caprine/issues/1189#issuecomment-607371563) on this. I get the privacy stuff but its running faster and better than Caprine (cause its official)

I have the opposite experience. The official app is slow on mac. Try cmd + w and then open the app again. It takes 1-3 seconds to do so, while it's instantly with Caprine. The UI is better on Caprine. You can reply to previous messages by clicking a button; you need to right click to access the option in the official app. 

I'll still use Caprine for now.

1nikolas:
> You can reply to previous messages by clicking a button; you need to right click to access the option in the official app.
You can reply by double clicking outside of the message (on the same row). That's actually kinda smart instead of having lots of buttons. Also you can react by long pressing the message

arjitj2:
I had to switch to the official simply for the lower memory usage. As @1nikolas pointed out it's not Caprine's fault, the official messenger app just gets to use the mobile interface & APIs which are faster.

As I speak right now the official messenger app is using 57 mb RAM, while Caprine is at ~570 mb.

[...]

Issue #1244, opened 2020-01-19
Memory usage increases upto 1300mb as far as seen. 


shohabmsk:
Just running in the background. The memory usage increases a lot without any message coming or doing anything. Running 64bit win 10 lates t with all updates installed. Didn't check earlier versions of caprine memory usages.

---

Caprine 2.42.0
Electron 7.1.9
win32 10.0.18363
Locale: en-GB

varystargaryen:
Same issue on macOS 10.15.3 Beta (19D62e) and Caprine 2.42.1

dondolarson:
I don't have such an issue as you guys. Autostart run 4 processes and running in background on my machine they takes just under 400MB of memory.

krasnoludkolo:
After  being open ~2 days Caprine currently is using 1,8GB of RAM. (Ubuntu 18.04 LTS)

ShaneOH:
I'm seeing a huge amount of energy consumption by Caprine as well. I'm wondering if this is a Caprine issue or an Electron one.

<img width="800" alt="Screen Shot 2020-01-22 at 4 17 15 PM" src="https://user-images.githubusercontent.com/5799260/72897352-cfab7180-3d32-11ea-9dbf-051a288a29cb.png">

tard3x:
4 processes with ~200MB in Windows 10 Pro 1909. No problem here.

krasnoludkolo:
Imo important question is: how long have the process been on?

dondolarson:
> 4 processes with ~200MB in Windows 10 Pro 1909. No problem here. Latest Caprine 2.42.1 version.

Same here. 200MB and even less, right after restarting my machine. Same setup as you. It's increasing when PC hasn't been restarted for days but I guess it's normal.

whitecrownclown:
This might be fixed by https://github.com/sindresorhus/caprine/pull/1247

jepes1981:
almost a Gig of memory consumption for only 6 hours of use. Running Windows 10 64-bit v1903. Relaunching reverts it back to 196Mb.

dondolarson:
![Caprine](https://user-images.githubusercontent.com/49354563/76768140-e2727d80-679a-11ea-8af5-c0ae24bf102e.jpg)

Everything is fine here with Caprine 2.44.0 on Windows 10 1909 February update. System restart since yesterday 22:19 (17-18 hours ago) and I've used Caprine couple of times since that. It's how it always look for me.

[...]

PR #1410, opened 2020-09-12
Add menu item to relaunch Caprine

kwongtn:
As per issues mentioned in 
https://github.com/sindresorhus/caprine/issues/1321#issuecomment-624578500
https://github.com/sindresorhus/caprine/issues/1244

Caprine uses quite a lot of RAM on extended usage, the best way to reclaim RAM is to close and relaunch Caprine.
To relaunch, we had to:
Close Caprine -> Go to taskbar and close caprine again -> Open Caprine

This patch introduces the option to do so directly from the "file" menu, making relaunching a simple click away.


source/menu.ts
```diff
@@ -51,6 +51,13 @@ export default async function updateMenu(): Promise<Menu> {
 			click() {
 				sendAction('log-out');
 			}
+		},
```
sindresorhus, line 54: There should be a separator here.

source/menu.ts
```diff
@@ -51,6 +51,13 @@ export default async function updateMenu(): Promise<Menu> {
...
+		{
+			label: 'Relaunch App',
```
sindresorhus, line 56: This is in the wrong position. Note the variable name of the array.

It should also be named `Relaunch Caprine`.

sindresorhus (changes requested):

kwongtn:
Hi @sindresorhus 
I've done the changes accordingly, please check 😅 

Sorry if this is wrong again -- I'm really very new to ts and electron 😂

PR #1417, opened 2020-09-26
Fix 2FA login issue

### LINKED ISSUE #1414: YubiKey 2FA not working (probable cause: new window is opened in default system browser?)
YubiKey 2FA doesn't work
Snap:
When trying to login using snap installed version, there is no way to use 2FA (there will be a page saying to insert the YubiKey and press the button, but nothing will happen when inserting or pressing it).
If I click on "try different method", a new windows will open in my browser (Firefox), since I'm already logged in there, it will just show my Facebook.

Deb:
When trying to login using deb package, there will be a prompt requiring YubiKey and inserting it and pressing the button will work, I will get a message that the 2FA login was successful, however there is no way to get past it. Caprine will open new tab in Firefox displaying the Facebook page and nothing will happen in the Caprine window. When I restart Caprine, the whole process will be lost and I will have to start over.

---

Caprine 2.49.0
Electron 10.1.1
linux 5.4.0-47-generic
Locale: en-US

### LINKED ISSUE #1415: Login with 2FA doesn't work
When trying to log in with 2FA enabled, the "Please verify your Facebook Account" page appears. Continuing and submitting the 2FA code gets you back to this page again, regardless of selecting save or don't save browser. After you are redirected, clicking on continue does nothing, Caprine is stuck on the "Please verify your Facebook Account" page. Quitting and restarting the application gets you back to the login page, starting the login progress over. Resetting app data or settings doesn't help.

This makes Caprine completely unusable as you're unable to login. Tried on 3 different devices, all running the same version of Arch Linux.

<hr></hr>

Caprine 2.48.0-1
Electron 9.2.1-1
Linux 5.8.10-arch1
Locale: en-US

alex-the-man:
FB changed their 2FA login url. Without this change, Electron will open the 2FA login page using external browser.

Fixes #1415
Fixes #1414

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -529,7 +529,7 @@ function createMainWindow(): BrowserWindow {
 		};
 
 		const isTwoFactorAuth = (url: string): boolean => {
-			const twoFactorAuthURL = 'https://www.facebook.com/checkpoint/start';
+			const twoFactorAuthURL = 'https://www.facebook.com/checkpoint';
 			return url.startsWith(twoFactorAuthURL);
 		};
 


jan-oratowski (approved):
👍

PR #1420, opened 2020-09-29
Manually specify sized icons for Linux .deb

mquevill:
I know this isn't the ideal way of handling this (i.e. electron-builder **should** be able to automatically create the sized icons), but consider it a hard-coded workaround. This also uses the **new** icon (as of ff22d8090ea2c2b5cbdd8374c556a8ff1e1eae8e).

Tested with Linux Mint 20 (Ubuntu 20.04, kernel 5.4.0-48).

To create the sized icons I used the following Bash script:
```bash
cd build/
for i in 16 32 48 64 128 256 512; do
    convert icon.png -resize ${i}x${i} icons/${i}x${i}.png
done
```

This could probably be incorporated into the build process somehow, but I'm not very familiar with npm or electron-builder.


tracker1:
Would suggest adding this to the build/postbuild process... I know it's easy enough to add these manually, but would be better to keep the source impact minimal.

sindresorhus:
The current change is fine. I don't want to complicate anything for a temporary workaround.

Issue #1155, opened 2019-10-30
dark mode - "you forwarded an image"


rtpHarry:
Found an unstyled text in 2.40.0:

![image](https://user-images.githubusercontent.com/1038062/67884669-25aab780-fb3e-11e9-96cf-65be1bc3c138.png)

Dev tools:

![image](https://user-images.githubusercontent.com/1038062/67884780-4e32b180-fb3e-11e9-8b27-f7df439ff6bc.png)

Repro:

1. Find an image in a conversation
1. Select forward image, and send it to another conversation
1. View the image in the new conversation

dusansimic:
This will be fixed by #1437

PR #1437, opened 2020-11-05
Fix dark mode

dusansimic:
Minor dark mode style fixes:

- [x] Label color above forwarded image (it used to be black)
- [x] Color of inner circle for chat color (it used to be white)

Closes: #1155


CvX (approved):
👍

Issue #1453, opened 2020-11-23
New icon is lowres and buggy


BloodRaven0:
![sshot-2020-11-23-18-44-39](https://user-images.githubusercontent.com/42444745/99990074-0f604100-2dbc-11eb-925b-1f18a49d59b8.png)
The new icon is capped at 38x38 pixels (a weird format, maybe the actual file has empty space around it and it's actually 48x48 or 64x64) on my desktop and won't scale further (signified by the box surrounding it). Also, it is strangely aliased and jaggy-looking. Official messenger app icon for comparison.
![sshot-2020-11-23-18-44-48](https://user-images.githubusercontent.com/42444745/99990079-0ff8d780-2dbc-11eb-8de3-8233e5b17819.png)
Also in the taskbar for some reason it is only rendered halfway. This was reported in another issue, so I'll just link it
https://github.com/sindresorhus/caprine/issues/1429

And the third problem is that with the new messenger UI, the tray icon always stays grey, no matter if you have unread messages.

P.S. I think the old icon without the white background looked better (and certainly worked better lol)
I'm using Windows 10 pro x64

dusansimic:
> The new icon is capped at 38x38 pixels (a weird format, maybe the actual file has empty space around it and it's actually 48x48 or 64x64) on my desktop and won't scale further (signified by the box surrounding it). Also, it is strangely aliased and jaggy-looking. Official messenger app icon for comparison.

This due to the design of the icon. It has some transparent space around it. @CvX a simple solution would be making an icon for MacOS+Linux (the current one) and an icon for Windows (one without the whitespace).

> Also in the taskbar for some reason it is only rendered halfway. This was reported in another issue, so I'll just link it

We'll track that issue on the one you linked, not here.

> And the third problem is that with the new messenger UI, the tray icon always stays grey, no matter if you have unread messages.

This will also be tracked on this issue #1439, not here.

> I think the old icon without the white background looked better (and certainly worked better lol)

These issues are unrelated to the icon design, rather to the Windows feature of adding a dot when there is an unread message.

Issue #1429, opened 2020-10-20
Taskbar icon displayed wrongly


maxoku:
After update taskbar icon (when running) is displayed wrongly, cut off mostly. Only a little from top is visible. When it's closed it's displayed normally, everywhere else too.
Win8.1 64-bit.

---

Caprine 2.51.0
Electron 10.1.3
win32 6.3.9600
Locale: pl

artiste212:
The Mac icon looks different, depending on whether apps are viewed in list or icon mode in the Finder. Curious!

maxoku:
Update. It's displayed wrongly if label is showed, it's okay if label is hidden.

jonyroda97:
![image](https://user-images.githubusercontent.com/8282833/96792691-cf8bef80-13f2-11eb-8011-29e10fec9f01.png)
![image](https://user-images.githubusercontent.com/8282833/96792695-d61a6700-13f2-11eb-9b09-04c19426b49f.png)

Same here. Added images for better understanding of the problem.
Both images are from the taskbar

jed84:
Not to mention, the icon has a wrong size and it doesnt look good on the white backhround and with cropped edges.
![obraz](https://user-images.githubusercontent.com/47121522/99286027-3654cb00-2830-11eb-91f6-1d170002a40c.png)

Can we revert to the previous icon?

kacnje:
Same here. The icon is cropped.

BloodRaven0:
I'd also love the old icon back.

dusansimic:
Seems like this https://github.com/electron-userland/electron-builder/issues/2128#issuecomment-411797743 is a valid solution. Just explicitly specify the icon. I've tried this out and seems like it works. I'll write a pr for this.

maxoku:
Will that be fixed in the next update? When we can expect it?

dusansimic:
The pull request was merged 6 minutes ago. The fix will be available in the next update however we're not sure when it will roll out.

maxoku:
Thank you for the answer.

sindresorhus:
https://github.com/sindresorhus/caprine/releases/tag/v2.51.2

maxoku:
Thank you for the update, it's fixed now.

PR #1455, opened 2020-11-28
Fix Windows icon

dusansimic:
Add fix for Windows app icon. This fixes the icon being cut in half and bad resolution of the icon. The solution for this issue was found here https://github.com/electron-userland/electron-builder/issues/2128.

Closes: #1429 
Closes: #1453

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,8 @@
 			"publish": "github"
 		},
 		"win": {
-			"verifyUpdateCodeSignature": false
+			"verifyUpdateCodeSignature": false,
+			"icon": "build/icon.png"
 		}
 	},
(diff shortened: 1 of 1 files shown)


CvX (approved):

sindresorhus:
Can you either link to a electron-builder issue here in the PR comments or open one about this?

dusansimic:
Here is the electron-builder issue where if found the solution https://github.com/electron-userland/electron-builder/issues/2128. It's closed due to inactivity however electron-builder still can't handle adding Windows icon properly when it's not explicitly specified. I can open a new issue regarding this if you think that's a better solution?

PR #1468, opened 2020-12-11
Dark theme for new Messenger UI

dexter93:
Why yes, it is a hack. YMMV

css checks will probably fail because of formatting, but it works


waryor88 (approved):
Tested it, works pretty fine for now.

SpacingBat3:
For me, this doesn't work with the icons – most of them are still black (and mixing that with the dark theme makes them barely visible).

dusansimic:
This breaks dark theme on Messenger with old design so there should either be some logic that detects weather the UI is new or old and depending on that enables the new dark mode or old dark mode or we will push this back until we have a better solution for this.

Also in order to provide consistent dark mode, it would be good if we can get some screenshots of that implementation of dark mode on the new Messenger design and keep the colors same as in old design.

Lastly, please fix linting errors. We want to have consistent code across the whole codebase and that's why we use linters.

dexter93:
facebook seems to be still working on their new ui, so at the moment there is no point to clean this up and commit properly.

this is here mostly for testing grounds and a temp solution for people willing to get their hands a bit dirty (me included), hence the title. it's a temp hack, not a fix.

I will however consider doing it cleanly once fb sorts their side.

dusansimic:
Hi again! Thanks for the dedication and the work you've been doing on this issue but here is a thought. I've just got the new Messenger UI and I've tried out adding `__fb-dark-mode` class name to html tag. It seems to work fine with a few bugs (black icons among other things). Why don't we just enable dark mode by adding that class name and adding a view tweaks to the CSS. That would result in much less work and much less code to maintain which is ideal to us since we have already a lot of features.

Let me know what are your thoughts on this.

dexter93:
> Hi again! Thanks for the dedication and the work you've been doing on this issue but here is a thought. I've just got the new Messenger UI and I've tried out adding `__fb-dark-mode` class name to html tag. It seems to work fine with a few bugs (black icons among other things). Why don't we just enable dark mode by adding that class name and adding a view tweaks to the CSS. That would result in much less work and much less code to maintain which is ideal to us since we have already a lot of features.
> 
> Let me know what are your thoughts on this.

Hello! That would be the ideal solution, since this patchset is basically using the official theme. I'm still going through Caprine's code and not sure where exactly that class name should go to enable it to switch. I'll probably need some pointers.

In other news, in today's commits those minor bugs are fixed.  At least on my end, the dark theme is pretty good

dusansimic:
> I'm still going through Caprine's code and not sure where exactly that class name should go to enable it to switch.

That class name should be added to html tag (root tag) and Messenger will enable the dark mode. You can try it out by opening Caprine in light mode, opening dev tools and adding the class name (`__fb-dark-mode`) to the html element.

The html tag sould look something like this:
```html
<html id="facebook" class="_9dls os-linux __fb-dark-mode is-window-inactive" style="--header-height:0px; background-color: rgb(30, 30, 30);" lang="en" dir="ltr">
```

This turn on the already existing dark mode in Messenger and does most of the work. We will just need minor tweaks regarding some icons, we won't need to copy any CSS from Facebook.

The class name can be added when menu option for dark mode is enabled. This is handled in `setDarkMode` function in `browser.ts`.

Additional styles for dark mode will be added in `dark-mode.css` inside `new-design` dir.

Jadex14rthenn:
Wet

JGabT:
yes it fucking works

dusansimic:
Duplicate of #1500

PR #1489, opened 2020-12-17
Add switch logic for Messenger UI

dusansimic:
Since Facebook is rolling out new Messenger design, some users have it and some user don't. This adds logic for dynamically switching css that is injected based on users Messenger design.

This is being added as a PR so contributors could use this for implementing broken features on the new design.


dexter93:
@dusansimic is this ready to go? I could rebase the dark theme on it

however, if people start commiting pr's with this, it will be eventually conflicting everywhere. Please consider merging it.

dusansimic:
@dexter93 since this is merged, we'll be using it from now on in fixing features. It might be reworked or modified a bit in the future (the switch logic) in order to handle both UIs on all features shortcuts for example, not just injected CSS.

PR #1492, opened 2020-12-17
Private mode for new Messenger UI

dexter93:
First attempt on this. Currently just blurring everything

Any feedback welcome

### DIFF
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -1,0 +1,41 @@
+/* -- Private Mode -- */
+/* Right sidebar: big user picture */
+html.private-mode .a8c37x1j.d2edcug0.sn7ne77z.bixrwtb6 {
+	filter: blur(5px);
+}
+/* Right sidebar: user name */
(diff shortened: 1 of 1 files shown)


dusansimic:
This blurs whole message view including message input and all of the buttons. That's also the case for contact list. It would be better if it would blur just the profile pictures, sent images and videos and message bubbles instead of every element on the page.

sindresorhus:
@dexter93 Bump

dusansimic:
@sindresorhus is the goal of private mode to blur everything except messages (what it does on old UI IIRC) or everything including messages (what this does). I'm already working on something so I'd like to know how to proceed.

Also I've been overwhelmed by work on my University so that's why I've been slow with fixing issues.

sindresorhus:
> is the goal of private mode to blur everything except messages (what it does on old UI IIRC) or everything including messages (what this does). I'm already working on something so I'd like to know how to proceed.

Make chat's anonymous for anyone looking, meaning blurring avatars and names.

PR #1543, opened 2021-01-31
Add Adwaita-style CSDs (Client-Side Decorations)

chaitunya:
This is a work-in-progress for adding CSDs to Caprine. The CSDs I added should work on Windows (although I haven't tested it), but the close button uses the Adwaita icon theme, so these CSDs are meant for Linux.

GNOME CSD Initiative: https://blogs.gnome.org/tbernard/2018/01/26/csd-initiative/

Advantages:
* Uses less space
* Fits the GNOME style
* Useful for users of electron-ozone since they might not have title bars.

Current Issues:
* Cannot resize from draggable region
* ~Close button does not work for messages to yourself~ Fixed
* Menu bar doesn't show
* ~No configuration option~ Done
* ~Doesn't change colors with dark mode~ Fixed
* ~Draggable region can be selected~ Fixed

Without CSDs:
nocsd

With CSDs:
csd

### DIFF
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -39,3 +39,7 @@ html.hide-dropdowns [role=menu].l9j0dhe7.swg4t2nn {
 }
 
 body::-webkit-scrollbar { width: 0px; }
+
+.nhd2j8a9 {
+	cursor: default !important;
+}
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -7,6 +7,7 @@ html.hide-dropdowns [role=menu].l9j0dhe7.swg4t2nn {
 .rq0escxv.l9j0dhe7.du4w35lb.j83agx80.pfnyh3mw.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.jei6r52m.wkznzc2l.n851cfcs.dhix69tm.ahb00how,
 .rq0escxv.l9j0dhe7.du4w35lb.j83agx80.pfnyh3mw.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.hv4rvrfc.dati1w0a.f10w8fjw.pybr56ya.b5q2rw42.lq239pai.mysgfdmx.hddg9phg {
 	-webkit-app-region: drag !important;
+	-webkit-user-select: none;
 }
 .rq0escxv.l9j0dhe7.du4w35lb.j83agx80.pfnyh3mw.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.jei6r52m.wkznzc2l.n851cfcs.dhix69tm.ahb00how {
 	margin: 0 !important;
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -23,3 +23,18 @@ html.hide-dropdowns [role=menu].l9j0dhe7.swg4t2nn {
 		padding-top: 36px !important;
 	}
 }
+
+/* Disable drag for info and call buttons on the top right corner */
+.rq0escxv.l9j0dhe7.du4w35lb.j83agx80.cbu4d94t.pfnyh3mw.d2edcug0.hpfvmrgz.p8fzw8mz.pcp91wgn.iuny7tx3.ipjc6fyt {
+	-webkit-app-region: no-drag !important;
+}
+/* Disable drag for status (below chat name) */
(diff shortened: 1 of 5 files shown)


sindresorhus:
I don't see how this could work on Windows as the menu items are integrated in the window frame.

chaitunya:
Do the menu bars work the same way in Linux? If so, do you know if we can try to use the menu options in the same way that Spotify does? The app has a button with a dropdown with the menu items instead of a regular menu bar.

image

dusansimic:
TLDR for maintainers;
Good work regarding the CSD adoption on Linux. Bad idea to do it like this. This should be abstracted into a module. I'd like to keep this open in order to track the progress and steer this in the right direction in order to have this implemented well and maybe have it as a feature.

#

Hi @chaitunya! First of all, I'd like to thank you for your contribution so far. I'm a Gnome user myself and I'd love to see more apps adapt the CSD design since it makes the desktop look more streamlined and personally for me it look much nicer than the title bar 😁.

Now I'll have to drop the big hammer since this is a quite large undertaking regarding that there is basically no standardized implementation for Gnome style (Adwaita) or any kind of Linux desktop CSD and this is mostly going to be written from scratch.

Caprine is currently filled with features that are largely broken due to new Messenger website redesign. I really appreciate the work you're doing and the time you've dedicated to this (don't think I don't), however this will right now just put strain on maintainers and contributors (of which there are very little). As I've said, I really want to see this happen but for the time being, I don't think it's possible until we clear out the mess caused by Messenger redesign.

It'd be very nice if you continue to work on this feature in this PR so I would be able to track it and see if there is any possibility of merging this into the project or this would only live as a patch set for adding CSD to Caprine.

Here are some thoughts about the work you've done so far.

The design looks nice, there are still some effects missing (button background on hover and color change on button press) but I guess that would come in after the behavior is handled well. The missing feature seems to be support for maximize and minimize buttons. They're supporter by Gnome but they're optional so we would need to check what is the configuration in gsettings. There is an npm package that can do that. Also that would mean to check if the buttons should be placed on the left corner of the window or on the right corner of the window. As you've already mentioned, resizing from draggable region is not working the menu bar is easily fixed since we create a menu bar from the Electron Menu class which can also be shown as a context menu (like in the case of Spotify).

Since this is already additional 90 lines of code to an already relatively large codebase, I don't think it would be a good idea to implement it this way. We would be responsible for maintaining a ton of code which might break any time since it's all experimental. I'd like to recommend another approach from which anyone who want's to add Gnome CSD to their app would benefit.

We could make a Node.js module that would add CSD to the Electron app while app developer would only specify some parameters (colors of the buttons, menu to be displayed, callback functions for button clicks, etc.). This would make it easy for developers since they would just care about what the buttons are doing (closing the app, minimizing/maximizing it...) and they would specify where the buttons should be added. That means less hacking together an UI that would work and more focus on other larger problems while the design and implementation of CSD would be done by the module developers. This would for now focus only on window buttons on Electron apps but there will always be more room to grow. I'd say that that is a far better approach to this issue.

As that idea is mostly for the CSD initiative, this should be directed to them. I'd like to help with that in any way I can so send me an email if you're interested. The discussion about the module should not be done here any more than it already is.

I'd like to note also that using Adwaita design of window buttons on Windows is not good practice. For windows, there are already some implementations of CSD and those implementations should be used instead of this one. We don't want to users on macOS/Linux to have Windows style window buttons so why would we want it to be the other way around.

chaitunya:
Thanks for the information :)

I probably should have said this before, but I don't really have any prior experience with Electron/TypeScript; I just wanted to make the app fit the GNOME desktop.

However, I'll be glad to learn some more and work on this project. This week, I'll research more into what you wrote about.

chaitunya:
@dusansimic I've been thinking about the Node.js module, and decided on a model that would be like this:

The app developer creates two empty divs with the class `window-controls`, one for the window controls on the left side and one on the right side, with the IDs `window-controls-left` and `window-controls-right`, respectively. The app developer then runs the function that initializes the window controls, providing callbacks, and other arguments. When the module initializes the window-controls, it'll add the buttons, check the layout, and detect any changes so it'll automatically update. I'll write minimal CSS for the window-controls. The app developer will be able to change the colors, hover animation, and size with their own CSS.

To access GSettings, I'll use [this package](https://www.npmjs.com/package/node-gsettings-wrapper). It hasn't been updated in a couple of years, but it still works from the basic testing that I've done.

chaitunya:
@dusansimic I added the main functionality for the module. Here's the repo for it: https://github.com/chaitunya/electron-gtk-csd

Here is a simple demo app:
Demo screencast

I'll integrate the module with Caprine tomorrow.

lastthyme (approved):

Issue #1536, opened 2021-01-26
Show Unread Badge not working on new design


nguyenngocduy20:
Preference > Show Unread Badge is not working on Caprine Version 2.52.0 (MacOS Big Sur 11.1)

dotpointer:
Is the Unread Badge the same as or maybe related to the icon in the system tray? Because it has stopped working too, not showing any new messages, as mentioned in another issue.

nguyenngocduy20:
It does not work on both menu bar & dock icon.

dusansimic:
Hi! This issue is due to the redesign of the messenger website. It's being worked on however any help is very welcome!

Issue #1028, opened 2019-08-07
Systray icon doesn't show new messages anymore


shagath:
It just stays grey. I love how the app can be hidden now to systray(was waiting for that feature a lot) but no indicators for a new message. It used to go blue iirc.

---

Caprine 2.34.0
Electron 5.0.6
win32 10.0.18945
Locale: en-US

riverstore:
I have same problem from version 2.48.0. And sometime, when I use Caprine to chat with my friend, I seen the message but on his side, Messenger don't indicate that I read message.

Anyone know the problem please help. Thanks a lot.
---

Caprine 2.49.0
Electron 10.1.1
win32 10.0.18362
Locale: en-US

dusansimic:
I think that the original issue is probably fixed by this time since it works on my machine but regarding your problem @riverstore, it could be that you have some unread message down in contact list. If that is not the case please let me know. If you know some specific way that I could reproduce this issue, that would be great.

riverstore:
I'm using the latest version, but the problem still not fixed. And I may familiar with it ^^

dotpointer:
The system tray icon does not change when a new message is available here either, it is always grey. But this is on Debian 10 MATE.

Caprine 2.51.2
Electron 10.1.5
linux 4.19.0-13-amd64
Locale: sv

dusansimic:
Facebook released a new Messenger UI for web which broke most of the Caprine features. Fixes are in progress for all issues including this one. You can see what is fixed so far [here](https://github.com/sindresorhus/caprine/pull/1514). Fixes will be available once that PR is merged.

dotpointer:
While waiting for this to be fixed I found a workaround for Linux desktops by displaying relevant part of the Caprine main window title on a panel on the desktop.

In case a new message appear, then a part of the Caprine window title will be shown on the panel, like:
"(<messages>) Messenger" or "<name> has sent a message"

Script to extract data from the window title, `xwininfo` is required for this to work:
```
#!/bin/bash
WNDTITLE=$(xwininfo -root -children|grep '"caprine" "Caprine"'|head -n1);
WNDTITLE=$(echo $WNDTITLE|sed 's/.*[0-9]x[0-9a-z]* "\(.*\)": .*/\1/');
if [ "$WNDTITLE" != "Messenger" ] && [ "$WNDTITLE" != "caprine" ]; then
  echo "$WNDTITLE";
else
  echo " ";
fi
```

[...]

PR #1557, opened 2021-02-14
Fix conversation list on new UI

dusansimic:
This would fix tray indicator for new messages, however there is one problem. The problem with excluding muted conversations from unread conversations is that once you get a new message in muted conversation, you lose the muted icon. The only way the to check if the conversation is muted is to open muted is to open conversation menu and check if there is an unmute button. The problem then arises when the user has already a different menu open which will result in closing of that menu. Because of that we cannot check if conversation is muted without causing unexpected behavior. Any thoughts on this issue are very welcome!

Closes: #1536
Closes: #1028

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -383,11 +383,11 @@ function setDarkMode(): void {
 	updateVibrancy();
 }
 
-function setPrivateMode(): void {
+function setPrivateMode(isNewDesign: boolean): void {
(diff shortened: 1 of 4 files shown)


sindresorhus:

sindresorhus:
I don't have any good suggestions for the above issue...

Issue #1544, opened 2021-02-01
Shortcuts not working on non US locale


goorkamateusz:

dusansimic:
Please update to the latest version of Caprine (v2.52).

goorkamateusz:
I download caprine installer yesterday, so I've the lastes version...

On console I have this report:
![image](https://user-images.githubusercontent.com/52463898/106457643-8f2b1b80-648f-11eb-99be-42f3e1cb2040.png)

dusansimic:
Okay, that's weird 😅. Could you please check these things:

1. Are you using shortcuts `Ctrl`+`+` and `Ctrl`+`-` for increasing and decreasing text size?
2. Are the menu options for increasing/decreasing text size working (View -> Increase/Decrease Text Size)?
3. Is there an error shown whenever you click those shortcuts or is there no error?
4. Do you have the Caprine window in focus?

I think that the error you see right now in the console is unrelated to this issue. It is referring to a Facebook url which means it's some issue regarding Facebooks code.

goorkamateusz:
1. Not working;
2. Not working;
3. No reactions on shortcuts, and increasing in the menu;
4. Yes, I have.

I noticed that any options in `View` menu not working apart from `Dark mode`. Also:
- from Converation -> Insert GIF to -> Attach file;
- File -> New Conversation;
- and File -> Messenger Settings;

And any keyboard shurtcuts not working apart from some like `Ctrl`+`C` / `V`.

I suppose it's some sdependency problem...

dusansimic:
Hi! The problem seems to be that we're targeting elements in the website using `aria-label` attributes whose values get changed when the locale gets changed. This could be fixed by using class names instead of aria labels. It will take some time to fix things but at least we know what is the solution.

PR #1566, opened 2021-02-25
fix "Create new room" icon color on dark mode

meszarosdezso:
This makes the _Create new room_ icon in the top bar white in dark mode

Before:
Screenshot 2021-02-25 at 22 24 42

After:
Screenshot 2021-02-25 at 22 24 09

### DIFF
--- a/css/new-design/dark-mode.css
+++ b/css/new-design/dark-mode.css
@@ -16,6 +16,12 @@
 	color: var(--primary-text);
 }
 
+/* New room: Icon color */
+[aria-label="Create new room"] .a8c37x1j.ms05siws.hwsy1cff.b7h9ocf4 {
+	fill: currentColor;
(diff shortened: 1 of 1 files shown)


dusansimic:
Hi! The aria label attribute value would change once the Facebook locale changes from en-us to something else. You should use class names instead of aria label attribute then.

meszarosdezso:
All the other rules above are using aria labels for selecting the icon, but I can try to find an _international_ selector for it. It's at least consistent now

dusansimic:
They are, sorry about that. I'll write a pr that will fix it. The main problem is if aria label attribute is used, issues like this #1544 would happen. The solution would be to use attributes that don't rely on locale, mainly `class` attribute.

meszarosdezso:
Cool, let's move with this to #1572. (:

Issue #1587, opened 2021-03-16
Picture displaying does not work on 2.52.1


girardcarillo:
Hi,
When I click on a picture someone sent me, it only displays a white screen, not the picture.
---

Caprine 2.52.1
Electron 10.1.5
linux 4.15.0-136-generic
Locale: fr

jonsolberg:
Same here. 2.52.1 on Windows 10.

MaxiBerdy:
Got the same problem
Caprine 2.52.1
macOS Mojave 10.14.6

not-AMM:
Same

Caprine 2.52.1
Windows 10 20H2

The only messages I could find were: 
- `[Violation] Forced reflow while executing JavaScript took **ms`- 

![imagen](https://user-images.githubusercontent.com/43765089/111534930-25f52400-872e-11eb-950e-c3c61145405e.png)

- `[Violation] 'message' handler took <N>ms`

![imagen](https://user-images.githubusercontent.com/43765089/111535399-aae03d80-872e-11eb-8301-c9a924e3b33d.png)

yardstudio:
Same problems
2.52.1
Win 10 Pro 
Edition	Windows 10 Pro
Version	20H2
Installed on	‎18.‎07.‎2020
OS build	19042.867
Experience	Windows Feature Experience Pack 120.2212.551.0

Issue #1585, opened 2021-03-12
Remove/Delete a message


locseo:
When I chat with my acquaintances. I cannot use the remove/delete utility a message in Caprine though it can work in the browser or app mobile. 
Thanks to the developers.

Issue #1580, opened 2021-03-10
New problems since Caprine 2.52.1


czerwinski76:
All my private messages history is gone.
Only 13 private chats/group discussion are still there.

Transfering pictures ou messages doesn't work anymore.

---

Caprine 2.52.1
Electron 10.1.5
win32 10.0.18363
Locale: fr

dusansimic:
Hi @czerwinski76! I'm not sure what you mean by private messages. Could you clarify what that private messages mean, what is the feature called?

Also I'm not really sure what transferring pictures mean. Do you mean you can't send pictures anymore?

jonsolberg:
+1

I also have problems with displaying images and reacting to messages (thumbs up on a message e.g.). Recently stopped working, but functioned before on 2.52.1.

Issue #1579, opened 2021-03-10
Reaction button doesn't work


Naimad1CZ:
Since today (I reopened Caprine program today after last few days, so it probably caused Caprine to update to newest version 2.52.1), when clicking on button "react" (to a message), nothing happens, no menu with emojis is displayed.

Tried reinstalling Caprine (from https://sindresorhus.com/caprine/), didn't help.

I have always had only the basic reaction menu with 7 emojis (which still appears when I try to react to a message on facebook website).

---

Caprine 2.52.1
Electron 10.1.5
win32 10.0.19042
Locale: en-US

Naimad1CZ:
Tried to install 2.52.0 version and this bug doesn't occur here.

Also noticed that loading older conversations (by scrolling down) doesn't work in 2.52.1 (older conversation appears ONLY in case when you search for some conversation (e.g. for name of someone) and click on it, then this one conversation appears with other few loaded), but works normally in 2.52.0
(should I open another issue? Or rename this issue to "Bugs in 2.52.1"?)

dusansimic:
Hi @Naimad1CZ! This is very unusual since I'm running the latest version (2.52.1) and it doesn't happen to me. I'll check on different platforms and report back here. Until then, let's just leave it as is, after I check we will see if there is any need to open new issues.

czerwinski76:
> on button "react" (to a message), nothing happens, no menu

Same problem on my computer.

MaxiBerdy:
Same problem here since the newest update.

LoricAndre:
Same on EndeavourOS (Arch-based, latest kernel)

girardcarillo:
Same problem with Ubuntu 18.04

jonsolberg:
Same issue. 2.52.1, Windows 10.

fuzzy76:
Was this supposed to be fixed in 2.52.2? Because it isn't.

fuzzy76:
Funny. Opening and closing the preferences windows fixes it. Not a good workaround, but hey... 😀

dusansimic:
Hi @fuzzy76.

> Was this supposed to be fixed in 2.52.2? Because it isn't.

[...]

PR #1589, opened 2021-03-18
Fix reaction, delete and view image bugs

dusansimic:
There are some bugs on latest release that cause inability to send reactions on messages, delete messages and conversation and load images when clicked. This happens because hide-preferences-window class is not removed from document when preferences window is closed because close button couldn't be found on some locales.

It's fixed by looking for the close button with class names instead of aria-label attribute.

Closes: #1579 
Closes: #1580 
Closes: #1585 
Closes: #1587


mcardillo55:
I'm still seeing the reaction & the view image issues on 2.52.3-1 in Arch Linux. Are these confirmed resolved?

PR #1614, opened 2021-04-20
Enable subpixel rendering on Linux

AidanGG:
As described in the [Electron FAQ](https://github.com/electron/electron/blob/master/docs/faq.md#the-font-looks-blurry-what-is-this-and-what-can-i-do) and https://github.com/electron/electron/issues/10955, I tested that this fix restores subpixel rendering on Linux.


sindresorhus:
I don't think this will work with dark mode.

AidanGG:
darkmode_subpixel

Seems to work for me.

AidanGG:
Actually, I notice that it applies subpixel rendering to the main chat window but strangely not to the chat previews on the left side. This occurs in both light and dark modes.

PR #1618, opened 2021-04-23
Hide additional attributes in private mode

mquevill:
- Change titlebar element (both name and picture)
- Read indicators (in group chats, the indicators that aren't the last one are controlled differently)

### DIFF
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -40,7 +40,7 @@ html.private-mode [data-pagelet=page] .d2edcug0.hpfvmrgz.qv66sw1b.c1et5uql.gk29l
 	filter: blur(5px);
 }
 /* Conversation: read indicator */
-html.private-mode [role=main] .j83agx80.aovydwv3.rl25f0pe.iyic8zgw {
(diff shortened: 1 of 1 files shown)


dusansimic:
For me, the name and profile pictures of the chat are not blured in the right sidebar. It would be good to have that blured to in case someone likes to use Caprine with right sidebar open.

mquevill:
@dusansimic Makes sense. I'll blur those too. On the right sidebar, should the "Chat Members" expandable list be blurred too? Or should that be up to the user to keep closed?

Also, I'm working on fixing an issue when chat colors are set to a gradient:
image

dusansimic:
> Or should that be up to the user to keep closed?

I think this will be fine. It's more useful to leave it as is and close it if user want's to hide it.

mquevill:
Summary of what's changed now:
- Added: Hide name label on message (group chat)
- Added: Read indicators in message
- Added: Name, picture, and active status in right sidebar
- Adjusted: Read indicators in message list
- Adjusted: Titlebar name and picture

dusansimic (approved):

dusansimic:
Thanks @mquevill! Facebook likes to change Messenger frontend every now and then and these features tend to break quite often lately 😅

PR #1621, opened 2021-04-26
Hide banner for Messenger App for Windows

mquevill:
This fully eliminates the banner advertising the Messenger App for Windows. Perhaps we may not want to remove it entirely, but if someone is using Caprine, they probably want to stick with it.

Regardless, the banner gets a bit mangled with the narrow sidebar:
image

The wide sidebar shows this correctly: 
image

Another possibility would be to remove the banner when the sidebar is narrow, but keep it when the sidebar is wide?

### DIFF
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -84,8 +84,13 @@ html.private-mode [role=main] .d2edcug0.hpfvmrgz.qv66sw1b.c1et5uql.b0tq1wua.a8c3
 	filter: blur(5px);
 }
 
-/* Hide the "Messenger App for Windows" banner */
+/* Hide the "Messenger App for Windows" banner in chat list */
 .rq0escxv.l9j0dhe7.du4w35lb.j83agx80.g5gj957u.rj1gh0hx.buofh1pr.hpfvmrgz.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.s9xz0pwp.c4m0enpj.rl78xhln.srn514ro.sn0e7ne5.f6rbj1fe.l3ldwz01 {
+	display: none;
+}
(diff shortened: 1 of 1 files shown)


dusansimic:
In my opinion, if I were to use Caprine on Windows, I'd like to have that removed since I'm using Caprine and I don't care about the Windows app. Probably most people already know about the Windows app and if they're sticking with Caprine, they're going to stick with Caprine.

sindresorhus:
Let's remove it completely.

dusansimic:
A friend of mine pointed out to me that there is also a button for Messenger app in the preferences menu. It would be good if that one is removed too.

image

mquevill:
All the menu items are the same class, so I had to use `a:last-of-type`. But the menu item is removed completely.

image

mquevill:
Using `a:last-of-type` would be an issue for non-Windows, since it would be removing the "Cookie Policy" entry. Also, it appears that the bottom separator bar is not present on messenger.com with Firefox on Red Hat 7.9 Linux.

I am now removing `hr:nth-of-type(4)` and `a:nth-of-type(6)`, which should be more precise at selecting the Messenger App for Windows menu item and its selector.

dusansimic:
Another option would be to use the class `os-platform` to check if the os is Windows and if it is, and it's new design, remove the specific element from menu. The class for Windows is `os-win32` I think and the class for new design is `_9dls`.

mquevill:
I added the `.os-win32` platform specification from Electron. I don't think there's a way to avoid using `nth-of-type`, because all of the `hr` separators are exactly the same. The `a:nth-of-type(6)` could be replaced with matching with `[href$="startDownload"]`, but the link may change in different locales.

dusansimic (approved):

Issue #1563, opened 2021-02-18
Refactor Github Actions for better automation


dusansimic:
Currently GitHub Actions is used as a CI/CD service for Caprine. There are some problems with it (every test on PR fails since build process is inside test job and there are no CSC keys for signing packages on forked repo and thus it fails). Also as far as I can tell, CI is not really automating snap deployment. Maintainer still needs to publish the snap manually by running `npm run release`. This could all be automated using Actions without the need for any manual intervention. Also there is an issue reported in #1554. Package hashes don't match the reported hash in yaml files which seems to be fixed by building each package on it's target OS.

I've already started testing out some configurations on my clone [dusansimic/caprine-playground](https://github.com/dusansimic/caprine-playground) and got to a point where everything seems to work fine. I'm using [Electron Builder Action](https://github.com/marketplace/actions/electron-builder-action) alongside [Snapcraft Action](https://github.com/marketplace/actions/snapcraft-action) which together allow me to build Caprine for all platforms and upload snap package to Snap Store. In addition I'm caching npm packages so the Action would be shorter. I haven't yet tested code signing but this seems to have good results since I don't have secrets set in my repo and the build still passes (just skips signing). That would mean PR won't fail anymore when there is no CSC secrets. I've also split the test job into two jobs: test and build. This would make it easier to understand which part of the testing pipeline failed (is it `npm test` or build process and on which platform it failed). I'm using the GitHub release workflow recommended by Electron Builder docs explained [here](https://www.electron.build/configuration/publish#recommended-github-releases-workflow). I don't know what current release process is but this seems to work without any problems for me.

I know this is adding more things to the build process but it automates every part of the deployment process. Maintainer just needs to create a draft release and once everything is ready, publish the release. It also fixes the problems on pull requests. There won't be any need for running `dist` script since it's run by Electron Builder Action and it would minimize the number of scripts in the package.json.

I'd like to hear some thoughts on this and recommendations for improving the process.

// @sindresorhus 
// @CvX

sindresorhus:
Sure. PR welcome 👍🏻

Issue #1554, opened 2021-02-08
latest.yml hash for .exe release does not match on my machine


v4k0nd:
According to the latest.yml the hash of the newest release for Windows (.exe) starts as 9GKiEbubaPc62vYCmCy8cYt... and on my machine it did not match.
I checked two ways on my win 10 (ver. 20H2 build 19042.746):
`powershell> Get-FileHash .\Caprine-Setup-2.52.0.exe -Algorithm SHA512`

`cmd> certutil -hashfile Caprine-Setup-2.52.0.exe SHA512`

and they both resulted in a hash starting as 58a1bcf1f7af889bbfd6222e...
Could somebody else check it? Did I check it in a wrong way?

v4k0nd:
Going back 3 releases it still does not match (it is probably just my machine or my way of checking was wrong then)

dusansimic:
Hi! You're right, the hash is not correct. I've checked hashes for all platforms (Windows, macOS and Linux) and they don't match. I'm not sure what is the problem since those hashes in yaml file are generated by Electron Builder. I'll leave this open until we figure out what is going on here.

v4k0nd:
Ah that's a relief on my end then, I was searching everywhere how could I be getting the wrong hashes. Thank you for the quick response!

v4k0nd:
I think I found something, there is a nearly two year old issue/answer in [electron-builder](https://github.com/electron-userland/electron-builder/issues/3913#issuecomment-504698845).
Under the mentioned issues, I found [this](https://github.com/electron-userland/electron-builder/issues/3119#issuecomment-405097164), where it says that electron-builder wont just calculate the SHA512 and output it in ascii, but after calculating the SHA512, it takes the binary and converts it to base64. I should have noticed that it is not just normal ascii since because the sha512 has a `==` at the end, which indicates padding for the encoding, usually.

However this does not answer me how to check it "manually" (only cmd / powershell / bash commands). 

I did find something for bash in [this issue](https://github.com/th-ch/youtube-music/issues/69#issuecomment-735396905) but it did not calculate the correct hash, nor any combination from the mentioned stackexchange link.

If I manage to find an answer I will try to keep this issue updated.

dusansimic:
[...]

Issue #1163, opened 2019-11-04
Auto-update failing on Mac for version 2.40.0


evanwon:
Currently running Caprine 2.39.0 (2.39.0.1630) on Mac OS 10.14.6 Mojave.

Caprine does not automatically update to version 2.40.0. The error from the logs indicates checksum mismatch:

```
[2019-11-04 12:00:21.165] [info] Checking for update
[2019-11-04 12:00:22.481] [info] Found version 2.40.0 (url: Caprine-2.40.0-mac.zip, Caprine-2.40.0.dmg)
[2019-11-04 12:00:22.484] [info] Downloading update from Caprine-2.40.0-mac.zip, Caprine-2.40.0.dmg
[2019-11-04 12:00:22.494] [info] No cached update info available
[2019-11-04 12:00:45.813] [error] Error: Error: sha512 checksum mismatch, expected e37XN4qn+DhJgHUPaFJyEdncAXidLIpiFtWFmPCl59cXDJqliu12MM2+s52IkE4KSlAn5gqYcUlMb40i0Vjcbw==, got HN+PplmEXKKY9ngSZRbmdZhAyxp2xjLK65Wx/Vr9COIIHi+ewQNLhW64lHVRDvZJDdskJ6PXaFeKaETe7/0KBA==
    at newError (/Applications/Caprine.app/Contents/Resources/app.asar/node_modules/builder-util-runtime/out/index.js:212:17)
    at DigestTransform.validate (/Applications/Caprine.app/Contents/Resources/app.asar/node_modules/builder-util-runtime/out/httpExecutor.js:402:35)
    at DigestTransform._flush (/Applications/Caprine.app/Contents/Resources/app.asar/node_modules/builder-util-runtime/out/httpExecutor.js:386:14)
    at DigestTransform.prefinish (_stream_transform.js:140:10)
    at DigestTransform.emit (events.js:200:13)
    at prefinish (_stream_writable.js:633:14)
    at finishMaybe (_stream_writable.js:641:5)
    at endWritable (_stream_writable.js:661:3)
    at DigestTransform.Writable.end (_stream_writable.js:592:5)
    at IncomingMessage.onend (_stream_readable.js:675:10)
```

I'll update manually, but wanted to give you a heads up that the [latest-mac.yml](https://github.com/sindresorhus/caprine/releases/download/v2.40.0/latest-mac.yml) checksum may be incorrect for 2.40.0.

Issue #1059, opened 2019-08-21
Auto-update issue on Mac OS X with v2.37.0


evanwon:
I noticed that Carpine was not auto-updating on my computer (macOS Mojave, 10.14.6). I checked out the logs today, which contained the following:

```
2019-08-20 16:26:19.694] [info] Checking for update
[2019-08-20 16:26:20.710] [error] Error: Error: Cannot find latest-mac.yml in the latest release artifacts (https://github.com/sindresorhus/caprine/releases/download/v2.37.0/latest-mac.yml): HttpError: 404 Not Found
"method: GET url: https://github.com/sindresorhus/caprine/releases/download/v2.37.0/latest-mac.yml\n\nPlease double check that your authentication token is correct. Due to security reasons actual status maybe not reported, but 404.\n"
```

When I look at the v2.37.0 release, I see that it doesn't contain `latest-mac.yml` as is mentioned in the error message.

CvX:
Yeah, the build process is currently broken (see: https://github.com/electron-userland/electron-builder/issues/4134). I've reverted that release to the "draft" status.

@sindresorhus could you build and publish locally in the meantime?

dusansimic:
We've switched to GitHub Actions and there seem to be no issues with auto updating.

PR #1627, opened 2021-05-03
Refactor GitHub Actions

dusansimic:
So far main workflow would test and build Caprine for Linux, macOS and Windows on macOS platform which would result in bad checksums. Also tests sometimes break for no reason at all.

The main difference here is that build and test jobs are split and build process is run on respecting platform for each package. This doesn't break the checksum. This can be seen in action in [playground](https://github.com/dusansimic/caprine-playground) repo.

Aditionally, everything is published automatically, including the snap package.

Closes: #1563 
Closes: #1554 
Closes: #1163

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,13 +7,31 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
         with:
           node-version: 14
-      - run: npm install
+      - uses: c-hive/gha-npm-cache@v1
+      - run: npm ci
       - run: npm test
-      - run: npm run dist
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


mquevill:
If this fixes the checksum generation for MacOS and Windows, it should also close #1059, #1163, #1554.

dusansimic:
Correct, I forgot about those 😅. I'm not sure if the first one you mentioned is the same issue so I'd rather leave it open. Maybe it's fixed by now so it could be closed but it's unrelated to the checksum issue.

michaelarnauts:
Reading this PR, it seems that the snap package should also be published to snapcraft? That doesn't seem to be the case: https://snapcraft.io/caprine

The latest snap there is 2.52.4.

dusansimic:
@michaelarnauts thank you very much for reporting this! I seem to have missed to mention that it's needed to add `snapcraft_token` to Action secrets which I had in my test repo. This is used to upload the snap package to snapcraft.

There was a message in logs that I didn't notice.
image

// @sindresorhus

sindresorhus:
@dusansimic Where do I find the Snapcraft token?

dusansimic:
@sindresorhus

First you would need to download snapcraft. https://snapcraft.io/docs/installing-snapcraft#heading--macos

After that you will need to log in into snap store with the cli app you've installed and acquire the login token. https://github.com/samuelmeuli/action-snapcraft#log-in

The long base64 string (exported login) is the token.

sindresorhus:
```
❯ snap login

Interacting with snapd is not yet supported on darwin.
This command has been left available for documentation purposes only.
```

🤷‍♂️

So it doesn't look like I'll be able to handle this until I get home where I have a Linux VM, which will be a while.

dusansimic:
@sindresorhus, a friendly ping. Some snap users are reporting issues due to lack of updates.

mquevill:
Another issue: #1766

sindresorhus:
Ok. I have added the secret now. GitHub forced it to be uppercase though. Not sure if that matters. `SNAPCRAFT_TOKEN`.

PR #1629, opened 2021-05-04
Apply dark mode to all new nodes

mquevill:
Even if dark mode was set, new nodes (dialogs, etc.) would appear with the `__fb-light-mode` tag, which breaks the theming. I've set up an observer for added nodes and only change ones that have some sort of light mode tag.

This may not be the best or most efficient way to do this, but it seems to apply to most dialogs that I was able to find: Preferences dropdown and overlay, emoji picker, etc.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -409,15 +409,15 @@ async function observeDarkMode(): Promise<void> {
 	/* Added nodes (dialogs, etc.) */
 	const observerNew = new MutationObserver((records: MutationRecord[]) => {
 		const nodeRecords = records.filter(record => record.addedNodes.length > 0);
-		nodeRecords.forEach(nodeRecord => {
-			nodeRecord.addedNodes.forEach(newNode => {
+		for (const nodeRecord of nodeRecords) {
+			for (const newNode of nodeRecord.addedNodes) {
 				const {classList} = (newNode as HTMLElement);
 				const isLight = classList.contains('light-mode') || classList.contains('__fb-light-mode');
 				if (api.nativeTheme.shouldUseDarkColors === isLight) {
 					setDarkModeElement(newNode as HTMLElement);
 				}
-			});
-		});
+			}
+		}
 	});
 
 	observerNew.observe(document, {subtree: true, childList: true, attributes: true, attributeFilter: ['addedNodes']});
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -420,7 +420,16 @@ async function observeDarkMode(): Promise<void> {
 		}
 	});
 
-	observerNew.observe(document, {subtree: true, childList: true, attributes: true, attributeFilter: ['addedNodes']});
+	/* Observe only elements where new nodes may need dark mode */
+	const menuElements = await elementReady<HTMLElement>('html > body > div:nth-of-type(1) > div > div:nth-of-type(1) > div > div:nth-of-type(3) > div', {stopOnDomReady: false});
+	if (menuElements) {
+		observerNew.observe(menuElements, {childList: true});
+	}
+
+	const modalElements = await elementReady<HTMLElement>('html > body > div:nth-of-type(1) > div > div:nth-of-type(1) > div > div:nth-of-type(2) > div > div > div:nth-of-type(2) > div', {stopOnDomReady: false});
+	if (modalElements) {
+		observerNew.observe(modalElements, {childList: true});
+	}
 }
 
 function setPrivateMode(isNewDesign: boolean): void {
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -371,27 +371,33 @@ ipc.answerMain('reload', () => {
 });
 
 function setDarkMode(): void {
+	setDarkModeElement(document.documentElement);
+	updateVibrancy();
+}
+
+function setDarkModeElement(element: HTMLElement): void {
 	if (is.macos && config.get('followSystemAppearance')) {
 		api.nativeTheme.themeSource = 'system';
 	} else {
 		api.nativeTheme.themeSource = config.get('darkMode') ? 'dark' : 'light';
 	}
 
-	document.documentElement.classList.toggle('dark-mode', api.nativeTheme.shouldUseDarkColors);
-	document.documentElement.classList.toggle('__fb-dark-mode', api.nativeTheme.shouldUseDarkColors);
-
-	updateVibrancy();
+	element.classList.toggle('dark-mode', api.nativeTheme.shouldUseDarkColors);
+	element.classList.toggle('light-mode', !api.nativeTheme.shouldUseDarkColors);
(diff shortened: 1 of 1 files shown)


dusansimic:
All popup menus seem to be created in `/html/body/div[1]/div/div[1]/div/div[2]/div/div/div[2]/div` and all modals seem to be created in `/html/body/div[1]/div/div[1]/div/div[3]/div`. It might be better to observe just those two elements instead of the whole document and it's subtree. In addition to that, you wouldn't need to observer the subtree at all since the direct children elements have light mode classes.

I'm not sure how much more efficient would it be but at lest it's better than observing the whole tree right from the document element 😊.

source/browser.ts
```diff
@@ -371,34 +371,56 @@ ipc.answerMain('reload', () => {
...
 		}
 	});
 
 	observer.observe(document.documentElement, {attributes: true, attributeFilter: ['class']});
+
+	/* Added nodes (dialogs, etc.) */
+	const observerNew = new MutationObserver((records: MutationRecord[]) => {
+		const nodeRecords = records.filter(record => record.addedNodes.length > 0);
+		nodeRecords.forEach(nodeRecord => {
+			nodeRecord.addedNodes.forEach(newNode => {
```
dusansimic, line 413: Preferably use `for of` instead of `forEach`.

dusansimic (changes requested):

mquevill:
I definitely agree that observing the entire tree is not ideal, which is why I was trying to limit it somewhat with attributes.

However, those elements do not exist when starting up Caprine, so it throws an error when trying to observe just those elements. Perhaps calling this observer at a later time in startup would get around this issue? Or there may be a more elegant way to address this.

dusansimic:
> However, those elements do not exist when starting up Caprine

You could start observing once the [element is ready](https://github.com/sindresorhus/element-ready). It's already being used in Caprine in similar situations.

> Or there may be a more elegant way to address this.

I was thinking that too but in the end settled on creating `observeDarkMode` function 😅. This might be the best solution for now but since now the Messenger website is more dynamic, it seems that we'll need more and more observers to patch the design where required.

source/browser.ts
```diff
@@ -371,34 +371,56 @@ ipc.answerMain('reload', () => {
...
+				const {classList} = (newNode as HTMLElement);
+				const isLight = classList.contains('light-mode') || classList.contains('__fb-light-mode');
+				if (api.nativeTheme.shouldUseDarkColors === isLight) {
+					setDarkModeElement(newNode as HTMLElement);
+				}
+			}
+		}
+	});
+
+	observerNew.observe(document, {subtree: true, childList: true, attributes: true, attributeFilter: ['addedNodes']});
```
dusansimic, line 423: I'm not sure this attribute filter does anything since `addedNodes` is not an attribute of those elements. By filtering out records you already ignore cases when no new nodes where added.

dusansimic (changes requested):

dusansimic (approved):

dusansimic:
Looks good, thanks ✌🏻

source/browser.ts
```diff
@@ -371,34 +371,65 @@ ipc.answerMain('reload', () => {
...
+				const isLight = classList.contains('light-mode') || classList.contains('__fb-light-mode');
+				if (api.nativeTheme.shouldUseDarkColors === isLight) {
+					setDarkModeElement(newNode as HTMLElement);
+				}
+			}
+		}
+	});
+
+	/* Observe only elements where new nodes may need dark mode */
+	const menuElements = await elementReady<HTMLElement>('div[class="rq0escxv l9j0dhe7 du4w35lb"] > div:nth-of-type(3) > div', {stopOnDomReady: false});
```
dusansimic, line 424: The query selectors are inverted and I'd prefer using dot notation for classes instead of attribute notation. That way if in some case they are rendered with different order it would still work but with attribute notation it won't.

```suggestion
	const modalElements = await elementReady<HTMLElement>('div.rq0escxv.l9j0dhe7.du4w35lb > div:nth-of-type(3) > div', {stopOnDomReady: false});
```

source/browser.ts
```diff
@@ -371,34 +371,65 @@ ipc.answerMain('reload', () => {
...
+	if (menuElements) {
+		observerNew.observe(menuElements, {childList: true});
```
dusansimic, line 426:
```suggestion
	if (modalElements) {
		observerNew.observe(modalElements, {childList: true});
```

source/browser.ts
```diff
@@ -371,34 +371,65 @@ ipc.answerMain('reload', () => {
...
+	}
+
+	const modalElements = await elementReady<HTMLElement>('.j83agx80.cbu4d94t.l9j0dhe7.jgljxmt5.be9z9djy > div:nth-of-type(2) > div', {stopOnDomReady: false});
+	if (modalElements) {
+		observerNew.observe(modalElements, {childList: true});
```
dusansimic, line 431:
```suggestion
	const menuElements = await elementReady<HTMLElement>('.j83agx80.cbu4d94t.l9j0dhe7.jgljxmt5.be9z9djy > div:nth-of-type(2) > div', {stopOnDomReady: false});
	if (menuElements) {
		observerNew.observe(menuElements, {childList: true});
```

dusansimic (changes requested):
Sorry for the nagging but once I've seen your other pr I've realized the issue with query selector and after that the mixed up variable names 😅

source/browser.ts
mquevill: The issue with the dot notation here is that there are many partial matches to `.rq0escxv.l9j0dhe7.du4w35lb`. Even with an exact match, there are still multiple elements that match. If we rely on the child structure `> div:nth-of-type(3) > div`, then there is only one match, regardless of partial/complete match. However, I was wanting to do an exact match, just in case the structure of partial matches changes later.

With the preferences overlay open:
```javascript
> document.querySelectorAll('.rq0escxv.l9j0dhe7.du4w35lb').length
<- 151
> document.querySelectorAll('div.rq0escxv.l9j0dhe7.du4w35lb').length
<- 132
> document.querySelectorAll('div[class="rq0escxv l9j0dhe7 du4w35lb"]').length
<- 3
```

Going with `'div.rq0escxv.l9j0dhe7.du4w35lb > div:nth-of-type(3) > div'` works for now, but is there a way to avoid partial matches with the dot notation?

dusansimic: Yeah, you're right. I don't know any good way to use dot notation and exclude partial matches so just leave it with attribute notation and change the variable name 😊.

mquevill:
> Sorry for the nagging but once I've seen your other pr I've realized the issue with query selector and after that the mixed up variable names

I'd rather things be right before getting merged! I just swapped the query selectors instead of changing the variable names. (Then they stay alphabetical too!)

dusansimic (approved):

PR #1633, opened 2021-05-06
Use full path to check that CSS file exists

mquevill:
The change in #1630 did not check for the *full* path. This was causing none of the CSS files to be read, since they don't exist in the `source/` directory.

Also, there was not a check for the `workchat.css` case. Now all of the `insertCSS` blocks have the same structure.


dusansimic (approved):
Looks good. Sorry for jazz reviews lately. I've had a lot of my plate and didn't really focus when reviewing code 😅.

PR #1637, opened 2021-05-10
Combine system appearance and dark mode settings

mquevill:
The `api.nativeTheme` should work for non-MacOS platforms. This PR combines the `followSystemAppearance` and `darkMode` settings into a `'theme'` configuration setting, which mirrors the `api.nativeTheme.themeSource` variable. However, setting the `'dark'` or `'light'` mode should override the system settings.

I'm not sure if removing the `darkTheme`, which is supposed to force dark mode with _some_ GTK+3 environments, is okay. It could cause some issues with those environments that rely on this configuration.

The theme only updates when you click on the window, but I don't think it should be constantly checking while Caprine is in the background.

Tested on:
- Windows 10, build 19042
- Linux Mint 20.1, Cinnamon DE

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -370,20 +370,20 @@ ipc.answerMain('reload', () => {
 	location.reload();
 });
 
-function setThemeMode(): void {
+function setTheme(): void {
 	api.nativeTheme.themeSource = config.get('theme');
-	setThemeModeElement(document.documentElement);
+	setThemeElement(document.documentElement);
 	updateVibrancy();
 }
 
-function setThemeModeElement(element: HTMLElement): void {
+function setThemeElement(element: HTMLElement): void {
 	element.classList.toggle('dark-mode', api.nativeTheme.shouldUseDarkColors);
 	element.classList.toggle('light-mode', !api.nativeTheme.shouldUseDarkColors);
 	element.classList.toggle('__fb-dark-mode', api.nativeTheme.shouldUseDarkColors);
 	element.classList.toggle('__fb-light-mode', !api.nativeTheme.shouldUseDarkColors);
 }
 
-async function observeThemeMode(): Promise<void> {
+async function observeTheme(): Promise<void> {
 	/* Main document's class list */
 	const observer = new MutationObserver((records: MutationRecord[]) => {
 		// Find records that had class attribute changed
@@ -395,7 +395,7 @@ async function observeThemeMode(): Promise<void> {
 		});
 		// If config and class list don't match, update class list
 		if (api.nativeTheme.shouldUseDarkColors !== isDark) {
-			setThemeMode();
+			setTheme();
 		}
 	});
 
@@ -409,7 +409,7 @@ async function observeThemeMode(): Promise<void> {
 				const {classList} = (newNode as HTMLElement);
 				const isLight = classList.contains('light-mode') || classList.contains('__fb-light-mode');
 				if (api.nativeTheme.shouldUseDarkColors === isLight) {
-					setThemeModeElement(newNode as HTMLElement);
+					setThemeElement(newNode as HTMLElement);
 				}
 			}
 		}
@@ -509,7 +509,7 @@ ipc.answerMain('update-sidebar', () => {
 	updateSidebar();
 });
 
-ipc.answerMain('set-theme-mode', setThemeMode);
+ipc.answerMain('set-theme', setTheme);
 
 ipc.answerMain('set-private-mode', setPrivateMode);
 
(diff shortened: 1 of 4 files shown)


source/menu.ts
```diff
@@ -88,6 +88,42 @@ export default async function updateMenu({isNewDesign}: INewDesign): Promise<Men
...
+			checked: config.get('theme') === 'system',
+			async click() {
+				config.set('theme', 'system');
+				sendAction('set-theme-mode');
+				await updateMenu({isNewDesign});
+			}
+		},
+		{
+			type: 'separator'
+		},
```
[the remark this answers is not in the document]
mquevill: I was mirroring the emoji selection menu (Native vs FB/Messenger options), but since there are fewer options, I'll just remove the separator.

source/browser.ts
```diff
@@ -370,25 +370,20 @@ ipc.answerMain('reload', () => {
 	location.reload();
 });
 
-function setDarkMode(): void {
-	setDarkModeElement(document.documentElement);
+function setThemeMode(): void {
```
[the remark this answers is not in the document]
mquevill: I think "color scheme" sounds like the user is actually choosing colors, whereas a "theme" is a guideline of dark/light and the corresponding colors are already chosen for the user. [This might just be up to each person's connotations of the words.] Using "theme" leaves room for "color scheme" options (e.g. choosing a "red" theme with dark or light mode) to be added in the future. I had debated between `setTheme` and `setThemeMode`, because I feel that `'system'` isn't _really_ a theme. Though I suppose it is just a placeholder for whatever the system has set.

source/index.ts
```diff
@@ -294,7 +293,6 @@ function createMainWindow(): BrowserWindow {
 		alwaysOnTop: config.get('alwaysOnTop'),
 		titleBarStyle: 'hiddenInset',
 		autoHideMenuBar: config.get('autoHideMenuBar'),
-		darkTheme: isDarkMode, // GTK+3
```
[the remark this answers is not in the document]
mquevill: If it is undefined, Electron will set it as false (since it's only a flag to _force_ dark mode, it doesn't force light mode). So my recent commit simplifies the logic.

source/index.ts
```diff
@@ -274,6 +274,7 @@ function setNotificationsMute(status: boolean): void {
 
 function createMainWindow(): BrowserWindow {
 	const lastWindowState = config.get('lastWindowState');
+	const isDarkMode = config.get('theme') === 'system' ? undefined : config.get('theme') === 'dark';
```
dusansimic: I think this behavior could be achieved by creating an object instead that would be spread in the window options object. The object would be empty in case `theme` is `system` and it would have `darkTheme` property in case it's either `dark` or `light` (`true` and `false` respectively). That would specify dark theme property in case it's needed and it wouldn't add it at all in case the theme is system. I'd really like if that could be done since it would make it consistent with the system theme at all times (it would need a restart though).

dusansimic (changes requested):

source/index.ts
mquevill: It does not seem like there's a good way to change this after the BrowserWindow has already been created. Perhaps it would only require a restart on platforms that use GTK+ 3? Also, I don't think there's a way to _force_ light mode, if GTK would normally use dark mode for the menubar, since `darkTheme: undefined` is the same as `darkTheme: false`, according to the following lines.

https://github.com/electron/electron/blob/9cf71d72abf97c7583b200063915aa9c8fb23e15/shell/browser/native_window_views.cc#L223-L234
```cc
#if defined(USE_X11)
  if (!features::IsUsingOzonePlatform()) {
    // Start monitoring window states.
    window_state_watcher_ = std::make_unique<WindowStateWatcher>(this);

    // Set _GTK_THEME_VARIANT to dark if we have "dark-theme" option set.
    bool use_dark_theme = false;
    if (options.Get(options::kDarkTheme, &use_dark_theme) && use_dark_theme) {
      SetGTKDarkThemeEnabled(use_dark_theme);
    }
  }
#endif

dusansimic: I'm not really sure how but I've managed with the first commit in this pr to have the title bar color change once theme is changed from dark to light or the system theme is changed and now I can't get it to work. I'm have no idea what's going on so let's just leave it without the `darkTheme` option (just as it was at the beginning of the pr). Everything else lgtm.

dusansimic (approved):

source/index.ts
mquevill: Okay, sounds good. Removed in 0dac6db.

PR #1641, opened 2021-05-21
Fix: Cannot read property 'classList' of null

J2TEAM:
Screenshot:
image


dusansimic:
With this if statement, the class will be added/removed only if html tag exists. Our goal is to add/remove the class once html tag gets created. A better approach would be using [`elementReady`](https://github.com/sindresorhus/element-ready) function which would wait for element to be created and then add/remove the class.

sindresorhus:
@J2TEAM Bump :)

PR #1642, opened 2021-05-21
Fix: Cannot read property 'children' of null

J2TEAM:
Screenshot:
image

image

### DIFF
--- a/source/browser/conversation-list.ts
+++ b/source/browser/conversation-list.ts
@@ -134,6 +134,10 @@ async function createConversation(element: HTMLElement): Promise<Conversation> {
 }
 
 async function getLabel(element: HTMLElement): Promise<string> {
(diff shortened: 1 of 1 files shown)


dusansimic:
Is there a way to reproduce this error? I'm not experiencing it on my machine (Caprine v2.53.0).

mquevill:
The error only comes up when the conversation list is narrow (only shows pictures, no text). Simply hovering over a conversation will raise this error. It does not appear with the wide sidebar.

dusansimic:
I guess that querying for the correct list (only the one that is full size, not compact) would be a better solution. There might event be chat name information in compact chat that needs to be extracted in a different way. I'll research this a bit.

sindresorhus:
Bump :)

PR #1643, opened 2021-05-21
Fix: Cannot set property 'checked' of null

J2TEAM:
Screenshot:

image

New selector based on attribute (`name`):

image

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -853,7 +853,12 @@ window.addEventListener('dblclick', (event: Event) => {
 
 window.addEventListener('load', () => {
 	if (location.pathname.startsWith('/login')) {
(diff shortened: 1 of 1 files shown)


dusansimic:
Hi @J2TEAM! Thanks for submitting a pull request for this issue!

I'd just like to point out that this issue is occurring because the design of Messenger website has changed and the check box is not the same element it used to be. A better solution would be to look for the element you specified (`[name="persistent"]`) only if the new design is present.

We're using the function [isNewDesing](https://github.com/sindresorhus/caprine/blob/main/source/browser.ts#L984) to check for the new design. Using this you can use the correct query selector (either old or new one).

Once you get the correct element rest of the logic remains the same.

sindresorhus:
Bump

PR #1652, opened 2021-06-05
[WIP] Sidebar Modes for New Design

mquevill:
I was able to properly create the "hidden" and "narrow" modes. The "wide" mode is currently a work in progress for the following reason (everything else works as expected). It seems that the design actually _removes_ the elements that contain the conversation previews. For the "narrow" mode, I used `visibility: hidden`, but going the opposite direction is not as easy.

image

Any advice for how to keep these previews with a wide sidebar? I'm thinking that it cannot be a CSS-only solution, unfortunately…

[One weird quirk with their CSS: Some changes are triggered with `max-width: 900px` and some are triggered with `max-width: 899 px`. This means that if you set the width to exactly 900 px, you'll get a narrow sidebar, but the conversation previews are there (but you can only see half of the first letter or so).]

### DIFF
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -133,3 +133,24 @@ html.sidebar-hidden .o36gj0jk.g0mhvs5p {
 	width: 0;
 	height: 0;
 }
+
+/* Narrow: Hide conversation previews */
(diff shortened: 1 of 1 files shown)


sindresorhus:
Is this something you have time to finish or should I close?

mquevill:
The first two commits (efdc4b6 and ec98e06) do fix current issues, but I have not been able to figure out how to work around the 900 px limits. We could pull those 2 commits in as partial fixes, if you wanted. Otherwise, feel free to close for now, and I may revisit later.

PR #1653, opened 2021-06-05
Eliminate check for `darkMode` setting

mquevill:
This check was missed in #1637. However, this might not even be necessary any more? I did some testing with Dark, Light, and System Appearance, and there were no flashes of incorrect colors with this section commented out.


dusansimic (approved):
I suppose we can't test this on the old design (there seems to be no white flash on the new design even without that if statement) but the logic seems to be fine with me 👍🏻.

PR #1655, opened 2021-06-06
Scale videos down to max width

### LINKED ISSUE #1631: Videos too big on chat
Can you make the videos look smaller while scrolling in chat?
I can't see a the full video frame without clicking on it.
Also, can you make it auto resize if it goes past the right side? It's useful when you don't use the window maximized.

image

mquevill:
This is done automatically for pictures, but videos are a static size normally (501 px). If the window gets smaller than that, the video will scale with the max. width of the conversation.

[I tried to make an additional scaling for the height, but ran into more issues. The height of a picture/video is set by the width, it seems. Also, the height of the conversation "box" is much larger than is actually visible (to contain cached conversation items), so calculating with percentages or `vh` is more complicated than expected...]

Fixes #1631.

### DIFF
--- a/css/new-design/browser.css
+++ b/css/new-design/browser.css
@@ -91,6 +91,11 @@ html.private-mode [role='main'] .d2edcug0.hpfvmrgz.qv66sw1b.c1et5uql.b0tq1wua.a8
 	filter: blur(5px);
 }
 
+/* Force max-width on videos */
+.opwvks06.hop1g133.linmgsc8.t63ysoy8.qutah8gn.ni8dbmo4.stjgntxs.ktxn16wu.jz9ahs1c.efwgsih4.e72ty7fz.qmr60zad.qlfml3jp.inkptoze {
+	max-width: 100%;
+}
+
 /* Hide the "Messenger App for Windows" banner in chat list */
 .os-win32 .rq0escxv.l9j0dhe7.du4w35lb.j83agx80.g5gj957u.rj1gh0hx.buofh1pr.hpfvmrgz.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.s9xz0pwp.c4m0enpj.rl78xhln.srn514ro.sn0e7ne5.f6rbj1fe.l3ldwz01 {
 	display: none;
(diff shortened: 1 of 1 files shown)


dusansimic:
I've look over the page source a bit and yes, it seems quite difficult to do this with pure css 😅. What comes to my mind is writing a mutation observer for video elements and changing their size as they are created (it seems to me there are no modifications to the element that has with 501 px after it's created which is great for us). That way with a bit of JS it would be very easy to calculate the exact width needed for the specific video element to have an acceptable height (I personally like 500-550 px height).

What are your thoughts on this approach? I know that we've already talked about the downsides of adding mutation observer of everything that needs a change but I don't really see right now how this can be done in a better way on the new design which is very unforgiving. I'm not sure what the performance consequences would be for using this many mutation observers but if they're focused on a specific part of the page (for this it would be the chat pane) it might not be that bad.

mquevill:
Natively, Facebook does not scale images or videos when they are too tall. I think this is related to the nature of a vertical scroll in the chat: If people need to scroll vertically to see previous/next messages, they can scroll to see an image. This PR only addressed the fact that Facebook didn't set a `max-width` scaling for videos. I think this was the main issue in #1631, but it would be an added benefit if `max-height` was implemented as well, to ensure the user sees the whole item, even if it gets scaled way down.

The image scaling is a bit more hard-coded in, and we would need to use the `!important` tag to overwrite the maximum values. But if we are able to limit the height of media, I would want to do it for both images and videos. (There might be other things that people can send that should get scaled, but these are the 2 main things that come to mind.)

In regards to no change seen with this PR: I was testing this with 2 videos I sent to myself in landscape mode. Both had the same class where it defines a width of 501px. The changes are shown below.

Can be larger than 501px:
image

Trying to make window narrower (video width should be less than 501px):
image

After the change implemented in this PR:
image

dusansimic:
Okay, I see what you mean now. I didn't understand the issue properly. I was thinking about the video height (that was an issue on the old design too which we fixed with custom css). I'll approve the changes once I've tested this again. Sorry for the inconvenience.

dusansimic:
This seems to work great for the videos that the user sends but it doesn't set the max width for the videos that user receives for some reason. It applies the style to the div element of the received video but it doesn't respect that style and remains the same size.

mquevill:
I was testing this with a video I had sent to myself, but it just counts as "sent", so thanks for pointing this out!

I had to add 2 elements for the "received" videos. In fact, all 3 are necessary. But the sent videos were able to be scaled with just the one.

dusansimic (approved):
Looks good now, thanks!

PR #1665, opened 2021-07-06
Migrate `darkMode` settings in config file

mquevill:
The `darkMode` setting was replaced by `theme`, but the configuration file was not updated accordingly.

I believe that `'system'` should be the default, but I wouldn't want to force it on everyone who may have had `darkMode` set for specific reasons. This migration honors any previous configuration with `darkMode`.

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -232,9 +232,28 @@ function updateSidebarSetting(store: Store<StoreType>): void {
 	}
 }
 
+function updateThemeSetting(store: Store<StoreType>): void {
+	if (store.get('followSystemAppearance')) {
+		store.set('theme', 'system');
+		// @ts-expect-error
+		store.delete('followSystemAppearance');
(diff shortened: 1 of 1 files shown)


dusansimic:
I think both `followSystemAppearance` and `darkMode` settings should be removed but this way only one is removed (`followSystemAppearance` has a priority and if it's disabled then `darkMode` setting is used and only one which is used is removed, not both of them). I'd rather just set the `theme` setting according to `followSystemAppearance` and `darkMode` if it's not already set and then remove old settings (both of them). Logic would be much cleaner.

mquevill:
I reused the logic from the old `setDarkModeElement()` function and then do conditional deletes. I feel like it might be a bit clunky with the `undefined` checks, but I think it should cover all cases.

dusansimic (approved):

PR #1676, opened 2021-08-01
Optimized images.

RDKRACZ:
Lossless png compression.


sindresorhus:
Thanks, but I prefer not to do this. Optimizing these images have caused problems in the past in the final app.

RDKRACZ:
I understand.

Issue #1503, opened 2020-12-20
Block chat seen option no longer working with new Messenger.com


kevin-fang:
The block chat seen option no longer works -- my guess is that the `change_read_status.php` endpoint has been changed with the new Messenger.com update, but I can't confirm.

To reproduce: enable "Block Seen Indicator" in the Caprine `File` menu, then open a chat, while looking on another Messenger.com instance, and see that the message will be marked as read.

---

Caprine 2.51.2
Electron 10.1.5
darwin 20.2.0
Locale: en-US

tomeczeka:
yeah everything's broken after the messenger.com update

ammanley:
@kevin-fang I've been trolling the threads a bit to see if anyone has a decent idea of the magnitude of refactoring that would be necessary to re-enable the mass of functionality broken by the FB Messenger update, I'm wondering if you've got any inkling on how big a change this _particular_ feature might be? I'm considering attempting to push a PR for this, but wading through what little we know so far has been mind-blowing.

kevin-fang:
@ammanley I think I've located where the hide read feature is implemented in the [source code](https://github.com/sindresorhus/caprine/blob/d128989ce5688917bde3e0825e914f0285d8face/source/index.ts#L228). I've tried to look into the network requests made when you read a message on the web client but can't find any requests made to a link with `change_read_status`, so it looks like they changed something about their API. I don't know what the requests looked like before the update, though, so I'm not sure it's relevant.

Theoretically a few tweaks to this function might work, but would require a bit of investigation on the Messenger website.

dusansimic:
Hi everyone and thanks for taking interest in this issue. So far, as you might assume, we've been overflown with issue reports about same issues over and over again with little to no investigation or details about how Messenger handles everything now. If anyone is interested in helping out or at least researching a bit more in depth this issue and how Messenger might be handling those "privacy" features (sending seen reports, typing reports and block delivery receipts), we'll be very thankful. Right now because of the scale of the update of Messenger, we are basically building every feature from scratch and it might take some time until we get back on our feet.

ammanley:
Would love to help, never worked with Electron before so reading up on that first. Looks like they might be using the graphql API rather than PHP endpoints for the Messenger stuff now, which...might be unfortunate if there is no way to access the `Message` GQL entity without it activating the read receipts on their backend :-\

AggelosM:
[...]

PR #1688, opened 2021-08-19
WIP: Fix privacy features

dusansimic:
This pr fixes two privacy features on new design: blocking typing indicator and blocking read indicators. ~It doesn't yet block delivery receipts since I haven't yet found out how they are sent (it's possible that they can't be blocked)~ It's blocking delivery receipts too and also active indicator can be blocked which is a feature that could be considered.

This is a WIP pr so it's not ready to be merged. Once it's ready to be merged I'll remove the WIP prefix in the title. I'm not sure what side effects are there but in my testing there seem to be none.

One big change here is __disabling context isolation__ on main window. This is done so Proxy of WebSocket class could be created. Otherwise it wouldn't be possible to execute code for creating a Proxy class before the native one is used. It might be a security risk so this needs to be reviewed very carefully but it is a working solution. Similar logic could be used for fetching receiving message content from which a notification could be made.

Any suggestions regarding this (or a better solution) are very welcome!

Closes: #1503


ChaimGoldstein:
I don't know if this is the proper place to post issues regarding the PR. 

I recently tested out the changes made in this Pull Request. Apparently the block seen feature has an error. Apparently if i activate the said feature Caprine is unable to send any outgoing messages. However, I can still react to incoming messages. I have tested this out. Here are some screenshots as proof, thank you.

image

dusansimic:
> if i activate the said feature Caprine is unable to send any outgoing messages

I'm aware of this but thanks for the report. I've haven't had time to work on this properly so it's currently in this broken state. Once I've fixed it and everything works as expected, I'll ping again for testing if that's fine.

ChaimGoldstein:
> > if i activate the said feature Caprine is unable to send any outgoing messages
> 
> I'm aware of this but thanks for the report. I've haven't had time to work on this properly so it's currently in this broken state. Once I've fixed it and everything works as expected, I'll ping again for testing if that's fine.

Thank you, I also noticed that it flags messages received the first time you load up Caprine. Probably because of a limitation too, I guess.

mavaa:
Hi, just found out about caprine and went stragiht for this PR as the seen blocker is the only thing I'm after 🙃 

What I found out:
- "seen" blocking works 🎉 
- Sending does not with blocking enabled, but:
- You can turn off blocking to send messages 🎉 

Even if I send the message with the blocker on, it seems like it will retry with a given interval until i turn it back off, and then successfully send it if I wait long enough. I did not experience any messages being marked as read when I started the application the first time, so that issue might be unrelated(?)

Could a workauroud for sending messages simply be to turn the block setting off and on again when a message is sent (if the option was enabled in the first place)? I would argue that sending a response is more or less a read receipt in itself.

AggelosM:
> Could a workauroud for sending messages simply be to turn the block setting off and on again when a message is sent (if the option was enabled in the first place)? I would argue that sending a response is more or less a read receipt in itself.

This could be a solution. And we could just mark the blocking option as "Experimental".

dusansimic:
> This could be a solution. And we could just mark the blocking option as "Experimental".

I'd prefer if we implemented the full feature and not and experimental workaround which is more of a hack than a workaround. I haven't really had free time for this to complete it and I'm not sure when I will. If anyone else is interested in doing this, go ahead, I've planned on doing something similar that's done in AlexStackDev's extension.

Currently privacy features are done by parsing websocket messages with regex which is not ideal and a bit hacky. I'd prefer to completely parse the websocket message like it's done in the mentioned extension but that would be a significant amount of work. Once those messages are parsed, I guess it could be possible to extract the received message contents and create a notifications on for received messages.

Amandaville:
> AlexStackDev's extension

Can you please provide a link to this extension?  I have searched and cannot find anything by that name.  Does it provide a "Block Seen Indicator" function for Messenger?

sindresorhus:
Any movement on this?

dusansimic:
> Any movement on this?

Not for some time, no. Feel free to close this if there's no one to take over. If at some point I find some free time (currently there is none at all) I'll get back to this and get it done.

Amandaville:
Does closing this issue mean that it's finally fixed after all this time?  I don't see a newer release than December 2021 and this was just closed.  When can we expect the updated release with these fixes to the privacy features?

TheWiresharkGuy:
Hi, is it fixed with [v2.55.5](https://github.com/sindresorhus/caprine/releases/tag/v2.55.5)?

Issue #1562, opened 2021-02-17
Notifications not working on new design


intpg:
Caprine 2.52.0, Windows 10 20H2
Visual notifications, sound notifications, badges aren't working.
I removed everything from AppData/Roaming and reinstalled Caprine.
Windows notifications for Caprine are enabled.

bycEEE:
Same issue on windows and OSX.

dusansimic:
Hi! Notifications are not working on new Messenger design since they are not implemented by Messenger website. Because we display reformatted notifications from the website, now we can't display any notifications because there isn't any. Once Messenger website adds notifications, we will add them to Caprine.

intpg:
Can't Caprine just detect a new message itself to bring notification?
There is always a new message ticker in the window, which could be an indicator.

dusansimic:
The notifications are not really the same as a new message indicator. I've fixed the issue with tray indicator not turning blue for new messages and is waiting to be merged (I'd expect it in the next update).

The notifications need to have some content in them which would be the message that was received. In [this](https://github.com/sindresorhus/caprine/issues/1522#issuecomment-757998055) comment I've mentioned that there would need to be reverse engineering and a lot of hacking put into just getting something to work (I can't say it will work well) and I don't really think it is worth while we have a lot of other features that need to be fixed because of the new design.

I'd rather wait for Messenger to implement notifications on the website than trying to add this feature by reverse engineering the whole websites communication system.

What could maybe be done is showing a simple "There is a new message" notification which would indicate a new message was received but it won't have any information about the message or the sender which would be expected.

intpg:
Anything what would allow the user to see that there is any message awaiting would be useful.
Either tray icon color, or general notification, that would be already helpful.

DarekSz:
> Anything what would allow the user to see that there is any message awaiting would be useful.
> Either tray icon color, or general notification, that would be already helpful.
Anything that is easy to notice. Just black vs. blue tray icon is a bit to easy to miss for me.
Tray icon blinking, red dot on taskbar icon (maybe with number of conversations that have new messages).
Something noticeable.

ErebosGR:
> I've fixed the issue with tray indicator not turning blue for new messages and is waiting to be merged (I'd expect it in the next update).

As of v2.52.1, this doesn't seem to be fixed. Is it still not merged?

dusansimic:
[...]

PR #1739, opened 2021-12-01
Workaround for notifications

cool00geek:
Notifications have not been working since the redesign (Workaround for #1562). This PR aims to work around this missing functionality.

The message sidebar changes to show the unread dot when a new message is received. A mutation event listener has been added to watch for this new blue dot. When the dot appears, we can obtain the message content, message author, and profile picture of the chat and send a notification using existing notification functions.

The downside with this workaround is that there is no notification when multiple new messages are received in the same chat (when the blue dot already exists and a new message is received). However, I believe just having some notification is far better than the current status of not having any notification.

There were also a couple of errors in obtaining the conversation list for other functions in `index.ts`. I have thus updated the class list and modified the functions to reduce the likelihood of errors.

Example notification on KDE Plasma:
image

### DIFF
--- a/source/browser/conversation-list.ts
+++ b/source/browser/conversation-list.ts
@@ -2,6 +2,7 @@ import {ipcRenderer as ipc} from 'electron-better-ipc';
 import elementReady = require('element-ready');
 import selectors from './selectors';
 import {isNewDesign} from '../browser';
+import {isNull} from 'lodash';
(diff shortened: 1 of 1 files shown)


Qynklee:
Thanks for your help! Hope your commit approve soon :smile: KDE too :smile:

Issue #1740, opened 2021-12-01
Rec'd Messages Blank


PreachJohn:
Load Caprine. Messages themselves have turned White from Dark Setting.
All Received Messages are blank. Sent by me are normal.

---

Caprine 2.55.0
Electron 10.4.7
win32 6.1.7601
Locale: en-US

mquevill:
Duplicate of #1737

PreachJohn:
#1737 only mention Colors Broken.
I mention also Received Messages blank.
Overlap, yes. Duplicate, no?

mquevill:
Are the messages actually blank? Or white text on white background? Try highlighting where the text should be and the message should still be there. The dark theme normally has light text, but on a light background it's nearly impossible to see.

PreachJohn:
Actually blank.
![Screen Shot 12-01-21 at 04 24 PM](https://user-images.githubusercontent.com/31393296/144330529-fdc635ec-ce52-4423-b864-dfafea4e7ca8.PNG)
Cheers

dusansimic:
@PreachJohn I think that's actually what Michael was talking about. The messages are not blank, they are just the same color as the background. This issue is caused by colors being broken.

You could select the text of the message, it's there but the color makes it invisible.

Issue #1738, opened 2021-12-01
Dark mode malfunctioning


DNM1008:
Dark mode now just made the conversation background white, including the text, making it quite hard (impossible) to read. the background  of the date is still black though
![image](https://user-images.githubusercontent.com/54259825/144159941-67afdc5b-52dc-4280-aac2-7e386daf6740.png)

---

Caprine 2.55.0
Electron 10.4.7
win32 10.0.19043
Locale: en-GB

vanastasis:
Same

Caprine 2.55.0
Electron 10.4.7
linux 5.14.0-kali4-amd64
Locale: en-GB

pernielsentikaer:
Duplicate of #1737

FunkyJustin:
Same case here

spyd3r-b0y:
how to solve?

PixsaOJ:
Duplicate of #1737

Close this one to help maintainers.

Issue #1737, opened 2021-11-30
Colors broken in dark mode


frozenpandaman:
Started happening within the past day. v2.55.0, installed via Homebrew on macOS 12.0.1.

![image](https://user-images.githubusercontent.com/7245874/144143047-b055dd80-3306-4487-aefe-80cc2c9380a7.png)

The black rectangles is my censoring of private information. The problems are that the chat body/background is white, black corners appear around messages sent by me, and text is invisible in messages sent by others.

joshualagat:
Experiencing the same issue on Manjaro Linux with Caprine from Official Repositories

Okiesmokie:
Same issue here on Windows 10, Caprine v2.55.0

huoi:
Same issue here on Arch Linux, Caprine 2.55.0.

jmmanlupig:
Can confirm as well.
Caprine 2.55.0
Arch Linux & Windows 10

jcatrdev:
Also experienced the same problem. Just open up Caprine this morning then I saw the issue.
![Screenshot_20211201_090122](https://user-images.githubusercontent.com/86816436/144152995-4ebc4eb2-e50f-4414-b012-d99adb6f9b41.png)

riverstore:
Hi, I meet same problem

Caprine 2.55.0
Electron 10.4.7
win32 10.0.18362
Locale: en-US

Regard.

quanKM:
Same issue on Mac OS, Caprine 2.55.0.

nkenschaft:
Same here

Pop!_OS 21.04
Caprine 2.55.0

mquevill:
It looks like the row backgrounds aren't getting the dark mode set properly. It seems like a widespread issue, not specific to any OS or version. It is most likely caused by a change on Facebook's end.

For reference, the backgrounds are contained in the `role="row"` items (example below). These incorrectly have `__fb-light-mode` when they need `__fb-dark-mode` to have the right colors.
```html
<div role="row" class="__fb-light-mode l9j0dhe7" style="--mwp-reply-background-color:var(--comment-footer-background);">
```

spyd3r-b0y:
last night everything was fine. Today I'm also facing same problem :|

TheBricktop:
same here win10 caprine 2.55

spyd3r-b0y:
how to solve this problem?

maxoku:
[...]

PR #1742, opened 2021-12-02
Fix dark mode

dusansimic:
This is a workaround for the issue with dark mode being broken by use of light mode classes in some elements in ui. This is creating a major problem since received messages are unreadable.

It's not an ideal solution since it uses mutation observer on root element so any suggestions are welcome.

Closes: #1737
Closes: #1738 
Closes: #1740


PixsaOJ:
Bless you

joshualagat:
Nice

frozenpandaman:
Thanks!

PR #1777, opened 2022-01-02
Use 'emoji' font alias on Linux for system emoji

AidanGG:
On Linux with fontconfig, the system font for emoji is the "emoji" font rather than "system-ui" [1-2]. When using the system-ui font on standard configurations, it falls back to using "sans-serif" before the emoji font, in my case Arimo before Noto Color Emoji:

```bash
$ fc-match system-ui
Arimo-Regular.ttf: "Arimo" "Regular"
```

This presents a problem for emoji characters which are included as monochrome unicode glyphs in the sans-serif font (such as the smiley face ☺), since these are rendered with the monochrome characters from sans-serif rather than the emoji font. This patch makes use of the emoji alias shipped by default in fontconfig on Linux systems to preferentially use the system emoji font.

[1] https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/conf.d/45-generic.conf
[2] https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/conf.d/60-generic.conf

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -539,7 +539,7 @@ ipc.answerMain('render-overlay-icon', (messageCount: number): {data: string; tex
 ipc.answerMain('render-native-emoji', (emoji: string): string => {
 	const canvas = document.createElement('canvas');
 	const context = canvas.getContext('2d')!;
-	const systemFont = is.linux ? 'emoji' : 'system-ui';
+	const systemFont = is.linux ? 'emoji, system-ui' : 'system-ui';
 	canvas.width = 256;
 	canvas.height = 256;
 	context.textAlign = 'center';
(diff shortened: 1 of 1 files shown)


source/browser.ts
```diff
@@ -539,16 +539,17 @@ ipc.answerMain('render-overlay-icon', (messageCount: number): {data: string; tex
 ipc.answerMain('render-native-emoji', (emoji: string): string => {
 	const canvas = document.createElement('canvas');
 	const context = canvas.getContext('2d')!;
+	const systemFont = is.linux ? 'emoji' : 'system-ui';
```
sindresorhus, line 542:
```suggestion
	const systemFont = is.linux ? 'emoji, system-ui' : 'system-ui';
```

Would probably be smart to fas back on `system-ui` in case some Linux distro doesn't have `emoji`?

Issue #1795, opened 2022-03-15
Please push the current version to the Snap store manually?


IndependentIgloo:
An issue with the build pipeline that was fixed in December (https://github.com/sindresorhus/caprine/issues/1766) meant that Caprine was not being pushed to the Snap store upon each release.  Now that the problem is fixed, the advice is to hang tight and wait for the next release of Caprine, which should now result in that release being pushed to Snap as normal.

However, it has been 3 months since this fix was put in and there has not yet been a new release of Caprine.  As is apparent from these issues (https://github.com/sindresorhus/caprine/issues/1788, https://github.com/sindresorhus/caprine/issues/1794), new users are being stung by the installing the now-buggy older version from Snap, which was released in _April 2021_.

Although a workaround exists in the form of installing the AppImage, this represents a change in workflow for those that rely on Snap to manage many of the apps on their system, and necessitates having to remember to revert back to Snap once a new release becomes available through the store.

It would spare Snap users a huge amount of pain if the current version of Caprine was pushed to Snap manually as a one-off, since there is no telling when the next release of Caprine might appear.

Thanks.

sindresorhus:
I'm unable to run the app locally because I got a new Apple silicon Mac. @dusansimic Would you be able to just run the app from `main` and make sure it runs ok?

dusansimic:
I've tried it out now (the latest version) and it seems to work correctly. There is still an issue with fonts in dialogs tho (#1284). I sadly don't have any free time to fix these bugs at the moment.

sindresorhus:
https://github.com/sindresorhus/caprine/releases/tag/v2.55.3

dusansimic:
Snap package still doesn't seem to be published. I'm not sure when `electron-builder` has changed but it seems that from the docs, `snapStore` needs to be explicitly specified in publish config (if it's redefined as it is done in `package.json` to add github as a provider).

In order to avoid waiting for a next release, I'd recommend publishing this release manually using the `snap` tool. Other option is to make this change, make a new release (2.55.4 for example) and publish it that way.

PR #1796, opened 2022-03-20
Fix Snap publishing

dusansimic:
Closes: #1795

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,15 @@
 				"camera",
 				"removable-media"
 			],
-			"publish": "github"
+			"publish": [
+				{
+					"provider": "github"
+				},
+				{
+					"provider": "snapStore",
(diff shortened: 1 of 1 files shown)


dusansimic:
That is due to snap publish not being complete. It breaks due to some issue with permissions/token. One idea I have is that at some point additional permissions were added as a requirement for publishing packages (I'm guessing the same token is used since snap publishing was added to the pipeline) so a solution could be generating another token. I'm honestly not sure weather this will work so I have no guarantees.

mquevill:
From what I'm reading in other projects with similar errors, the errors in the CI logs seem to be related to expired Snapcraft credentials, since it's only the Ubuntu builds that are failing. I think @dusansimic's advice for generating another token is worth a try.

PR #1822, opened 2022-04-19
Fix workaround for notifications

cool00geek:
A fix for system notifications post-redesign (#1562) was merged (#1739).

A change in Messenger has caused a null element causing the notification workaround to no longer function.

This will now verify that the element is not null before attempting to use its properties, no longer resulting in an exception and allowing the notification functionality to function as intended again.


Neo-29:
What about this issue?
Still?

PR #1853, opened 2022-06-26
Select log out as last menu item

mquevill:
Added functionality to select negatively indexed menu items, since "Log out" is always last in the menu item list.

[Selecting item 11 broke when another "New! Messenger for Windows" item was added to the list. I think that "Log out" will stay as the last item, so this should be more future-proof, but we'll see what the Messenger backend devs do.]

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -66,7 +66,12 @@ async function selectMenuItem(isNewDesign: boolean, itemNumber: number): Promise
 			`${conversationMenuSelectorNewDesign} [role=menuitem]`
 		);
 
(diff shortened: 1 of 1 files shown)


dusansimic (approved):

sindresorhus:
@mquevill Would you be interested in joining the project as a maintainer? You have done a lot of great improvements to Caprine. I know this is a little bit late, but I got busy during the summer.

mquevill:
@sindresorhus Yes, I would gladly step in as a maintainer! I've also been a bit busy more recently, but should be able to help out with this project on a regular basis.

sindresorhus:
@mquevill Awesome :)

PR #1870, opened 2022-08-06
Restore snap publishing

dusansimic:
I have access to the snap package on the snap store and have managed to manually publish the latest version. I however don't have access to change the access token for ci so I will either need that access in this repo or I could send Sindre the token somehow so he can add it himself.

// @sindresorhus

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,10 +28,10 @@ jobs:
         with:
           node-version: 14
       - uses: c-hive/gha-npm-cache@v1
-      # - uses: samuelmeuli/action-snapcraft@v1
-      #   if: startsWith(matrix.os, 'ubuntu')
-      #   with:
-      #     snapcraft_token: ${{ secrets.snapcraft_token }}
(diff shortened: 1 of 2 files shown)


lefterisgar:
Is this why CI fails?

dusansimic:
> Is this why CI fails?

No, ci fails on `npm test` (specifically typescript lint). It started failing on tests since `xo` and `typescript` dependencies were bumped (https://github.com/sindresorhus/caprine/commit/6effe7aa281c556ad0bfa5f6d0a5f3ef0a6ea5dd). A solution for that would be making linter happy and fixing it 😁.

Before that, ci failed because electron-builder tried to publish a release (during the macOS build so it's unrelated to the snap blockage that was fixed in https://github.com/sindresorhus/caprine/commit/6effe7aa281c556ad0bfa5f6d0a5f3ef0a6ea5dd). I'll investigate that and probably tweak the ci config a bit so releases are only made on tagged commits.

lefterisgar:
I see... I'm trying to make linter happy so don't worry.

PR #1879, opened 2022-08-19
Remove code related to old design

lefterisgar:
Old design was deprecated more than a year ago by Facebook.

Why have redundant code that does nothing and conditions (for new design) that always return true? This PR removes some of the old code but not everything. There is a lot going on under the hood and a followup PR might be needed.

Open for discussion.


dusansimic (approved):
Code seemes good. If there is no way to be sure that Facebook completely removed the old design, we should keep a close eye on this.

PR #1881, opened 2022-08-19
Fix "Show Message Buttons" option

lefterisgar:
This PR updates the message buttons' classes which were changed by the latest Messenger UI update.

I must say I have noticed a bug after the UI update. If you disable "Show Message Buttons", message buttons are hidden as intended. However, after restarting Caprine, message buttons are visible again and you have to toggle the option on and back off for it to work.


source/index.ts
```diff
@@ -497,8 +497,6 @@ function createMainWindow(): BrowserWindow {
 			defaultStatus: config.get('notificationsMuted'),
 		}));
 
-		ipcMain.callRenderer(mainWindow, 'toggle-message-buttons', config.get('showMessageButtons'));
```
dusansimic, line 500: Why is this removed?

dusansimic (changes requested):

source/index.ts
lefterisgar: It doesn't seem to do anything. It is broken anyways.

dusansimic: If you'd like to remove the auto-toggle on app start, then the option should be reset too (it should be on by default).

I however think it would be better to find a fix for this issue rather than removing a feature. Some people might interpret option not being applied on start as an issue.

dusansimic: I'm tracking down the issue right now. It seems that the preferences modal is not being opened and which is done for toggleing notifications sound (it's done at start in index.ts [link](https://github.com/sindresorhus/caprine/blob/main/source/index.ts#L495)). The execution gets blocked and Node.js never reaches the next statement which is toggleing message buttons visibility. I'm trying to find out how to fix it as I'm writing this so I'll push a patch on this branch once it's fixed.

lefterisgar: > If you'd like to remove the auto-toggle on app start, then the option should be reset too (it should be on by default).
> 
> I however think it would be better to find a fix for this issue rather than removing a feature. Some people might interpret option not being applied on start as an issue.

No I also want this feature to work! It's just that this line of code fixes nothing at the moment.

lefterisgar: > I'm tracking down the issue right now. It seems that the preferences modal is not being opened and which is done for toggleing notifications sound (it's done at start in index.ts [link](https://github.com/sindresorhus/caprine/blob/main/source/index.ts#L495)). The execution gets blocked and Node.js never reaches the next statement which is toggleing message buttons visibility. I'm trying to find out how to fix it as I'm writing this so I'll push a patch on this branch once it's fixed.

Great! Hope you eventually sort it out!

lefterisgar: > The execution gets blocked and Node.js never reaches the next statement which is toggleing message buttons visibility.

Exactly! Running the code before the offending line fixed it!

PR #1884, opened 2022-08-20
Fix private mode

lefterisgar:
This PR fixes private mode which of course broke after Facebook's latest UI redesign.


dusansimic:
Some profile pics that are blured have a cutoff and some don't. Imho, it looks okay and we don't necessarily need to fix that now but it should be something to keep in mind since it's a bit inconsistent.
Screenshot from 2022-08-21 10-33-32

dusansimic (approved):

lefterisgar:
> cutoff

I agree. Is this probably caused from excessive blur on such a small picture? I assume you mean the read indicator picture.

dusansimic:
> I assume you mean the read indicator picture.

Both the read indicator and the profile pictures on the left. You can se that my profile picture is blured fine but the picture of the first chat is cut off. On the second chat which is a group chat, you can see that the first profile picture (one on the left) is blured fine but the second one is cut off. The behavior is a bit weird but we could look into that later.

PR #1834, opened 2022-05-20
Flathub link, small additions and fixes

Danik1601:
Added many small additions to readme.md:
- Various information about Flatpak and AppImage and Immutable Linux distributions
- "Download on Flathub" button
- Fixes to Touch Bar support link
- Other minor tweaks and changes
- Small tweak in Fedora section to comply with Markdown guidelines

PR #1885, opened 2022-08-20
Various improvements to readme

lefterisgar:
This PR adds a couple of changes to the readme:

Based on original work by @Danik1601.

- Table that shows how to install Caprine for (almost) every Linux distribution out there along with how to setup automatic updates and who is responsible for maintaining the package.
- Deb-get & Packagecloud repository (Linux)
- Improved installation instructions
- Info about immutable Linux distributions
- Mark some features as macOS only and fix some links that are caused from that
- Remove "Disable autoplay for videos" from highlights until it's actually fixed
- Centered logo and description below

Feel free to discuss any changes.

Note: The squashed commit must be co-authored by @Danik1601. Thanks!


Danik1601:
Thank you! I defended my bachelor's degree and entered the master's program this summer so I was quit busy and stuff

dusansimic:
Thank you both @lefterisgar and @Danik1601 for this work. It looks much cleaner and easier to understand. I just have a couple of suggestions which I'll file as a review.

@Danik1601 congratulations for the graduation and entering a masters programme!

readme.md
```diff
@@ -59,20 +59,124 @@ Or with [Homebrew](https://brew.sh): `$ brew install caprine`
...
 
-Arch Linux: `pacman -S caprine`
+```sh
+curl -s https://packagecloud.io/install/repositories/lefterisgar/caprine/script.deb.sh | sudo bash
+```
+
+Note: The Packagecloud repository has limited bandwidth. This option will stop working each month when the bandwidth limit is reached.
 
-Fedora:
+#### Fedora Copr:
```
dusansimic, line 157: COPR repo supports various distros as noted in the table and it might be confusing for some people that it says here Fedora. It might be better to use just COPR or list all supported distros. A list of distros might look better since it doesn't require people to understand what COPR is and just requires the to know which distro they're using.

Also, instructions for adding a repo to OpenSuse are not the same as for RHEL/Fedora since OS uses a different package manager from RHEL. We should either add all instructions or point people to the repo website so they can figure it out there.

readme.md
```diff
@@ -59,20 +59,124 @@ Or with [Homebrew](https://brew.sh): `$ brew install caprine`
...
+</table>
+
+#### Installation using deb-get:
+
+* Download and install [deb-get](https://github.com/wimpysworld/deb-get).
+* Run `deb-get install caprine`.
+
+Note: deb-get is 3rd party software, not to be associated with apt-get.
+
+#### Debian Packagecloud APT repository:
```
dusansimic, line 147: Is Packagecloud Debian specific or is it a regular APT repo. If it's Debian specific then it's fine but if it's not, Debian should be removed from here to avoid confusion, just like with COPR instructions.

lefterisgar:
> Thank you! I defended my bachelor's degree and entered the master's program this summer so I was quite busy and stuff
> 
> UPD: Just looked into it and I absolutely love the table you made. I hope that the maintainers will approve all that

Congratulations on your graduation! You're amazing! I wish you all the best!

That's great! I didn't really know if it was okay in terms of readability but I'm glad you like it!

lefterisgar:
> It looks much cleaner and easier to understand. I just have a couple of suggestions which I'll file as a review.

@dusansimic Great feel free to suggest whatever you want. I also think it needs a couple of improvements here and there.

readme.md
```diff
@@ -59,20 +59,124 @@ Or with [Homebrew](https://brew.sh): `$ brew install caprine`
...
+		<td>RHEL / Fedora / openSUSE</td>
+		<td>Copr</td>
+		<td align="center">✔️</td>
+		<td>Dušan Simić</td>
+		<td>
+			Follow the <a href=#fedora-copr>instructions below</a>
+		</td>
+	</tr>
+	<tr>
+		<td>AppImage (Universal)</td>
```
dusansimic, line 112: I'm not sure if it's entirelly correct to say AppImage (or for that matter Flatpak and Snap) is universal since some distros might not have support for it. It's aim __is__ to be universal but weather that's achieved or not is a bit tricky to say. I'm open for discussion here but imho, universal might not be a good fit here. We just don't want to mislead users.

readme.md
```diff
@@ -59,20 +59,124 @@ Or with [Homebrew](https://brew.sh): `$ brew install caprine`
...
+Make it [executable](https://discourse.appimage.org/t/how-to-run-an-appimage/80):
+
+```sh
+chmod +x Caprine-2.xx.x.AppImage
+```
+
+Then run it!
+
+#### About immutable Linux distributions:
+[Fedora Silverblue](https://silverblue.fedoraproject.org), [Fedora Kinoite](https://kinoite.fedoraproject.org), [EndlessOS](https://endlessos.com), [CarbonOS](https://carbon.sh) and other immutable distributions only support Flatpak and/or AppImage.*
```
dusansimic, line 177: Do those distros support Snap? If yes, Snap could be added here too. Afaik, Snap is not really tested and aimed to be supported on Fedora Silverblue but it's worth to check it.

readme.md
```diff
@@ -59,20 +59,124 @@ Or with [Homebrew](https://brew.sh): `$ brew install caprine`
...
 
-*The AppImage needs to be [made executable](https://discourse.appimage.org/t/how-to-run-an-appimage/80) after download.*
+*Note: On some distributions Flatpak must be [pre-configured manually](https://flatpak.org/setup).*
 
 ### Windows
```
dusansimic, line 181: We could add a list of repos you can download Caprine on Windows like choco and win-get while we're updating the readme.

dusansimic (changes requested):

readme.md
lefterisgar: Yes you are right. Is it better to write "APT repository (Powered by Packagecloud)" to give emphasis to the word "APT" so users don't get confused?

readme.md
lefterisgar: Yeah, sure!

readme.md
lefterisgar: > We should either add all instructions or point people to the repo website so they can figure it out there.

The repo website at the moment has the same instructions we have:

εικόνα

The instructions for openSUSE are different in the sense that on openSUSE you need to install dnf first, or is it something I"m missing?

readme.md
lefterisgar: AppImage, Flatpak and Snap are "universal" in the sense that they don't only run on, say Debian.
They run on multiple distros, more than you can fit inside the table 😄.

All the distributions we **currently support** have AppImage integration and BOTH Snap and Flatpak pages whose link is provided have instructions for most distros out there:

Snapcraft:
εικόνα

Flathub:
εικόνα


Bottom line: Yes, there might be some obscure distribution on which neither format can work but **WE DON'T SUPPORT** that!

readme.md
lefterisgar: Snap might be trickier to run on those. Also, (I know this is a big discussion) we should encourage the use of Flatpak for performance reasons. Users right now are complaining about Caprine being slow etc. and we shouldn't worsen it.

readme.md
lefterisgar: I have just found out that they haven't updated yet to 2.56.0. That means the maintainers are slow to update and this could be problematic. I prefer to not add those in order to ensure that the users have the best possible experience.

lefterisgar: Chocolatey:
εικόνα

Win-get:
εικόνα

readme.md
dusansimic: openSUSE uses zypper and from what I know, it has a different repos dir from dnf (dnf uses `/etc/yum.repos.d` and zypper uses `/etc/zypp/repos.d`) so you can't use dnf to enable the copr repo and then use zypper to install the package. You'll need to manually add a repo (copying a file and putting it in the correct location).

In general it's not ideal to distribute oS packages via copr since oS has it's own distribution platform (Open Build System). This is just a temporary solution for users of copr repos.

An ideal solution might be to just add a separate section for oS and note that users would have to add the repo manually and point them to the repo website.

readme.md
dusansimic: "APT repository (Packagecloud)" might be just a bit cleaner.

As a side note, OBS from openSUSE could be considered as an option for third party repos (the one I mentioned in the comment about COPR). It supports APT repos from what I know and it doesn't have a bandwith limit. There is also [gemfury](https://gemfury.com/) which I saw some packages use.

readme.md
lefterisgar: Ok, so would this one-liner work in oS tw:
```sh
$ curl -s https://copr.fedorainfracloud.org/coprs/dusansimic/caprine/repo/opensuse-tumbleweed/dusansimic-caprine-opensuse-tumbleweed.repo | sudo tee /etc/zypp/repos.d/caprine.repo
```

readme.md
dusansimic: I agree but my point was that in my opinion we should avoid describing pacakge format, we should just list them out. As I've already said in a pr that's also refactoring the readme ([link](https://github.com/sindresorhus/caprine/pull/1834)), opinions are fine but we should avoid them. What we could do is recommend official packages (which are AppImage and Snap for example).

These are just my thoughts on this topic but I don't see it as an issue either way!

readme.md
lefterisgar: > "APT repository (Packagecloud)" might be just a bit cleaner.

Okay, I will use that.

> As a side note, OBS from openSUSE could be considered as an option for third party repos (the one I mentioned in the comment about COPR). It supports APT repos from what I know and it doesn't have a bandwith limit.

Great option! So if I can get this to work, does it mean openSUSE users won't need to use Copr?

readme.md
dusansimic: > we should encourage the use of Flatpak for performance reasons

Users should have better performance with Flatpak but for one it's not an official package still but we could push for it to be official soon and also as I've already said, we shouldn't give our opinion we should just provide packages. Users should be free to choose whichever format they like based on their own opinion.

readme.md
dusansimic: That's probably because they are not in the loop with the updates, we could contact them to see what's the status of those packages and weather they are still willing to maintain them. Also regarding the choco pacakge, v2.55.7 didn't get a release published on GitHub so the maintainer probably didn't get a notification for that and only got it for v2.56 which was released a few days ago. Win-get seems to be the [same case](https://github.com/microsoft/winget-pkgs/tree/master/manifests/c/Caprine/Caprine) acording to the GitHub repo.

readme.md
dusansimic: > So if I can get this to work, does it mean openSUSE users won't need to use Copr?

If you build a pacakge for oS on OBS, yes. Users won't need to use copr.

readme.md
lefterisgar: 1) If it is provided by you, I would consider it to be official 😄
2) I agree that we shouldn't give our opinion but installing snap on an immutable distro seems too much effort for nothing. In short, immutable distros probably are mostly designed around Flatpak.

readme.md
lefterisgar: Sure! No problem at all! So do you want to get those options out of the table?

readme.md
lefterisgar: I guess we could add those repos after ensuring they still receive regular maintenance. This could happen in a second PR because adding them while they are still a few versions behind will only result in further confusion.

lefterisgar: > Also regarding the choco pacakge, v2.55.7 didn't get a release published on GitHub so the maintainer probably didn't get a notification for that and only got it for v2.56 which was released a few days ago.

Fair enough. Let's give them some time and see what's going to happen.

readme.md
dusansimic: Personally, I don't like the idea of maintaining one liners for a distro afaik none of the maintainers use. I'd much rather just say what they need to do and let them do it. I just find it more difficult to make sure oneliners like these work correctly than some standardized commands like `dnf`.

readme.md
dusansimic: No, just remove the "(Universal)" bit :)

lefterisgar: Oh, that's fine!

readme.md
dusansimic: > If it is provided by you, I would consider it to be official

Kind of true 😅. I'll look into adding the flatpak manifest to the caprine repo so it could be maintained here and then just copied to the flathub repo on release.

> I agree that we shouldn't give our opinion but installing snap on an immutable distro seems too much effort for nothing. In short, immutable distros probably are mostly designed around Flatpak.

I agree. Leave it as is. If some issue gets filed about this we'll revisit it.

readme.md
lefterisgar: Yes, this is not going to be a problem if I find out how to setup OBS.

readme.md
lefterisgar: Unfortunately, I wasn't able to setup OBS because it requires a complete source code upload and scripts which needs some effort and of course maintenance. I tested out Gemfury and it seems set and forget just like Packagecloud. I wrote a simple one liner for GitHub Actions to upload to it and everything looks okay. Now I'm going to update the readme file.

readme.md
lefterisgar: Resolved!

readme.md
```diff
@@ -59,20 +59,136 @@ Or with [Homebrew](https://brew.sh): `$ brew install caprine`
...
+
+```sh
+curl -s https://gist.githubusercontent.com/lefterisgar/c5f738fcb4d06d2efd7eb9215cf3dbe6/raw/35eff2fffa1275a32fe69ae2886e7e6394b5471b/addRepo.sh | sudo bash
+```
+
+Alternatively (for advanced users):
+```sh
+# Add the repository
+echo "deb [trusted=yes] https://TOKEN@apt.fury.io/USERNAME/ /" > \
+/etc/apt/sources.list.d/fury.list
```
dusansimic, line 159:
```suggestion
echo "deb [trusted=yes] https://apt.fury.io/lefterisgar/ /" > \
/etc/apt/sources.list.d/caprine.list
```

dusansimic (changes requested):

readme.md
lefterisgar: Oops! I missed it!

dusansimic:
Everything looks fantastic now! Thank you for this.

lefterisgar:
Well, if it wasn't for your help we wouldn't be there!

dusansimic (approved):

PR #1886, opened 2022-08-22
Add script for automatic APT repo setup

lefterisgar:
Gemfury implementation is complete. However, in contrast to Packagecloud, Gemfury does not provide a script for automated repo installation. That means providing a user-friendly way to for setting up the APT repo is up to us.

I have created a script that does the following:
- Checks if the system has APT installed (which means Debian-based).
- Adds the repository and checks to see if it was added properly.
- Updates the repositories (apt update) so Caprine can now be installed.
- Asks the user if he wants to install Caprine right away.

It can be run with a one liner which we will provide in the readme when #1885 is merged.

Here is how it looks like:
Screenshot_20220822_163013

### DIFF
--- a/addRepo.sh
+++ b/addRepo.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+# Made by Lefteris Garyfalakis
+# Last update: 22-08-2022
+
+COL_NC='\e[0m' # No Color
+COL_LIGHT_RED='\e[1;31m'
+COL_LIGHT_GREEN='\e[1;32m'
+COL_LIGHT_BLUE='\e[1;94m'
+COL_LIGHT_YELLOW='\e[1;93m'
+TICK="${COL_NC}[${COL_LIGHT_GREEN}✓${COL_NC}]"
+CROSS="${COL_NC}[${COL_LIGHT_RED}✗${COL_NC}]"
+INFO="${COL_NC}[${COL_LIGHT_YELLOW}i${COL_NC}]"
+QUESTION="${COL_NC}[${COL_LIGHT_BLUE}?${COL_NC}]"
+
+APT_SOURCE_PATH="/etc/apt/sources.list.d/caprine.list"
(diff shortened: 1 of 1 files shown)


dusansimic:
Personally, I like this script (and easy install scripts in general) and I think it's a great way for users to add repos for Caprine. The thing is, gemfury repo would be a third party repo technically and thus I think this script should be too. You could create a GitHub Gist for this script and maintain it there (it's the easiest way for users to review the script) and link it in the readme.

I'm just saying this so we would have less official stuff for now until we get Caprine in a fairly working order. Once we get on our feet with the app itself, I'm all for making as much of this stuff official asap (that would include ci jobs for publishing packages to other repos).

dusansimic:
Also readme could contain the same snippets for adding the repo manually, as an alternative for users who would prefer to do it themselves.

lefterisgar:
> Personally, I like this script (and easy install scripts in general) and I think it's a great way for users to add repos for Caprine. The thing is, gemfury repo would be a third party repo technically and thus I think this script should be too. You could create a GitHub Gist for this script and maintain it there (it's the easiest way for users to review the script) and link it in the readme.
> 
> I'm just saying this so we would have less official stuff for now until we get Caprine in a fairly working order. Once we get on our feet with the app itself, I'm all for making as much of this stuff official asap (that would include ci jobs for publishing packages to other repos).

That's fine! I'm going to do that and add a link in the readme.

> Also readme could contain the same snippets for adding the repo manually, as an alternative for users who would prefer to do it themselves.

I'm not really against that but personally, I think that the script ensures every user will have the best experience possible while adding the APT repo and we are 100% sure it is going to be properly added. So I will put the script option front and center and include an alternative way which is to run the commands as you said.

dusansimic:
Thanks! Great work with the script and setting up gemfury repo! Caprine now finally has a good way to natively (with apt) update on deb distros.

PR #1889, opened 2022-08-23
Set window title to "Caprine"

lefterisgar:
I have noticed that by default, messenger.com overrides the window title and sets it to "Messenger". Wouldn't it be better if it was set to "Caprine"?

εικόνα

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -442,6 +442,9 @@ function createMainWindow(): BrowserWindow {
 	const {webContents} = mainWindow;
 
 	webContents.on('dom-ready', async () => {
+		// Set window title to Caprine
+		mainWindow.setTitle(app.name);
+
 		const isNewDesign = await ipcMain.callRenderer<undefined, boolean>(mainWindow, 'check-new-ui');
 
 		await updateAppMenu({isNewDesign});


dusansimic:
Would this also set the number of unread messages in the title? Something makes me think that was also the behavior of the website title.

lefterisgar:
> Would this also set the number of unread messages in the title? Something makes me think that was also the behavior of the website title.

I see. Let me check that.

lefterisgar:
Mine stays the same. Always displays "Messenger" regardless of the number of unread messages.

lefterisgar:
Is that what you mean?
εικόνα

dusansimic:
Yes. I thought at some point messenger showed number of unread messages in page title but it might be something on previous design.

lefterisgar:
> Yes. I thought at some point messenger showed number of unread messages in page title but it might be something on previous design.

Maybe. Do you remember seeing it recently?

dusansimic:
I'm not sure. It's fine for now :) We can revisit it sometime later as part of working on fixing and implementing notifications and other indicators. It's nothing urgent.

lefterisgar:
True. The notifications part definitely needs an overhaul!

PR #1893, opened 2022-08-24
Completely remove old design code

lefterisgar:
This PR completely removes ALL code related to Messenger's old design which was phased out in late 2020. It will for sure make debugging the project easier.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -578,19 +578,31 @@ async function openMuteModal(): Promise<void> {
 	});
 }
 
+/*
+This function assumes:
+- There is a selected conversation.
+- That the conversation already has its conversation menu open.
+
+In other words, you should only use this function within a callback that is provided to `withConversationMenu()`, because `withConversationMenu()` makes sure to have the conversation menu open before executing the callback and closes the conversation menu afterwards.
+*/
+function isSelectedConversationGroup(): boolean {
+	return Boolean(document.querySelector<HTMLElement>(`${conversationMenuSelectorNewDesign} [role=menuitem]:nth-child(4)`));
+}
+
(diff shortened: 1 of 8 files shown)


source/browser.ts
```diff
@@ -550,85 +536,59 @@ async function selectConversation(isNewDesign: boolean, index: number): Promise<
...
-
-In other words, you should only use this function within a callback that is provided to `withConversationMenu()`, because `withConversationMenu()` makes sure to have the conversation menu open before executing the callback and closes the conversation menu afterwards.
-*/
-function isSelectedConversationGroup(isNewDesign: boolean): boolean {
-	const separator = isNewDesign;
-
-	document.querySelector<HTMLElement>(`${conversationMenuSelectorNewDesign} [role=menuitem]:nth-child(4)`);
-
-	return Boolean(separator);
-}
```
dusansimic, line 619: I think this function is broken since https://github.com/sindresorhus/caprine/pull/1879, which I missed back then. It should return a boolean representing the existence of an element, not weather new design is present or not.

```js
return Boolean(document.querySelector<HTMLElement>(`${conversationMenuSelectorNewDesign} [role=menuitem]:nth-child(4)`));
```

This would also influnece `hideSelectedConversation` and `deleteSelectedConversation` functions.

dusansimic (changes requested):

source/browser.ts
lefterisgar: > This would also influnece hideSelectedConversation and deleteSelectedConversation functions.

By the way, those functions don't work as well, but that's unrelated to my changes.

lefterisgar: Is it good now?

dusansimic (approved):

PR #1896, opened 2022-08-25
Add rpm spec

dusansimic:
Adding rpm spec and all other required soruces for building the rpm package. The rpm package needs to be updated before every release.


lefterisgar:
Can this process somehow be automated?

dusansimic:
> Can this process somehow be automated?

I theory, a part of it could be automated but it's generally prefered not to automate writing spec files. For example, version changing and changelog updating could be automated but it's better to do it by hand in case something needs to be changed manually.

lefterisgar (approved):
Great idea to move the script!

Issue #1900, opened 2022-08-30
[Linux] Custom CSS is Ineffective


taylor85345:
custom.css is ineffective. No changes to the file are visible following a reload.

Tested on Arch Linux using both the official package and the AppImage. Also removed the ~/.config/Caprine directory and started from scratch for good measure. Using Caprine in Hyprland (Wayland/wlroots), running both wayland-native and in Xwayland.

EDIT: For the record, I am only modifying the provided variables in the default custom.css. I did attempt loading [someone else's css file](https://gist.github.com/tamasviktor/d8333f8abe695aa165a11e0d61424bbe) for caprine, which was outdated and severely broke elements of the page, but that at least proves that `custom.css` is being loaded and applied, but for some reason the default variables are not taking effect when modified.

---

Caprine 2.56.1
Electron 20.0.3
linux 5.19.3-xanmod1-1-x64v2
Locale: en-US

rosoau:
Exactly. Developers just ignored that useless file and kept providing it. Not nice!
Try this one. It definitely does change it. Perhaps we get some help and elliminate
![Messenger_005](https://user-images.githubusercontent.com/13397926/187780388-dce47947-ff94-4319-900a-61e588ce5679.jpg)
![Messenger_004](https://user-images.githubusercontent.com/13397926/187780403-b7d54768-9836-48e9-8cea-adf3d5df26d0.jpg)

ineffective lines. Pictures show login window and open chat window.
[css-file.zip](https://github.com/sindresorhus/caprine/files/9464941/css-file.zip)

rosoau:
The css file needed to be zipped up

dusansimic:
@taylor85345 that's because those variables were set for the old website design dark mode which is no longer in use. On new design dark mode is available by default which is used now. In order to modify the default dark mode, you'll need to edit the variables from Messengers css files (you can find them in dev tools).

We will remove this code and probably supply an empty file to avoid the confusion.

PR #1903, opened 2022-09-01
Fix custom css default file

dusansimic:
Default file for custom css contained color scheme variables for old design. This caused some confusion which lead to users thinking the custom styles didn't load. I'm proposing a fix which is just removing those variables. Users can create a custom color scheme by copying css variables from Messenger website and putting them in custom css file and modifying them.

Closes: #1900


lefterisgar (approved):
Code looks good to me! Didn't find any issues.

Issue #1909, opened 2022-09-09
Cannot view large image


Jess4da:
when click on image that anyone sent to chat then whole screen disappeared, need to hit `esc` to exit from it, no big image shown

![image](https://user-images.githubusercontent.com/55921466/189293390-d172d895-ce28-4549-ac36-78df9b9a78e1.png)

---

Caprine 2.56.1
Electron 20.0.3
win32 10.0.19044
Locale: en-GB

dusansimic:
Happens due to #1906. New version will have the fix.

tloc241:
man, i have this issue when the app have a new design :(

Invactive:
Same, error still exist. Version 2.56.1

VasilisTachta:
Removing hide-preferences-window class from html tag seems to fix the issue until relaunch.

tolgaerok:
> Removing hide-preferences-window class from html tag seems to fix the issue until relaunch.

how?

tolgaerok:
oh ok found the class by pressing ctl+shift+i.... so where is this html tag stored on the local machine?

dusansimic:
> so where is this html tag stored on the local machine

It's part of a Caprine mechanism for loading saved preferences. This will be fixed by #1910 and it will work fine on the new release. I'd recommend that you don't change the Caprine code in order to remove hide-preferences-window class since it might cause other unwanted issues.

tolgaerok:
> > so where is this html tag stored on the local machine
> 
> It's part of a Caprine mechanism for loading saved preferences. This will be fixed by #1910 and it will work fine on the new release. I'd recommend that you don't change the Caprine code in order to remove hide-preferences-window class since it might cause other unwanted issues.

ok, any idea when update will come out

dusansimic:
> ok, any idea when update will come out

When we (developers) have enough free time to fix it. I can't promise anything now.

If anyone is interested in helping out, feel free to send a pull request with a proposed fix.

Issue #1859, opened 2022-07-18
clicking on an image to see it "full-screen" yields a blank windows until CTRL-R is pressed


jprobichaud:
I have this weird issue, both on my freshly install Windows 11 laptop and my linux (Pop OS/Ubuntu 22.04) laptop.

When in a chat, someone sends me a picture, if I click on it, I get just a blank window (no buttons, no images).  Then I hit CTRL-R and I can see the picture, the messenger icon/button and the the download and forward buttons (and the prev/next arrows).

How can I debug this?

jarnalyrkar:
I have the same issue, using version 2.56 on windows.

TRENT310:
Same issue since 2.53 and still happening as of today's 2.56 release. Darn I was hoping this update would resolve it, but of course there are always changes happening behind the scenes at Meta that are uncoordinated with the development of independent clients like this. 

My workaround for now, in order to load full-screen media attachments is to use Ctrl+, or File > Messenger Settings which for some reason renders the image or video without a full reload.

lefterisgar:
Does it still occur with the latest release (v2.56.1)?

lefterisgar:
> Does it still occur with the latest release (v2.56.1)?

EDIT: You might be using Messenger's new UI which hasn't rolled out to everybody. If so, unless someone from the dev team receives it, we can't debug the issue.

TRENT310:
Okay, 2.56.1 has broken my workaround completely using the Ctrl+, shortcut. I had to roll back to 2.56.0 for now to continue using Ctrl+, to load images.

I cannot control which users have access to certain user interface versions or what switches to trigger it manually or go back to the previous version. FB/Meta has been known to do this in the past even with other features or redesigns.

dusansimic:
@TRENT310 Are you using some different locale from English?

TRENT310:
@dusansimic No, the locale is showing en_US. 

The other clue is that the 'Install Messenger App' nag button is back in whatever this new UI version is, so the previous workarounds to suppress that have changed once again. Interesting that only appears in Caprine and not when accessing messenger.com with Firefox or Chromium.

lefterisgar:
> @dusansimic No, the locale is showing en_US.
> 
> The other clue is that the 'Install Messenger App' nag button is back in whatever this new UI version is, so the previous workarounds to suppress that have changed once again. Interesting that only appears in Caprine and not when accessing messenger.com with Firefox or Chromium.

[...]

PR #1910, opened 2022-09-10
Fix new design user menu

dusansimic:
This patch fixes the query selector for user menu on new design.

Checklist:
- [x] Fix user menu query
- [x] Check for new design
- [x] Clean up selectors

Closes: #1909 
Closes: #1859
Closes: #1918 
Closes: #1923


dusansimic:
So here's a progress update. Right now I'm wrestling with typescript config because some packages that I'm bumping in this pr (for example `element-ready`) have switched to ES module system which is now causing trouble because for some reason, electron is importing our source code as a commonjs module. I'm still trying to figure out how to manage this but if anyone has any info or some examples, help is very welcome.

dusansimic:
One solution that seems to work for some people is splitting the build process for main electron code, preloaded electron code and renderer code. All of those parts are build with vite and tsc ([repo](https://github.com/cawa-93/vite-electron-builder)). It's worth looking into since it would clean up the code structure a bit and it might be a bit easier to maintain. That would however require adding a few extra steps to the build process.

dusansimic:
A short term solution could be an user set preference for the new desing (if users have the new design, they can just chech the option in advanced settings in menu bar). It would require user intervention but it would fix the issue for now without refactoring the code to vite + tsc.

lefterisgar (approved):
It seems to fix said issues. Looks good to me!

dusansimic:
@lefterisgar if you still have access to the old sidebar design, please test weather checking for new design works. It should work okay with openning and closing the preferences and it should remove the `hide-preferences-window` class once it's done.

lefterisgar:
> @lefterisgar if you still have access to the old sidebar design, please test weather checking for new design works. It should work okay with openning and closing the preferences and it should remove the `hide-preferences-window` class once it's done.

Unfortunately I don't have access to it anymore. I think it has rolled out to most users out there.

lefterisgar:
Is this ready to merge?

dusansimic:
> Is this ready to merge?

Essentially yes, I'd just like to complete the last task from the list in this pr. I'll see if I can do it today.

PR #1933, opened 2022-10-13
Fix attachment and receiver pill background in dark mode

### LINKED ISSUE #1932: Attachment indicator broken color in dark mode
Screenshot from 2022-10-12 08-43-51

It should probably have a dark background so it would blend in a bit more.

sfrunza13:
…priate color for the time being

Closes #1932

@dusansimic 

image
image

### DIFF
--- a/css/dark-mode.css
+++ b/css/dark-mode.css
@@ -206,7 +206,7 @@ html.dark-mode .x14hiurz {
 
 /* Attachement Background Color in Chat Box */
 /* TODO: Remove when fixed by fb, since #d6d9dd is too bright here */
-/* Hover overlay is same style as add attachment button */
+/* Current Background-Color is same style as add attachment button */
 html.dark-mode .x1ybostu {
 	background-color: rgba(134,142,153,.25);
 }
--- a/css/dark-mode.css
+++ b/css/dark-mode.css
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
 html.dark-mode .x14hiurz {
 	background-color: transparent !important;
 }
+
+/* Attachement Background Color in Chat Box*/
+/* TODO: Remove when fixed by fb, since --always-white is not good here */
+.__fb-dark-mode .x1ybostu {
+    background-color: #65676b;
+}
+
+/* FB has these "To:" tags in shades of blue */
+/* TODO: Remove when fixed by fb, since --always-white is not good here */
+.__fb-dark-mode .x4noic3{
+	background-color: transparent;
+}
--- a/css/dark-mode.css
+++ b/css/dark-mode.css
@@ -15,7 +15,6 @@
 	--list-header-color: #222;
 	--blue: #0084ff;
 	--white: #fff;
-	--hover-overlay: rgba(255, 255, 255, 0.1);
 }
 
 html.dark-mode body {
@@ -205,16 +204,16 @@ html.dark-mode .x14hiurz {
 	background-color: transparent !important;
 }
 
-/* Attachement Background Color in Chat Box*/
+/* Attachement Background Color in Chat Box */
 /* TODO: Remove when fixed by fb, since #d6d9dd is too bright here */
 /* Hover overlay is same style as add attachment button */
 html.dark-mode .x1ybostu {
-    background-color: var(--hover-overlay);
+	background-color: rgba(134,142,153,.25);
 }
 
-/* Recipient Pills in new message form*/
+/* Recipient Pills in new message form */
 /* TODO: Remove when fixed by fb, since #f3f3f4 is too bright here */
 /* It seems FB uses shades of blue here */
-html.dark-mode .x4noic3{
-	background-color: transparent;
+html.dark-mode .x4noic3 {
+	background-color: var(--secondary-button-background);
 }
--- a/css/dark-mode.css
+++ b/css/dark-mode.css
@@ -208,7 +208,7 @@ html.dark-mode .x14hiurz {
 /* TODO: Remove when fixed by fb, since #d6d9dd is too bright here */
 /* Current Background-Color is same style as add attachment button */
 html.dark-mode .x1ybostu {
-	background-color: rgba(134,142,153,.25);
+	background-color: rgba(134 142 153 / 25%);
 }
 
 /* Recipient Pills in new message form */
--- a/css/dark-mode.css
+++ b/css/dark-mode.css
@@ -15,6 +15,7 @@
 	--list-header-color: #222;
 	--blue: #0084ff;
 	--white: #fff;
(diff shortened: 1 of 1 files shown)


css/dark-mode.css
```diff
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
 html.dark-mode .x14hiurz {
 	background-color: transparent !important;
 }
+
+/* Attachement Background Color in Chat Box*/
+/* TODO: Remove when fixed by fb, since --always-white is not good here */
```
dusansimic, line 208: This comment should reflect what is causing the white background issue. I've just checked and it's not the `--always-white` variable, rather a custom white color on `x1ybostu`  class.

css/dark-mode.css
```diff
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
...
+.__fb-dark-mode .x1ybostu {
+    background-color: #65676b;
```
dusansimic, line 210: Prefer using variables over hardcoded colors. Check what background color/variable is used for the add attachment button on the left.

css/dark-mode.css
```diff
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
 html.dark-mode .x14hiurz {
 	background-color: transparent !important;
 }
+
+/* Attachement Background Color in Chat Box*/
+/* TODO: Remove when fixed by fb, since --always-white is not good here */
+.__fb-dark-mode .x1ybostu {
```
dusansimic, line 209:
```suggestion
html.dark-mode .x1ybostu {
```

css/dark-mode.css
```diff
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
...
+    background-color: #65676b;
+}
+
+/* FB has these "To:" tags in shades of blue */
+/* TODO: Remove when fixed by fb, since --always-white is not good here */
```
dusansimic, line 214: This element also seems to be broken by a hardcoded color in the `x4noic3` class so this comment is not correct.

css/dark-mode.css
```diff
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
...
 	background-color: transparent !important;
 }
+
+/* Attachement Background Color in Chat Box*/
+/* TODO: Remove when fixed by fb, since --always-white is not good here */
+.__fb-dark-mode .x1ybostu {
+    background-color: #65676b;
+}
+
+/* FB has these "To:" tags in shades of blue */
```
dusansimic, line 213: The first comment line should describe what element is the following css patching. I'd recommend to put here someething like "Recipient pill in New message form".

css/dark-mode.css
```diff
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
...
+/* TODO: Remove when fixed by fb, since --always-white is not good here */
+.__fb-dark-mode .x4noic3{
```
dusansimic, line 215:
```suggestion
html.dark-mode .x4noic3{
```

css/dark-mode.css
```diff
@@ -203,3 +203,15 @@ html.dark-mode ._9xq0 {
...
+	background-color: transparent;
+}
```
dusansimic, line 217: Add a new line to the end of file.

dusansimic (changes requested):

sfrunza13:
@dusansimic I hope this time it's better, let me know :D

css/dark-mode.css
```diff
@@ -203,3 +204,17 @@ html.dark-mode ._9xq0 {
 html.dark-mode .x14hiurz {
 	background-color: transparent !important;
 }
+
+/* Attachement Background Color in Chat Box*/
+/* TODO: Remove when fixed by fb, since #d6d9dd is too bright here */
+/* Hover overlay is same style as add attachment button */
+html.dark-mode .x1ybostu {
+    background-color: var(--hover-overlay);
```
dusansimic, line 212: Somehow this one still looks a bit off. There's a slight difference between add attachment button and this one. I'm not sure why. IMHO, it's fine for now but if we're going to leave it as is, there should be another todo comment about this.

css/dark-mode.css
```diff
@@ -15,6 +15,7 @@
 	--list-header-color: #222;
 	--blue: #0084ff;
 	--white: #fff;
+	--hover-overlay: rgba(255, 255, 255, 0.1);
```
dusansimic, line 18: This is already defined by Facebooks css variables. There is no need to redefine it.
```suggestion
```

css/dark-mode.css
```diff
@@ -203,3 +204,17 @@ html.dark-mode ._9xq0 {
 html.dark-mode .x14hiurz {
 	background-color: transparent !important;
 }
+
+/* Attachement Background Color in Chat Box*/
```
dusansimic, line 208:
```suggestion
/* Attachement Background Color in Chat Box */
```

css/dark-mode.css
```diff
@@ -203,3 +204,17 @@ html.dark-mode ._9xq0 {
...
+/* TODO: Remove when fixed by fb, since #d6d9dd is too bright here */
+/* Hover overlay is same style as add attachment button */
+html.dark-mode .x1ybostu {
+    background-color: var(--hover-overlay);
+}
+
+/* Recipient Pills in new message form*/
+/* TODO: Remove when fixed by fb, since #f3f3f4 is too bright here */
+/* It seems FB uses shades of blue here */
+html.dark-mode .x4noic3{
```
dusansimic, line 218:
```suggestion
/* Recipient Pills in new message form */
/* TODO: Remove when fixed by fb, since #f3f3f4 is too bright here */
/* It seems FB uses shades of blue here */
html.dark-mode .x4noic3 {
```

css/dark-mode.css
```diff
@@ -203,3 +204,17 @@ html.dark-mode ._9xq0 {
...
+	background-color: transparent;
```
dusansimic, line 219: This one looks better with secondary button background color imho. You can then clearly see the pill shape.
```suggestion
	background-color: var(--secondary-button-background);
```

dusansimic (changes requested):

dusansimic:
Just a few more changes and I think it's ready to go :)

css/dark-mode.css
sfrunza13: I agree, I found that I was looking at the wrong element for reference. New color should be matching.

dusansimic: It is! Fantastic work 🚀

dusansimic (approved):

dusansimic:
Thanks for this great work! 😁

PR #1938, opened 2022-10-15
Fix "Join room" background button on dark theme

jrafaelsantana:
This patch fixes the background color of "Join room" notification button on dark theme.

Before:
image

After:
image


css/dark-mode.css
```diff
@@ -217,3 +217,8 @@ html.dark-mode .x1ybostu {
 html.dark-mode .x4noic3 {
 	background-color: var(--secondary-button-background);
 }
+
+/* Invite notification: join room */
+html.dark-mode .x1qhmfi1 {
+	background-color: var(--container-color);
```
dusansimic, line 223: What about secondary button background color? It might look better in this case. The darker button on a lighter widget which is itself on a darker background seems kind of out of place.

```suggestion
	background-color: var(--secondary-button-background);
```

dusansimic (changes requested):
Also add a new line to the end of file.

jrafaelsantana:
Done!

Could you tag this PR with the hacktoberfest-accepted label?

Thank you

dusansimic:
There was an issue with the variable I recommended, it was overwritten by inline css on the frontend (bug in the website, not our fault). Once it was reset to regular variable value it works as expected. This however means you didn't check weather it works after applying the patch I recommended 😁 but it's fine, the important thing is that it's fixed.

dusansimic (approved):

jrafaelsantana:
Even after your fix, the background still the wrong color.

Before

I made a fix based on your commit. Just fixed the css selectors.

image

Issue #1936, opened 2022-10-15
Vote button wrong color in dark mode


dusansimic:
Vote button in chat has light background in dark mode.

Anuran12:
Hi. @dusansimic  i I want to fix this.

dusansimic:
Feel free to work on this :). Send a pull request once you've made a working version. Take a look at some of past pull requests (#1933 and #1931 for example) as an inspiration how to fix the issue.

Anuran12:
@dusansimic  Okay thanks

Anuran12:
@dusansimic  hey.. please help. where can i find the class name of the vote button ?

dusansimic:
@Anuran12 you can find it in the dev tools (open them by pressing F12). You can use the inspect element tool to easily find the element you're looking for in the html tree and from there you can look for the specific style which is causing the issue. Once you find the style, just check which class applies that style and you found the class name you need to patch.

Anuran12:
@dusansimic  Is it looking good ?

![image](https://user-images.githubusercontent.com/104062485/196019121-9e4cbed5-b033-4dbe-99f7-a44ff621b61b.png)

dusansimic:
On light mode, the light color is used which in case of light mode is a secondary color. Blue is the primary color. I think in this cases it would be better to use a dark color. There is one available in the variable `--secondary-button-background`. You can use that css variable for the background of this button.

Also, feel free to open a pull request so I can see the progress and changes in the stylesheets :)

Anuran12:
@dusansimic  Hey.. `--secondary-button-background` is not working.

Anuran12:
@dusansimic 

![image](https://user-images.githubusercontent.com/104062485/196033865-e9d12d41-9cdb-4ca9-a5a9-8358604dda10.png)

The variable is not mentioned there.

Anuran12:
So Only `rgba()` is working..

![image](https://user-images.githubusercontent.com/104062485/196033948-1d825be2-9eac-4ab2-99da-fd3e4c5fcbf2.png)

![image](https://user-images.githubusercontent.com/104062485/196033962-5120c6bc-6e73-4ff8-a4f0-51472bcf7af5.png)

dusansimic:
> The variable is not mentioned there.

The variable is defined by Messenger website in `__fb-dark-mode` class.

[...]

PR #1943, opened 2022-10-16
fix Dark mode color on vote button

Anuran12:
#1936 

@dusansimic

### DIFF
--- a/css/dark-mode.css
+++ b/css/dark-mode.css
@@ -232,3 +232,7 @@ html.dark-mode .xxymvpz.x6ikm8r.x10wlt62.x5yr21d.x230xth.xmjcpbm.x168nmei.x13lgx
 html.dark-mode .x1qhmfi1 {
 	background-color: var(--secondary-button-background);
 }
+
+html.dark-mode .x1qhmfi1 {
+	background-color: rgba(134 142 153 / 25%);
+}


css/dark-mode.css
```diff
@@ -225,3 +225,7 @@ html.dark-mode .xxymvpz.x6ikm8r.x10wlt62.x5yr21d.x230xth.xmjcpbm.x168nmei.x13lgx
 html.dark-mode .x1qhmfi1 {
 	background-color: var(--secondary-button-background);
 }
+
+html.dark-mode .x1qhmfi1 {
```
dusansimic, line 229: This class name is already redefined directly above. In this case you shouldn't hardcode the color, rather patch some other class which is causing the issue. Secondary button background variable is using `rgba(255,255,255,.1)` so that would mean that it's transparent to some degree and there is another element whose background color is light and is causing the button to be light. You'll need to find that element and redefine it's background color.

dusansimic (changes requested):

css/dark-mode.css
Anuran12: ohh okay. got it.

Anuran12:
@dusansimic  Sorry I am unable to find that :-(

dusansimic:
That's fine :) I'll close this pr so someone else can work on it.

PR #2002, opened 2023-04-05
Minor fixes

Alex313031:
Just some minor fixes, and adding nvm and npm rc files to ease local development on new or old linux distros.


sindresorhus:
Thanks for the suggestion, but none of these changes are required or wanted.

Issue #1982, opened 2023-01-19
Links opening in some kind of internal browser?


maetthew:
Hi,

I'm not sure when this started happening. But lately when I'm clicking links in a chat they open in some kind of internal browser of Caprine's?

I can't find a setting to change this and have to copy paste links into my browser. Is this behaviour intended?

I'm on Arch Linux, Caprine 2.57.0. Not sure what other information needs to be provided but happy to provide if needed.

dusansimic:
This is very unusual. I don't experience this behavior (I also use Arch Linux). What you should experience is link being opened in the default browser. Check if you have a default browser set.

Also, could you send here an example of a link that is misbehaving? I'd like to see if it's related to the specific link.

maetthew:
All the links that I've clicked last couple of days are opening in this 'internal' browser for some reason. All links from all other apps (Telegram, Discord, Thunderbird etc) are opening in my browser as normal.

I'm using KDE and my browser is set in "Default Applications" under "System Settings", also my `BROWSER` environment variable is correctly setup (and has been for at least a year).

maetthew:
I've also tried deleting settings and app data, to no avail. Maybe I should try to reinstall completely. I'm experiencing other weird behaviours like a lot of the menu items seems to have no effect what so ever, like none of the options under "Conversation" works. Under "View" I can't change theme or change sidebar settings for instance.

maetthew:
FWIW I uninstalled Caprine, and nuked the config folder, with no effect.

dusansimic:
Facebook usually partially rolls out new things on the messenger website so that's why I didn't have the same behavior. Now it happens on my account too.

dusansimic:
I'm now able to reproduce this behavior. I can't guarantee when it's going to be fixed but help is welcome.

SlouchyButton:
This is the default behavior of electron, and is probably not related to Facebook Messenger in any way.
This might be broken because Caprine uses system electron instead of prepackaging one and is not tied to a specific version, but rather uses the default electron on system (most likely the most recent one). There was a breaking API change in electron, that deals with how to deal with new window opens. I did not check how Caprine handles this, but I encountered this while doing my own Messenger app, because most tutorials use old API.

https://www.electronjs.org/docs/latest/breaking-changes#removed-webcontents-new-window-event

Correct way of handling of link open (better said new window open event) is like this now:
```
win.webContents.setWindowOpenHandler((details) => {
[...]

PR #2007, opened 2023-04-29
Fix open links issue

dusansimic:
Closes: #1982

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -505,7 +505,7 @@ function createMainWindow(): BrowserWindow {
 		}
 	});
 
-	webContents.setWindowOpenHandler((details) => {
+	webContents.setWindowOpenHandler(details => {
 		if (details.disposition === 'new-window') {
 			if (details.url === 'about:blank' || details.url === 'about:blank#blocked') {
 				if (details.frameName !== 'about:blank') {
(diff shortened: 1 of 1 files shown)


source/index.ts
```diff
@@ -505,7 +505,7 @@ function createMainWindow(): BrowserWindow {
 		}
 	});
 
-	webContents.setWindowOpenHandler((details) => {
+	webContents.setWindowOpenHandler(details => {
 		if (details.disposition === 'new-window') {
```
SlouchyButton, line 509: We can not use 'new-window' here, because if a user clicks on a link, it's not always a new window. For me, debugger says that value of disposition is 'foreground-tab'.
This relates to issue 1982 not being solved after this merge.

If I remove this if, it starts working. I have no idea when would messenger open a window that we do not want to open in an external browser, but I'd expect it would be related to the check that was present even before this fix - Voice/video call popup. I feel like that check should still work. I have tested it, and if we comment out the if it works and calling also works and opens in new electron window.

SlouchyButton (commented):
Regarding the issue #1982 that is not fixed after this PR, we can not check for 'new-window' disposition. It is not true in my case (and apparently in other cases as well)

source/index.ts
dusansimic: Seems like I managed to click on the link that opens a new window when I tested this 😅. I'll push out a new release with a well tested fix.

PR #2012, opened 2023-04-30
Update a few things before making legacy branch

Alex313031:


dusansimic:
Almost exactly same changes were rejected in #2002. Those changes are not needed for the current development, build and distribution workflow. If you think they are needed or could somehow contribute to the current processes, those changes could be discussed in an issue.

PR #2013, opened 2023-04-30
Workflow changes

Alex313031:
Add npm and nvmrc files. Useful for linux and macos. For example I have to use it on ubuntu 22.04 which only comes with node 14.

Update repository URL.
Add appName as per electron builder documentation.

Add rimraf to facilitate cross platform usage of the clean and distclean scripts

Make clean, distclean, and dist scripts to make building/testing/cleaning artifacts easier.


.npmrc
```diff
@@ -0,0 +1,2 @@
+target=v18.16.0
+node-version=18.16.0
```
dusansimic, line 2: Why is this needed? I can't find any docs on the `target` property. Also why is there a need to set `node-version` property?

.nvmrc
```diff
@@ -0,0 +1 @@
+18.16.0
```
dusansimic, line 1: It's better to use an alias instead of a specific version. There is no need to pin a specific point release. I think v18 is `hydrogen`.

```suggestion
lts/hydrogen
```

package.json
```diff
@@ -1,10 +1,11 @@
 {
 	"name": "caprine",
+	"appName": "caprine",
```
dusansimic, line 3: Could you link said docs? I can't find anything about this property on [electron.build](https://www.electron.build/).

package.json
```diff
@@ -1,10 +1,11 @@
...
 	"productName": "Caprine",
 	"version": "2.57.3",
 	"description": "Elegant Facebook Messenger desktop app",
 	"license": "MIT",
-	"repository": "sindresorhus/caprine",
+	"repository": "https://github.com/sindresorhus/caprine",
```
dusansimic, line 8: That's not needed. `package.json` supports the [shortcut syntax](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository).
```suggestion
	"repository": "sindresorhus/caprine",
```

package.json
```diff
@@ -19,7 +20,10 @@
 		"lint": "tsc && xo && stylelint \"css/**/*.css\"",
 		"test": "npm run lint",
 		"start": "tsc && electron .",
-		"build": "tsc",
+		"build": "tsc && electron-builder --dir",
```
dusansimic, line 23: This would break our GitHub pipelines ([action-electron-builder](https://github.com/marketplace/actions/electron-builder-action)). The build script should only contain commands for compiling TS code, not for building the app.
```suggestion
		"build": "tsc",
```

package.json
```diff
@@ -19,7 +20,10 @@
...
+		"clean": "rimraf -v node_modules",
+		"distclean": "rimraf -v node_modules && rimraf -v dist",
```
dusansimic, line 25: Those directories don't get removed that often, there is no need for a script to do that.

package.json
```diff
@@ -19,7 +20,10 @@
...
+		"dist": "electron-builder",
```
dusansimic, line 26: We use our GitHub pipelines for that. There is no need for a developer to do those tasks. We could add pipelines on every commit tho, that would allow us to test artifact build on every commit. That should however be discussed in an issue.
```suggestion
```

package.json
```diff
@@ -52,6 +56,7 @@
 		"husky": "^8.0.1",
 		"np": "^7.6.2",
 		"patch-package": "^6.4.7",
+		"rimraf": "^4.4.1",
```
dusansimic, line 59:
```suggestion
```

dusansimic (changes requested):

.npmrc
Alex313031: @dusansimic When running node gyp under a custom npm (like if you compiled/installed node or npm manually) this tells it what version to target for

.nvmrc
Alex313031: I didnt know that, thanks.

package.json
Alex313031: I know, but it can be good for if a user wants to copy/paste the url or for some node extenders if say, someone wanted to grep the repo url and use it for some reason. npm supports both so I have always made it a habit to just go ahead and use the full url

package.json
Alex313031: Im sorry, not for electron build, but electron notarize

package.json
Alex313031: Ahh ok. Will remove

package.json
Alex313031: Useful for local development though. Why not have it?

package.json
Alex313031: Yeah saw there is already del-cli

package.json
Alex313031: Like above comment, useful for local development, so why not have it. It would only be run by a developer wanting to make their own packages like me lol

.npmrc
dusansimic: When is this the case? Doesn't seem like something that happens quite often.

package.json
dusansimic: Docs?

package.json
Alex313031: @dusansimic OK made some new changes, including updating deps. All the versions have been updated to versions that npm would have installed anyway, with the exception of electron-context-menu which was updated to 3.6.1 instead of 3.5.0 (which is what npm would have installed) because theres no syntax changes and should still work (everything works on my end, I just made a fresh deb and installed it and checked the right click menus)

package.json
dusansimic: I still don't really think this needs to change. If a user wants to copy/paste a url of a repo while looking at a package.json file, I think they can manage with just the user/repo slug while we get a clean package.json 😁

package.json
Alex313031: @dusansimic Looks like they removed this in favor of just using "name": I will remove

Alex313031: anyway the docs are at https://github.com/electron/notarize electron-notarize is a subdep of electron-builder to sign macos apps

dusansimic: > anyway the docs are at https://github.com/electron/notarize electron-notarize is a subdep of electron-builder to sign macos apps

I figured out the the GitHub repo, just couldn't find the property in question so that's why I asked for docs :)

.npmrc
Alex313031: @dusansimic Not something that would happen often, but if it DOES happen, its a b**** to figure out what the issue is, as was my delemma when working on atom-ng > https://github.com/Alex313031/atom-ng

BTW atom-ng is my continued fork of atom since micro$oft acquired github and they deprecated atom. It has many updates, including the node and electron version, if you wanted to check it out or are/were an atom user

package.json
dusansimic: Since we don't use it as part of a development workflow, there is no need for it to be a script.

package.json
dusansimic: > All the versions have been updated to versions that npm would have installed anyway

Don't do that in this PR. Make a separate one and explain what dependencies are bumped, why and what could be side-effects if any. Caprine is quite fragile when it comes to dependencies since some of them switched to esm and we need to be careful when bumping them.

.npmrc
dusansimic: This might be useful for legacy branch but not for main.

package.json
Alex313031: OK

package.json
Alex313031: ok I can remove

package.json
Alex313031: There, I reverted stuff, which basically makes this PR what would have been my second PR. All this does now is updates dep versions. I reverted the electron-context-menu to 3.5.0 since you are worried.

TLDR: Now all this does is bump revisions to the revisions that npm would have installed anyway. So it will not change the behavior of caprine at all. Just a nice cleanup

dusansimic: > Now all this does is bump revisions to the revisions that npm would have installed anyway

Why is it needed then? Don't make changes that are not needed

dusansimic: And as I've already said, this pr is for workflow changes (as the title suggests). For other things, create a new pr

Alex313031: @dusansimic I think its good to keep versions in package json up to date with the versions actually installed, so that when working on stuff or doing bisection, you can go look up the docs for that specific version to see what has changed

dusansimic:
I'm closing this pr since it's getting out of hand. If you want to contribute something, do so in an organized manner. Not multiple unrelated changes in one pr, some of them not even discussed or suggested beforehand while being undocumented. As I've already stated, I'm willing to accept contributions in order to support Windows 7 as long as Electron supports it but I won't be approving pr-s like this one.

Issue #1857, opened 2022-07-13
macOS Ventura beta: window is unmovable and title bar is broken.


apotenza92:
Traffic lights are overlapping profile picture, entire window is unmovable. It's like top area is not registering as a draggable element.
<img width="996" alt="image" src="https://user-images.githubusercontent.com/53612685/178653048-7332969a-8e7a-4938-a32e-27a3674b1169.png">

sheldonhull:
- Related: #410
- Not sure it's specific to Ventura, as this is an issue on 12.5 (21G72) (Monterey).

I'm on 2.55.6 and seems to have the same issue. 
I can only drag if the window is not focused, but as soon as focused there is no drag region.

relink2013:
I can confirm this on 12.5 as well.

andrewbrereton:
This is still an issue on macOS Monetery 12.6.2.

<img width="171" alt="image" src="https://user-images.githubusercontent.com/682893/208218607-658741a3-dde1-481d-b368-e5ad3af4db9f.png">

Issue #1488, opened 2020-12-17
Re-enable window dragging (macOS)


colineberly:
#### *I decided to make this a dedicated post so as to get visibility to anyone that comes across the issue looking for a simple fix. Hope this is OK. This is in reference to #1471.*

---

If you are having an issue dragging your window in the **macOS** version, you can add the following CSS fix. This tweak enables draggable areas in the app window.

I've added 2 draggable regions, one being the top area of the sidebar above the search box, and the other area is the header in the main chat window, where the username is along with call and info buttons. I've also thrown in a couple positioning tweaks to the top area of the sidebar as well as removed search in collapsed sidebar view (under 900px). **_Note: In collapsed view, sidebar area is no longer a draggable region._**

![mockup-messenger](https://user-images.githubusercontent.com/61343/102463175-ac706b80-4042-11eb-8cd7-e1378275fc2e.png)

#### Place the following into the top of your `custom.css` file:

```css
.rq0escxv.l9j0dhe7.du4w35lb.j83agx80.pfnyh3mw.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.jei6r52m.wkznzc2l.n851cfcs.dhix69tm.ahb00how,
.rq0escxv.l9j0dhe7.du4w35lb.j83agx80.pfnyh3mw.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.hv4rvrfc.dati1w0a.f10w8fjw.pybr56ya.b5q2rw42.lq239pai.mysgfdmx.hddg9phg {
		-webkit-app-region: drag !important;
}

.rq0escxv.l9j0dhe7.du4w35lb.j83agx80.pfnyh3mw.i1fnvgqd.bp9cbjyn.owycx6da.btwxx1t3.jei6r52m.wkznzc2l.n851cfcs.dhix69tm.ahb00how {
		margin: 0 !important;
		padding: 12px 16px 0 !important;
}

.rq0escxv.l9j0dhe7.du4w35lb.j83agx80.cbu4d94t.g5gj957u.d2edcug0.hpfvmrgz.kud993qy.buofh1pr {
    margin-top: 24px !important;
}

@media (max-width: 900px) {
	.rq0escxv.l9j0dhe7.du4w35lb.n851cfcs.aahdfvyu {
	    display: none !important;
	}
[...]

PR #2014, opened 2023-05-01
Fix macOS ui

dusansimic:
TODO:
- [x] Fix traffic light buttons position
- [x] Fix draggable area
- [ ] Move page title ("Chats") when maximised
- [ ] Hide search input when window is small

Screenshots:
image

Closes: #1488 
Closes: #1857


dusansimic:
The rest of the issues will be fixed later. For now we need to push a fix for macOS users.

PR #2017, opened 2023-05-01
Add nvmrc file

Alex313031:
Here you go sir.


Alex313031:
@dusansimic That was to make a "legacy" branch, not to be merged into main. The nvmrc should  have hydrogen for the main branch, not gallium.

This is because the legacy branch will use electron 22 and node 16, while the main branch uses electron 24 and node 18.

dusansimic:
The pr was opened to the main branch, not legacy. Legacy branch does not yet exist. I've asked you if you'd like to contribute to that branch to send me an email (it's available on my profile page) so we could plan it out.

Alex313031:
@dusansimic Oh you're right, sorry. I will email you.

PR #2019, opened 2023-05-08
Updates for main

Alex313031:
This adds more icon sizes for linux, and adds a new windows icon constant.

Right now, our windows builds have no icon. The icon is generated by electron builder from the Icon.png.

This is problematic because only will the icon be set if someone is using the .exe installer. Using a .zip or building from source on Linux makes it use the default electron icon.

Another problem is that the icon generated only has one size, 512px.
A proper windows icon (and proper windows electron app) should use a windows .ico file that has 16px, 32px, 48px, 64px, and 256px embedded inside.

This also adds a new menu item to the debug menu, opening a chrome://gpu window to check hardware acceleration status. It also checks the hardwareacceleration user config, and appends ignore-gpu-blocklist to electron if it is not false.

Other changes include adding my name to the About window, and adding QUIC support, as many facebook and google domains use this, and it enables faster handshaking.


dusansimic:
Here are some comments but I will open reviews for a few pieces of code.

> This adds more icon sizes for linux

Why?

> This is problematic because only will the icon be set if someone is using the .exe installer. Using a .zip or building from source on Linux makes it use the default electron icon.

We only provide exe installer. This is not an issue then.

> Another problem is that the icon generated only has one size, 512px.

Okay.

> Other changes include adding my name to the About window,

Your name will be present on the win7 variant.

css/browser.css
```diff
@@ -28,12 +28,18 @@ body {
...
 }
+[role='navigation'] .x16n37ib {
+	-webkit-app-region: no-drag;
+}
+.x9f619.x1n2onr6.x1ja2u2z.x78zum5.xdt5ytf.x193iq5w.xeuugli.x1r8uery.x1iyjqo2.xs83m0k.xsyo7zv.x16hj40l.x10b6aqq.x1yrsyyn {
+	-webkit-app-region: drag;
+}
+[role='navigation'] .x1heor9g.x1qlqyl8.x1pd3egz.x1a2a7pz {
+	margin-left: 60px;
+}
```
dusansimic, line 42: I this is for macOS, it's already fixed https://github.com/sindresorhus/caprine/pull/2014

package.json
```diff
@@ -102,14 +102,22 @@
...
-			"electronUpdaterCompatibility": ">=2.15.0",
+			"electronUpdaterCompatibility": ">=4.5.2",
 			"darkModeSupport": true,
+			"target": {
+				"target": "default",
+				"arch": [
+					"x64",
+					"arm64"
+				]
+			},
```
dusansimic, line 120: Already added https://github.com/sindresorhus/caprine/pull/2015

readme.md
```diff
@@ -47,7 +47,7 @@
 
 ## Install
 
-*macOS 10.10+, Linux, and Windows 10+ are supported (64-bit only).*
+*macOS 10.12+ (Intel and Apple Silicon), Linux (x64 and arm64), and Windows 10+ (64-bit) are supported.*
```
dusansimic, line 50: Already added https://github.com/sindresorhus/caprine/pull/2015

source/index.ts
```diff
@@ -59,6 +59,14 @@ if (!config.get('hardwareAcceleration')) {
 	app.disableHardwareAcceleration();
 }
 
+if (config.get('hardwareAcceleration')) {
```
dusansimic, line 62: If you're reading config of the app, the property should be defined in [source/config.ts](https://github.com/sindresorhus/caprine/blob/main/source/config.ts) and also you should define migrations so the property gets created after an update.

source/index.ts
```diff
@@ -288,11 +296,15 @@ function createMainWindow(): BrowserWindow {
...
-		icon: is.linux ? caprineIconPath : undefined,
+		icon: is.linux || is.macos ? caprineIconPath : caprineWinIconPath,
 		minWidth: 400,
 		minHeight: 200,
 		alwaysOnTop: config.get('alwaysOnTop'),
 		titleBarStyle: 'hiddenInset',
+		trafficLightPosition: {
+			x: 80,
+			y: 20,
+		},
```
dusansimic, line 307: Already fixed https://github.com/sindresorhus/caprine/pull/2014

dusansimic (changes requested):

dusansimic:
And please rebase to main so this branch would be up to date.

PR #2024, opened 2023-05-31
Fix macOS icon

### LINKED ISSUE #2018: macOS icon size too large
The icon should have some padding ([Apple docs](https://developer.apple.com/design/human-interface-guidelines/app-icons)).

image

This was done in https://github.com/sindresorhus/caprine/commit/cf78ee346aeb22bbd32e7455d1d5e8df5a0808b2 since the previous icon has some padding which made icon look a lot smaller compared to other ones on other platforms (namely Windows and Linux). A solution would be to use two different icons, one with white background and padding and one without white background and less padding for macOS and other platforms respectively.

BienGudBoy:
Fix #2018


dusansimic:
Will this use two different icons (one for macOS and one for other platforms) as stated in the issue? That is the ultimate goal for this.

BienGudBoy:
> Will this use two different icons (one for macOS and one for other platforms) as stated in the issue? That is the ultimate goal for this.

I changed the image for macOS build only (switching from a static .png image to a dynamic, resizeable .icns from the official Messenger build). Other platforms should be the same.

dusansimic:
The goal is for other platforms to have a different icon than before.

The icon was changed in https://github.com/sindresorhus/caprine/commit/cf78ee346aeb22bbd32e7455d1d5e8df5a0808b2 to have less padding so it wouldn't look small on platforms other than macOS but on macOS it should have padding. My idea was to use the regular messenger icon ([link](https://cdn.svgporn.com/logos/messenger.svg)) for all platforms except macOS and the macOS icon should be like the current one but with more padding.

BienGudBoy:
@dusansimic I'm not sure if the icon for other platforms has the same problem in macOS, because I do not have a Windows or Linux device to test.
This padding issue is macOS specific - as it can't resize the app icon base on the size of the dock.
This is because normally, macOS apps use .icns image format, which contains multiple icons with different dimensions (for example 24x24, 48x48, 72x72, etc.)
Depending on the screen's DPI, or the dock's size, macOS will choose the "good" dimension of the app icon to show on screen - that's how your app icons scale well with your dock's size, or your screen's size.
This obviously won't work with a single, static icon (.png, .jpg whatever). macOS has no choice but to display the whole icon, and as it has no way to resize, it causes the behavior like in your screenshot.

You can try building Caprine with my commit, or replace your current app's icon with the ICNS in my commit. It should work as I tested this before.

dusansimic:
Okay, I've tried it on macOS and Linux and it seems to work as you've said 😄. I didn't find this behavior documented anywhere on Electron Builder website so I didn't know it existed.

Thanks for the contribution!

PR #2041, opened 2023-07-28
Tray icon/indicator Fix

### LINKED ISSUE #1823: Tray Icon - No notification badge
From almost a year badges don't work. I don't understand why no one cares to restore this essential feature?
---

Caprine 2.55.4
Electron 10.4.7
win32 6.3.9600
Locale: pl

### LINKED ISSUE #1825: Notifications and tray indicator are broken
Linux Mint 20.3 Cinnamon 5.2.7 on Linux 5.13.0-40-generic if it helps. I installed using the deb package.

### LINKED ISSUE #1995: Tray/notification icon color
Hi all.
Please, change the color of the notification icon. I don't know anyone who is using a white taskbar, but definitely, more than 70%/80% of users are using gray or just dark taskbar. So please change the color of this icon to white or light blue because it's invisible. Tested on KDE/Mate/Cinnamon. Look at the screenshot from the newest KDE Plasma:

image

On Windows, the situation is a bit better, but also not perfect (see the screenshot):

image

### LINKED ISSUE #2031: Tray icon/indicator Fix
Hello everyone. I was able to find a fix for the issue where the tray icon does not indicate if there are any unread messages(related issues #1995, #1825, #1823). I'm reluctant to open a pull request because I'm a complete newbie to ts/js/electron (I've been working on this for a few days) and I'm not sure about the quality of the code.

From what I understand, there seem to be two problems:

- The observer, which "observes" the sidebar, does not trigger when the conversation list changes.

- The conversation list selector is not receiving the actual list of the conversations.

The commit:
https://github.com/stkrknds/caprine/commit/31c6fec71e544832fbe3a57dab2d9fb7b33a71e9

As I said, I don't know much about web development, so the code is probably lacking in quality. I would appreciate it if someone could take a look at it and suggest improvements.
Thank you very much for your time. 
Unread Icon

stkrknds:
Hello this is a fix for the tray icon/indicator issue. After running the tests, I got an error about putting await inside of a loop. So unfortunately, I had to change the selector of the sidebar and remove the loop.

Partially fixes #1825 (this wouldn't implement notifications).

Fixes #1995
Fixes #1823
Fixes #2031

### DIFF
--- a/source/browser/conversation-list.ts
+++ b/source/browser/conversation-list.ts
@@ -131,7 +131,7 @@ async function createConversationNewDesign(element: HTMLElement): Promise<Conver
 	*/
 
 	conversation.selected = Boolean(element.querySelector('[role=row] [role=link] > div:only-child'));
-	conversation.unread = Boolean(element.querySelector('[aria-label="Mark as Read"]'));
+	conversation.unread = Boolean(element.querySelector(selectors.markAsRead));
 
 	const unparsedLabel = element.querySelector<HTMLElement>('.a8c37x1j.ni8dbmo4.stjgntxs.l9j0dhe7 > span > span')!;
 	conversation.label = await getLabel(unparsedLabel);
@@ -154,7 +154,14 @@ async function createConversationList(): Promise<Conversation[]> {
 		return [];
 	}
 
-	const elements: HTMLElement[] = [...list.children] as HTMLElement[];
+	const convList = list.querySelector('div[class=x1n2onr6]');
+
+	if (!convList) {
+		console.error('No conversations found');
+		return [];
+	}
+
+	const elements: HTMLElement[] = [...convList.children] as HTMLElement[];
 
 	// Remove last element from childer list
 	elements.splice(-1, 1);
@@ -214,7 +221,7 @@ function countUnread(mutationsList: MutationRecord[]): void {
 }
 
 window.addEventListener('load', async () => {
-	const sidebar = await elementReady('[role=navigation]', {stopOnDomReady: false});
+	const sidebar = await elementReady('[class="x78zum5 xdt5ytf x1iyjqo2 xh8yej3"]', {stopOnDomReady: false});
 
 	if (sidebar) {
 		const conversationListObserver = new MutationObserver(async () => sendConversationList());
--- a/source/browser/conversation-list.ts
+++ b/source/browser/conversation-list.ts
@@ -214,9 +214,7 @@ function countUnread(mutationsList: MutationRecord[]): void {
 }
 
 async function updateTrayIcon(): Promise<void> {
-	const chatsIconSelector = selectors.chatsIcon;
-
-	const chatsIcon = await elementReady(chatsIconSelector, {
+	const chatsIcon = await elementReady(selectors.chatsIcon, {
 		stopOnDomReady: false,
 	});
 
--- a/source/browser/conversation-list.ts
+++ b/source/browser/conversation-list.ts
@@ -131,7 +131,7 @@ async function createConversationNewDesign(element: HTMLElement): Promise<Conver
 	*/
 
 	conversation.selected = Boolean(element.querySelector('[role=row] [role=link] > div:only-child'));
-	conversation.unread = Boolean(element.querySelector(selectors.markAsRead));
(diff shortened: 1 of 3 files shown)


source/browser/conversation-list.ts
```diff
@@ -213,12 +213,25 @@ function countUnread(mutationsList: MutationRecord[]): void {
 	}
 }
 
+async function updateTrayIcon(): Promise<void> {
+	const chatsIconSelector = selectors.chatsIcon;
+
+	const chatsIcon = await elementReady(chatsIconSelector, {
```
dusansimic, line 219:
```suggestion
	const chatsIcon = await elementReady(selectors.chatsIcon, {
```

source/browser/conversation-list.ts
```diff
@@ -233,5 +246,12 @@ window.addEventListener('load', async () => {
 			attributes: true,
 			attributeFilter: ['class'],
 		});
+
+		chatsIconObserver.observe(sidebar, {
+			subtree: true,
+			childList: true,
+			attributes: true,
+			attributeFilter: ['class', 'aria-label'],
+		});
```
dusansimic, line 255: This probably won't be very efficient but let's say it's fine for now.

It would be better to monitor the chat icon directly for attribute changes (since that's where you're extracting the number from). That would use far less resources than monitoring the whole chat list.

This could be addressed in a second pr.

source/browser/selectors.ts
```diff
@@ -1,4 +1,5 @@
 export default {
+	chatsIcon: '[class="x1i10hfl xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx x2lwn1j xeuugli x4uap5 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x87ps6o x1lku1pv x1a2a7pz x6s0dn4 x1q0g3np xn3w4p2 x1nn3v0j x1120s5i x1av1boa x1lq5wgf xgqcy7u x30kzoy x9jhf4c xdj266r x11i5rnm xat24cr x1mh8g0r x78zum5"]',
```
dusansimic, line 2: Add a comment that would suggest to check this selector if the tray icon breaks.

```suggestion
	chatsIcon: '[class="x1i10hfl xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx x2lwn1j xeuugli x4uap5 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x87ps6o x1lku1pv x1a2a7pz x6s0dn4 x1q0g3np xn3w4p2 x1nn3v0j x1120s5i x1av1boa x1lq5wgf xgqcy7u x30kzoy x9jhf4c xdj266r x11i5rnm xat24cr x1mh8g0r x78zum5"]', # ! Tray icon dependency
```

dusansimic (changes requested):

dusansimic:
Thanks @stkrknds for the contribution! I've suggested some changes so once they're implemented, we can merge this.

source/browser/conversation-list.ts
stkrknds: Hello @dusansimic. This is a simple change, do you want me to change that here instead of opening a new pr?

stkrknds:
I have made some changes, so that only the left sidebar is observed for the chats icon. I think that's the best that can be done in terms of efficiency. The only downside is that there is now an additional selector(for the left sidebar) which may break in the future.

dusansimic:
> The only downside is that there is now an additional selector(for the left sidebar) which may break in the future.

That is true, but I think it's better to have a new selector that can be fixed than having one more mutation observer that checks a lot of elements on the page for changes. That would consume more resources and in turn make Caprine slower and more resource hungry which is already an issue. I think the path forward is either optimizing existing observer or removing them altogether.

dusansimic (approved):

Issue #2033, opened 2023-07-21
carpine locks on all new messages


cobra3282000:
---

Caprine 2.57.4
Electron 24.1.3
linux 6.4.3-060403-generic
Locale: en-US

every time i get a message on caprine it freezes and locks.   every time.   cant do anything really except restart it.  not sure what is going on.   i started carpine terminal to see error messages, this is what i get.   dmeg shows no errors.  

checkForUpdatesAndNotify called, downloadPromise is null
(node:6684) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of null (reading 'checked')
    at /opt/Caprine/resources/app.asar/dist-js/browser.js:208:34
    at async EventEmitter.listener (/opt/Caprine/resources/app.asar/node_modules/electron-better-ipc/source/renderer.js:46:34)
(Use `caprine --trace-warnings ...` to show where the warning was created)
(node:6684) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)

JoelFeiner-viimed:
I'm getting this same error now on startup. Caprine no longer loads anything and freezes immediately. I cannot use it.

JoelFeiner:
I'm getting this same issue. It happens on startup and it will not load new messages since Friday. It also freezes and never recovers. I can't use Caprine at all.

Also Caprine 2.57.4 via Snap on Ubuntu 23.04 (6.2.0-25-generic).

cobra3282000:
mine does that on startup too, and some times it reloads and works, 99 percent of the time though i get a message and boom its locked.    but the menu bar still works, but settings wont open either.   i have tried on a few distro's, and same problem.  go figure, i also use linux mint 21.2 xfce and mate.

gnar03:
Same issue here as of July 26. It worked fine before, now can't get it to work anymore at all.

shauneccles:
Same issue on Win64 with 2.58.0

cobra3282000:
my issues have changed now, now dark mode is not working, but the messenger part is working again, and cant make any changes to settings.    under linux, does any know where the settings are stored at ?   what directory ?   maybe manually edit the settings file ?

[...]

PR #2051, opened 2023-08-17
Fix notifications workaround

### LINKED ISSUE #1562: Notifications not working on new design

### LINKED ISSUE #1825: Notifications and tray indicator are broken

stkrknds:
Since Messenger does not send notifications, the use of this workaround is necessary. This pull request fixes/updates the workaround in order to work with the new design. I've been testing it for a few weeks and it works fine. However,  I would be grateful if someone else could test it as well and report back.  Thanks in advance!

Fixes #1562
Fixes #1825


Alex313031:
@stkrknds I wanna add this to my fork,

source/browser/conversation-list.ts
```diff
@@ -169,38 +169,50 @@ export async function sendConversationList(): Promise<void> {
...
 
+function genStringFromNode(element: Element): string {
+	const cloneElement = element.cloneNode(true) as Element;
+	let emojiString;
+
+	const images = cloneElement.querySelectorAll('img');
+	for (const image of images) {
+		emojiString = image.alt;
+		// Replace facebook's thumbs up with emoji
+		if (emojiString === '(Y)') {
```
dusansimic, line 180: `(Y)` could also be lowercase like `(y)`

You should also consider the svg type of emoji that is usually sent through messenger app (and Caprine). That one should also be shown as a thumbs up.

dusansimic (changes requested):

dusansimic:
Sorry, I haven't yet had time to thoroughly test it but the code seems okay. One problem that exists is that only the chats that are loaded in the sidebar would be monitored for new messages. I'm not sure how it would behave when a new message is received but I'll check it out.

The ultimate solution would be intercepting the websocket packets that are sent to Messenger website and from it so we could both filter out seen signals and catch new messages as they arrive with much less overhead (compared to monitoring the ui). I started some work in https://github.com/sindresorhus/caprine/pull/1688 but it requires completely reverse engineering the websocket protocol they designed so it will definitely take time.

source/browser/conversation-list.ts
stkrknds: Do you think that it would be better to use svg in general(for all the emojis) or only in the specific case of the thumbs up emoji?

dusansimic: I think it's only important for the thumbs up emoji. That one is unique for Messenger and should be displayed in notifications but all others are imho not that important and most stickers can't really be displayed in notifications.

dusansimic:
Once this gets merged, I'll publish a new release.

source/browser/conversation-list.ts
dusansimic, line 180:
```suggestion
		if (emojiString === '(Y)' || emojiString === '(y)') {
```

dusansimic (changes requested):

dusansimic:
If I'm testing this correctly, this only shows a notification once a first new message is received. Once the second one comes in but the first one is not yet read, the notification won't be shown.

Alex313031:
@dusansimic Did you merge this (in modified form) here > https://github.com/Alex313031/caprine-ng/compare/main...sindresorhus:caprine:main

I merged @stkrknds commits into my fork, but now mine is diverged too much. Should I remove his, and add the commits of yours, or try to combine them together?

dusansimic:
> Did you merge this (in modified form) here

No, I've added my own commit for lowercase `(y)` (https://github.com/sindresorhus/caprine/pull/2051/commits/c0b49cd80181661c6fc01e1492f79e15e1ff7dff) and resolved a conflict with the main branch (https://github.com/sindresorhus/caprine/pull/2051/commits/c7eb882c61a023e58cbf6b8710bd198704079732) but I'll need to look at the code better before finally merging into main. Right now we just need to have everything working and we'll clean up the redundant code before merge.

> I merged stkrknds commits into my fork, but now mine is diverged too much. Should I remove his, and add the commits of yours, or try to combine them together?

IMHO, you could just leave it as is for now and merge the final code when we merge it into main. Then you could just revert commit you've made and merge the main branch from upstream. That would be the ideal process for the future, however that would mean the fork will have to wait for changes to land in main on upstream, or at least a green light for merging.

Alex313031:
@stkrknds This patch appears to break on Linux. Whenever someone sends a message, it dings, but then the viewport freezes. The menubar still works, but clicking anywhere in the window does nothing. The only way to resolve it is to relaunch caprine.

dusansimic:
@Alex313031 

> This patch appears to break on Linux. Whenever someone sends a message, it dings, but then the viewport freezes. The menubar still works, but clicking anywhere in the window does nothing. The only way to resolve it is to relaunch caprine.

I think this is related to #2033. That issue already exists without this patch. I still haven't had enough time to properly investigate the cause of that bug. It could be any number of things that are in Caprine but most likely one is some observer enters a loop because some element gets modified repeatedly.

stkrknds:
Sorry for the late response.
> @stkrknds This patch appears to break on Linux. Whenever someone sends a message, it dings, but then the viewport freezes. The menubar still works, but clicking anywhere in the window does nothing. The only way to resolve it is to relaunch caprine.

@Alex313031 I don't face this issue with this patch, are you sure that's caused specifically by it?

> If I'm testing this correctly, this only shows a notification once a first new message is received. Once the second one comes in but the first one is not yet read, the notification won't be shown.

@dusansimic Yes you are right. I'm trying to make it send notifications for every new message, but it is quite difficult because, taking the worst case scenario(1 emoji & new message identical to the previous one), the only mutation that is observed is the time(next to the message). The issue is that, I don't know when such a mutation can be classified as unread, considering that the time changes even if no new message arrives. The only way I can think of is to check if the text is equal to `1 minute`, and interpret  that as a new message, but that's of course troublesome because it depends on the language.

Alex313031:
@stkrknds @dusansimic Reverting the patch makes Caprine work normally again. I made two clones of my repo, and in one, I reverted the commits. I then use npm ci to make sure that the package-lock.json was used, to remove variables. Running the one with the patch leads to caprine freezing.

dusansimic:
@stkrknds 

> Yes you are right. I'm trying to make it send notifications for every new message, but it is quite difficult because, taking the worst case scenario(1 emoji & new message identical to the previous one), the only mutation that is observed is the time(next to the message). The issue is that, I don't know when such a mutation can be classified as unread, considering that the time changes even if no new message arrives.

That is indeed problematic. We could use two things to check if a message is new. If the chat gets promoted up the chat list, it means that it's a new message. And if the text of the message is changed, it's a new message. With those two cases, we could fairly reliably determine if some change in the ui is a new message. There is an extreme case when the new message is the same as the old one and the chat is at the top of the chat list already but I'm guessing we could overlook that for now since that is quite a special case.

Alex313031:
@stkrknds @dusansimic It is definitely something you guys did
I narrowed it down to one file.

https://github.com/sindresorhus/caprine/blob/main/source/browser/conversation-list.ts the conversation-list.ts

This commit broke caprine > https://github.com/sindresorhus/caprine/commit/0d38a719ef29db2b1f1594ed6008d18550595047

The other two files are fine.

I made two clones of the repo. They were both identical except one had the conversation-list.ts file from the previous revision. This one works. The one with the new conversation-list.ts file causes caprine to freeze. IDK how to fix it, but now that you know what it is, maybe yall can figure it out.

stkrknds:
Made some changes. Needs some more testing. The code is a bit messy, but will clean it later.

@Alex313031 I can't reproduce the issue. Don't know if it happens only on specific distros(?). I'm running arch and it's working fine.

Alex313031:
@stkrknds You are running Arch with what DE and WM?

And OK I will try it out.

Alex313031:
@stkrknds @dusansimic This seems to fix it! Cool.

stkrknds:
@Alex313031 I'm using the awesomeWM. It's good that it seems to be working fine. I will test it some more over the weekend and hopefully clean the code a bit.

Alex313031:
@stkrknds Yup, once again, cloning and building breaks it, but clonining, git checking out the notifications branch and building works fine again. I think it was some sort of race condition, based on a quick glance at the changed you made.

Waiting for the code cleanup, and then will rebase my fork on it and make some releases. My fork is mostly identical, except I enable all features (including "dev only" ones), I made a new versions window that shows versions of electron, node_modules, etc., and my fork is still using Electron 22: for Windows 7/8/8.1 support and Ubuntu 16.04 support.

Another minor change that I think I will make a pull request for, is setting the icon explicitly on linux instead of relying only on the packaged version's .desktop file. What I mean is that for Windows and Mac, the icon is set in the code, and will be used even when running `npm run start`. On linux, it is not set at all, and will show a blank icon when running in dev mode. Only when making a .deb or .appimage is the icon set, but again, it is not actually set at the electron level and only via the .desktop file. Directly running /opt/Caprine/caprine leads to no icon. I fixed this by making a 64x64 .png and placing it in the //static dir, and the setting constants for the icon's location(s), with the value depending on platform, and then setting icon: iconPath, in the BrowserWindow webPreferences.

Alex313031:
@stkrknds Also, it DOES appear to be DE/WM specific. Installing Arch and awesomeWM, openBox, or XFCE works fine (I tested in a VM, just to investigate your experience with it. However, KDE and GNOME (which Ubuntu uses a modified version of) breaks it.

Alex313031:
@stkrknds @dusansimic How goes the testing on this? Did you guys read me above comments?

Will it be merged? I've been using it on Windows and Linux with no problems.

stkrknds:
@Alex313031 Sorry, I'm going to clean the code this weekend.

stkrknds:
Okay, I think it's pretty good now, tested it and it works fine.

dusansimic:
> Okay, I think it's pretty good now, tested it and it works fine.

It doesn't seem to work for receiving multiple messages. Also, the contents of the messages is not shown but it might be just on my machine (I run GNOME, but the weird thing is that it worked when I tried it last time).

stkrknds:
> > Okay, I think it's pretty good now, tested it and it works fine.
> 
> It doesn't seem to work for receiving multiple messages. Also, the contents of the messages is not shown but it might be just on my machine (I run GNOME, but the weird thing is that it worked when I tried it last time).

Was working fine yesterday, but today indeed it doesn't work 🙁. Probably some selector needs to be changed. Will take a look in a bit.

Alex313031:
@stkrknds Can confirm, seems broken. I hate this cat and mouse chasing with facebook everytime they decide to update their stuff.

stkrknds:
An update: Right now the only case where notifications are not working is when the same emoji is being sent multiple times continuously. I can't find a way to fix this but I think it is not really that important. I would appreciate it if you could confirm that, with the exception of this particular case,  notifications are working as intended. Thanks.

source/browser/conversation-list.ts
```diff
@@ -169,38 +169,81 @@ export async function sendConversationList(): Promise<void> {
...
-		let curr = (mutation.addedNodes[0] as Element).parentElement;
-		// Find the parent element with the message preview and sender
-		while (curr?.className !== 'rq0escxv l9j0dhe7 du4w35lb j83agx80 pfnyh3mw i1fnvgqd bp9cbjyn owycx6da btwxx1t3') {
-			curr = curr?.parentElement ?? null;
+	const unreadMutations = mutationsList.filter(mutation =>
+		// When a conversations "becomes unread".
+		(
+			mutation.type === 'childList'
+			&& mutation.addedNodes.length > 0
+			&& ((mutation.addedNodes[0] as Element).className === 'x1i10hfl x1qjc9v5 xjbqb8w xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx xdj266r x11i5rnm xat24cr x1mh8g0r x2lwn1j xeuugli xexx8yu x4uap5 x18d9i69 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x1o1ewxj x3x9cwd x1e5q0jg x13rtm0m x1q0g3np x87ps6o x1lku1pv x78zum5 x1a2a7pz')
```
dusansimic, line 198: Could we move all the query selectors to the `selectors.ts` file and add comments about what exactly they are meant for. It would be far easier to maintain this feature and fix them in case they break.

Other than this, pr looks good.

dusansimic (changes requested):

stkrknds:
@dusansimic Made some changes. Created variables for the selectors and fixed another bug that I found. 

The freeze issue(#2053) might be caused by the currently broken countUnread function. By fixing this workaround we may indirectly solve the issue. I can't say for sure, since I can't reproduce it.

dusansimic:
@stkrknds I just wanted to push this out quickly so I made a few changes myself. I've moved the selectors to the `selectors.ts` file. Also, I've changed the `any` type of the `alreadyChecked` array to `string[]` which caused some type errors and I've fixed them by skipping an iteration in case `href` is null or undefined (https://github.com/sindresorhus/caprine/pull/2051/files#diff-07559914018c1e28768b2f3ce54fd49cc8071201423637fad35b372c318ebed7R224). Could you double check that this doesn't affect negatively this feature.

If everything looks good, we're merging this and I'll push out a release today.

stkrknds:
Works fine.

dusansimic:
Thanks for the help!

mzso:
I had no notifications so far. Does it look for english text or something? Facebook appears localized in Caprine for me.

stkrknds:
@mzso No, the language shouldn't matter. Right now there are some selectors that need to be updated, so notifications may not work. Did you use the version 2.59 a week ago? I think notifications worked then.

dusansimic:
@stkrknds 

> I think notifications worked then.

They did, the selectors seem to be broken.

mzso:
It seems to be working now. At least I saw the tray icon turn purple a couple times.

PR #2063, opened 2023-09-14
Use more specific  CSS selector for Search results Close button

mquevill:
The previous selector (`.x14hiurz`) was not specific enough for the search results "Close" button, since it also affected preferences selectors below.

image
image

The color is updated so the "Close" actually looks like a clickable button and no longer affects the preferences checkboxes.

image
image

(I think I'd like *more* contrast on the checkboxes from a design perspective, but am reverting to default styles for now.)


dusansimic:
> I think I'd like more contrast on the checkboxes from a design perspective, but am reverting to default styles for now.

I completely agree, but that could be in an another pr. Thanks for this!

Issue #1892, opened 2022-08-24
Better error handling and reporting


dusansimic:
So far, vast majority of patches (just subjectively) are related to either css fixes or fixing css selectors in Caprine logic (shortcuts or other systems). That's the thing that changes most ofthen. The one we can't really influcence much are css fixes because when they happen, the only thing we can do is create a new release with a fix. The one we __can__ influcence are css selectors.

Once they change and break, most of the time we just patch it up and send out a new release. The issue reports are usually "<feature> is not working" and it's up to us to track down what the root cause of it is. It doesn't usually take much time but in case of #1881 it took quite a bit of time to track down what the actuall casue of the issue was (it was a css selector for something else altogether which blocked the execution of code).

Right now elements are queried with `document.querySelector` or `elementReady` functions and they could return a `null` or `undefined` respectively. Throughout the whole codebase, this is handled by using [non-null assertation operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) or [optional chaining operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) which either just "makes the code work" but could throw errors at some point (which happens) or doesn't throw anything altogether and just nothing happens which is painful to debug. It's not that use of these operators is bad, to me it just feels like it's implemented with an expectation that query selectors wouldn't change.

My recommendation for making maintenance and fixing issues easier is to log in some way what queries fail exactly so once an issue occurs, user could just paste the log into the issue report and maintainer could very quickly track down the issue and fix it. Caprine is already having trouble handling all the changes in Messenger website so this would help a lot imho.

The logging I had in mind is writing failed queries with timestamps and any other useful information to a file which would be cleared at every start (so it doesn't clog up the disk if users restart Caprine in sane intervals). On report an issue button in help menu, Caprine could upload or paste directly the log to the issue that's being opened so users wouldn't need to dig around Caprines local files. It seems like the easiest solution and it should add too much to users workflow. The whole refactoring would of course thake some effort but I think it would pay off.

lefterisgar:
I agree with all your points. At its current state, Caprine is a total nightmare to debug, let alone fix the issues themselves.

My only concerns are:
* Could this potentially affect Caprine's performance (which according to numerous reports is terrible already)
* How will users receive this? Is it going to be optional?

Also we will have to refactor the code and document everything so people can see what is being collected and how.

dusansimic:
[...]

PR #2064, opened 2023-09-14
Verify that notificationCheckbox is non-null

mquevill:
Since the notifications are currently somewhat broken, this is a temporary patch that returns `false` if the selector is null. I would be wary of simply toggling the "Do not disturb" mode, since it normally sets a time limit for that mode, which may require more of an overhaul of notification options. I'm open to handling this differently, but it does handle this promise/error.

The error on the current version:
```
(node:10656) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of null (reading 'checked')
    at ...\caprine\dist-js\browser.js:208:34
    at async EventEmitter.listener (...\caprine\node_modules\electron-better-ipc\source\renderer.js:46:34)
```

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -272,6 +272,11 @@ ipc.answerMain('toggle-mute-notifications', async () => {
 		await closePreferences();
 	}
 
+	// TODO: Fix notifications
+	if (notificationCheckbox === null) {
+		return false;
+	}
+
 	return !notificationCheckbox.checked;
 });
 


dusansimic:
A problem is that those options don't show up when preferences modal is programmatically open. When it's opened manually by a user, both "Do not disturb" and "Notification sounds" as well as "Active status" show up. We still haven't figured out why it happens but it's the root cause of this issue.

This workaround merely avoids throwing an error when the options don't show up which is both good and bad. I'd rather have it log at least some message so we could know that it's happening. It would be something like what was suggested in https://github.com/sindresorhus/caprine/issues/1892. I haven't still had time to finish up that patch but I'll try to work on it in following weeks and push at least something for start.

stkrknds:
Hello. I've been working on this. I think this should be merged since there is no notification checkbox right now in messenger's preferences. 

> A problem is that those options don't show up when preferences modal is programmatically open. When it's opened manually by a user, both "Do not disturb" and "Notification sounds" as well as "Active status" show up. We still haven't figured out why it happens but it's the root cause of this issue.

Yes this is true, but even when these options show up, there is simply not a notification checkbox(the sounds checkbox I think is supposedly handled by the function toggleSounds, so it's unrelated). By the way, I found a fix for this. If you wait for the menu to get displayed, then the preferences always include the missing options. You can achieve this by changing line 68 in browser.ts to:
```await elementReady(`${selectors.conversationMenuSelectorNewDesign} > div > div > div> div > div:nth-child(3)[style*="display: block;"]`, {stopOnDomReady: false});```
Of course by changing this, the menu becomes visible for a split second. As far as I can tell, the function responsible for hiding it is the withMenu, that adds the `hide-dropdowns` class specifically for this purpose, which needs to be fixed as well.

PR #2065, opened 2023-09-14
Select views for new sidebar

### LINKED ISSUE #1984: Most view options aren't working
Running Caprine 2.57.0 on Windows 10 - this is a fresh install from today, as I just realized that Caprine has Windows builds too.

Most of the Caprine-specific options in the View menu are having no effects for me. While the zoom and theme controls work fine, the rest of the controls have issues.

"Hide Names and Avatars," the entire Sidebar menu, and "Show Message Buttons" can be selected and toggled but don't seem to do anything regardless of state. 

"Show Active Contacts," "Show Message Requests," and "Show Hidden Threads" don't do what their names state, but instead seem to put the Communities section in the sidebar into edit mode (as if you had clicked the "manage" button).

I have already tried relogging, relaunching, and reinstalling the app through both the installer and Chocolatey and the issue persists.

mquevill:
I'm not sure if it's safe to assume that (most) people have the new sidebar at this point, or if we should retain past compatibility. This should allow for selecting view through the View menu, if users want to select it that way.

I renamed the actions to reflect what Messenger calls the different views as well.

I am very open to any suggestions about how this should be handled or adjusted as needed.

Fixes #1984.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -87,9 +87,25 @@ async function selectOtherListViews(itemNumber: number): Promise<void> {
 	// In case one of other views is shown
 	clickBackButton();
 
-	await withSettingsMenu(() => {
-		selectMenuItem(itemNumber);
-	});
+	const newSidebar = await isNewSidebar();
+
+	if (newSidebar) {
+		const items = document.querySelectorAll<HTMLElement>(
+			`${selectors.viewsMenu} > span > a`
+		);
+
+		const selector = itemNumber <= items.length ? items[itemNumber - 1] : null;
+
+		console.log(selector);
+
+		if (selector) {
+			selector.click();
+		}
+	} else {
+		await withSettingsMenu(() => {
+			selectMenuItem(itemNumber);
+		});
+	}
 }
 
 function clickBackButton(): void {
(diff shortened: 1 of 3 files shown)


dusansimic:
> I'm not sure if it's safe to assume that (most) people have the new sidebar at this point

Really? Why? It's been quite a long time now and we haven't had any reports about it. It's mostly that selectors got broken because of ui changes.

mquevill:
> > I'm not sure if it's safe to assume that (most) people have the new sidebar at this point
> 
> Really? Why? It's been quite a long time now and we haven't had any reports about it. It's mostly that selectors got broken because of ui changes.

It seems like there have been several generations of what's considered "new". I'm good with removing the checks in `selectOtherListViews()` if you think that's fine.

dusansimic:
> It seems like there have been several generations of what's considered "new".

That's correct. I'd just like to deprecate checks like these at some point since it becomes quite rare. This could be safely checked with anonymous telemetry which should be opt in in order to respect users privacy but should be recommended in order to help us maintain Caprine. That's however a topic for another discussion.

I'll merge this now. Thanks for the good work!

PR #2108, opened 2024-01-12
Hidden dialog issue Fix

stkrknds:
Hello. Changed/updated some selectors in order to fix the hidden dialog issue.

Fixes #2106.
Fixes #2100(?) (not sure about the camera part)
Fixes #2098.
Fixes #2096.
Edit: Fixes #2113.

PR #2109, opened 2024-01-12
Tray icon Fix

stkrknds:
The tray icon does not change when there are unread messages. This is due to the broken chatsIcon selector.

### DIFF
--- a/source/browser/selectors.ts
+++ b/source/browser/selectors.ts
@@ -1,6 +1,6 @@
 export default {
 	leftSidebar: '[role="navigation"] > div > div', // ! Tray icon dependency
-	chatsIcon: '[data-pagelet="MWCMInboxLeftRailSidebar"] > div > div > div > span:nth-child(1) > a', // ! Tray icon dependency
+	chatsIcon: '[data-pagelet="MWCMInboxLeftRailSidebar"] > div > div > div:nth-child(1) a', // ! Tray icon dependency
 	conversationList: '[role=navigation] [role=grid] [class="x1n2onr6"]',
 	conversationSelector: '[role=main] [role=grid]',
 	conversationSidebarUnreadDot: 'x1i10hfl x1qjc9v5 xjbqb8w xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx xdj266r x11i5rnm xat24cr x1mh8g0r x2lwn1j xeuugli xexx8yu x4uap5 x18d9i69 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x1o1ewxj x3x9cwd x1e5q0jg x13rtm0m x1q0g3np x87ps6o x1lku1pv x78zum5 x1a2a7pz',
--- a/source/browser/selectors.ts
+++ b/source/browser/selectors.ts
@@ -1,6 +1,6 @@
 export default {
 	leftSidebar: '[role="navigation"] > div > div', // ! Tray icon dependency
-	chatsIcon: '[data-pagelet="MWCMInboxLeftRailSidebar"] > div > div > div:nth-child(1) a', // ! Tray icon dependency
+	chatsIcon: '[class="x1i10hfl xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx x2lwn1j xeuugli x4uap5 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x87ps6o x1lku1pv x1a2a7pz x6s0dn4 x1q0g3np xn3w4p2 x1nn3v0j x1120s5i x1ry3o7a x1lq5wgf xgqcy7u x30kzoy x9jhf4c xdj266r x11i5rnm xat24cr x1mh8g0r x78zum5"]', // ! Tray icon dependency
 	conversationList: '[role=navigation] [role=grid] [class="x1n2onr6"]',
 	conversationSelector: '[role=main] [role=grid]',
 	conversationSidebarUnreadDot: 'x1i10hfl x1qjc9v5 xjbqb8w xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx xdj266r x11i5rnm xat24cr x1mh8g0r x2lwn1j xeuugli xexx8yu x4uap5 x18d9i69 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x1o1ewxj x3x9cwd x1e5q0jg x13rtm0m x1q0g3np x87ps6o x1lku1pv x78zum5 x1a2a7pz',
--- a/source/browser/selectors.ts
+++ b/source/browser/selectors.ts
@@ -1,6 +1,6 @@
 export default {
 	leftSidebar: '[role="navigation"] > div > div', // ! Tray icon dependency
-	chatsIcon: '[class="x1i10hfl xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx x2lwn1j xeuugli x4uap5 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x87ps6o x1lku1pv x1a2a7pz x6s0dn4 x1q0g3np xn3w4p2 x1nn3v0j x1120s5i x1av1boa x1lq5wgf xgqcy7u x30kzoy x9jhf4c xdj266r x11i5rnm xat24cr x1mh8g0r x78zum5"]', // ! Tray icon dependency
+	chatsIcon: '[data-pagelet="MWCMInboxLeftRailSidebar"] > div > div > div > span:nth-child(1) > a', // ! Tray icon dependency
 	conversationList: '[role=navigation] [role=grid] [class="x1n2onr6"]',
 	conversationSelector: '[role=main] [role=grid]',
 	conversationSidebarUnreadDot: 'x1i10hfl x1qjc9v5 xjbqb8w xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx xdj266r x11i5rnm xat24cr x1mh8g0r x2lwn1j xeuugli xexx8yu x4uap5 x18d9i69 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x1o1ewxj x3x9cwd x1e5q0jg x13rtm0m x1q0g3np x87ps6o x1lku1pv x78zum5 x1a2a7pz',
--- a/source/browser/selectors.ts
+++ b/source/browser/selectors.ts
@@ -1,6 +1,6 @@
 export default {
 	leftSidebar: '[role="navigation"] > div > div', // ! Tray icon dependency
-	chatsIcon: '[class="x1i10hfl xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx x2lwn1j xeuugli x4uap5 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x87ps6o x1lku1pv x1a2a7pz x6s0dn4 x1q0g3np xn3w4p2 x1nn3v0j x1120s5i x1ry3o7a x1lq5wgf xgqcy7u x30kzoy x9jhf4c xdj266r x11i5rnm xat24cr x1mh8g0r x78zum5"]', // ! Tray icon dependency
+	chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a', // ! Tray icon dependency
 	conversationList: '[role=navigation] [role=grid] [class="x1n2onr6"]',
 	conversationSelector: '[role=main] [role=grid]',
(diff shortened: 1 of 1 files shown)


devCKVargas:
This didn't fix the issue on Win 11 Pro 22631.3085. How could I show the logs?

stkrknds:
Hi @devCKVargas. Just to be sure, are you using caprine from https://github.com/stkrknds/caprine/tree/tray-icon ?
I've changed the selector again, so please try the new version and let me know if it fixes the issue. Thanks!

sgtcoder:
For me, this didn't fix the issue with CPU usage and I still had to put `"electron": "^28.2.1",` for my devDependencies or it just showed a white screen.

The status tray icon still didn't change when I have an unread notification and it still showed elevated usage.

```
git clone https://github.com/sindresorhus/caprine

cd caprine

## Pull PR ##
git pull https://github.com/sindresorhus/caprine refs/pull/2109/head

nvm use
npm install

## Start ##
npm start

nano package.json
"electron": "^28.2.1",
```

devCKVargas:
> Hi @devCKVargas. Just to be sure, are you using caprine from https://github.com/stkrknds/caprine/tree/tray-icon ? I've changed the selector again, so please try the new version and let me know if it fixes the issue. Thanks!

Unfortunately, still broken.
Pulled the latest commit did npm install && npm start.
Issue still exists. Tray icon didn't light up when I received new messages, same with notification.

stkrknds:
Thanks for your comments. To be honest, I have no idea why it is not working for you.. Changed the selector again, could you please check if it works now. Again, thanks a lot for your time.

sgtcoder:
Are you talking about the Chat icon in the Messenger App sidebar? OR the Chat Icon in the Systray/XFCE panel?

stkrknds:
> Are you talking about the Chat icon in the Messenger App sidebar? OR the Chat Icon in the Systray/XFCE panel?

Caprine uses the Chats icon in Messenger's sidebar to update the tray icon.

sgtcoder:
The tray icon doesn't work, but the Messenger icon does.

stkrknds:
@sgtcoder Could you provide the class name of the Chats icon? You can do that by first, opening the console(pressing F12). Then Ctrl+Shift+C and click on the "Chats" icon. The element we want is the one with the aria-label attribute.

sgtcoder:
```
aria-label="Chats, 1 unread"

class="x1i10hfl xjbqb8w xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx x2lwn1j xeuugli x4uap5 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x87ps6o x1lku1pv x1a2a7pz x6s0dn4 x1q0g3np xn3w4p2 x1nn3v0j x1120s5i x1lq5wgf xgqcy7u x30kzoy x9jhf4c xdj266r x11i5rnm xat24cr x1mh8g0r x78zum5"
```

sgtcoder:
looks like there is a slight difference in the one in your code change.

So you guys have to do this like everytime Facebook pushes a new update out?

sgtcoder:
Not saying it's a good idea, but it's not like Facebook labels ids on these, can you just do first span > a?

stkrknds:
> looks like there is a slight difference in the one in your code change.
> 
> So you guys have to do this like everytime Facebook pushes a new update out?

Yes, unfortunately.

stkrknds:
> Not saying it's a good idea, but it's not like Facebook labels ids on these, can you just do first span > a?

Tried to do something similar with the previous commits as it is harder to break, but that still didn't work for you, while it worked for me.

sgtcoder:
MWCMInboxLeftRailSidebar class doesn't exist for me. Let me see if I can find something that will work, now that I know what this is for.

That's crazy you guys have to do this all the time.

sgtcoder:
Something like this

```
chatsIcon: '[aria-label="Inbox switcher"][aria-label^="Chats"]', // ! Tray icon dependency
```

but I am trying to figure out why that doesn't work.

stkrknds:
> Something like this
> 
> ```
> chatsIcon: '[aria-label="Inbox switcher"][aria-label^="Chats"]', // ! Tray icon dependency
> ```
> 
> but I am trying to figure out why that doesn't work.

It's better to avoid using aria-label in a selector, because it will not work for every language.

sgtcoder:
Good point on that, didn't realize. OMG, I got it to work with updating the class. And no crazy CPU usage.

Now let me see if I can get it more dynamic.

sgtcoder:
```
chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] > span:nth-child(1) > a', // ! Tray icon dependency
```
This works too, it's not too terrible.

sgtcoder:
Notification alerts are coming in now too.

I noticed that the Messenger settings, etc aren't working either.

stkrknds:
> ```
> chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] > span:nth-child(1) > a', // ! Tray icon dependency
> ```
> 
> This works too, it's not too terrible.

It doesn't work when the sidebar is expanded. I would suggest to modify this selector ```[data-pagelet="MWCMInboxLeftRailSidebar"] > div > div > div:nth-child(1) a``` by replacing ```[data-pagelet="MWCMInboxLeftRailSidebar"]``` with something that works for you. For example, something like this: ```[role="navigation"] > div > div > div > div > div > div:nth-child(1) a```

stkrknds:
> Notification alerts are coming in now too.
> 
> I noticed that the Messenger settings, etc aren't working either.

This is related to #2108

sgtcoder:
Gotcha. Still working on it, but role navigation isn't quite unique

stkrknds:
Okay.. I modified the selector that you found, in order to work with the expanded sidebar as well. Can you test if it works on your end?

sgtcoder:
Yes, I saw your commit and that worked. Should we do nth selector on that a tag? Either one works though.

```
chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a:nth-child(1)', // ! Tray icon dependency
```

stkrknds:
> Yes, I saw your commit and that worked. Should we do nth selector on that a tag? Either one works though.
> 
> ```
> chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a:nth-child(1)', // ! Tray icon dependency
> ```
Doesn't really matter. It picks the first one that it finds regardless. I think we can leave like this for now. Thanks for your help.

sgtcoder:
I think also it wouldn't be bad to update some of the packages in package.json. I have a list of ones that can be updated without noticing any breaking. Or at least element-ready since if notifications break again at least the CPU won't cook.

sgtcoder:
Ahh, I see, wasn't sure how that was connected in the Electron app. Thanks for clarifying and you are welcome.

Let me know if you need anything else. I am a web developer/programmer as well. Caprine is great on my build so I don't have to pull out the phone.

devCKVargas:
Latest commit seemed to have fixed it including the notification. Thanks @stkrknds!

sgtcoder:
https://github.com/sgtcoder/caprine/releases/tag/v2.59.3

Here is a working version with all the package dependencies updated that could be updated, with the Tray icon fix and the hidden dialog fix as well. Seems to be working well for me, I will continue to test/use.

hlqviet:
> https://github.com/sgtcoder/caprine/releases/tag/v2.59.3
> 
> Here is a working version with all the package dependencies updated that could be updated, with the Tray icon fix and the hidden dialog fix as well. Seems to be working well for me, I will continue to test/use.

Awesome. Thank you. It has been 4 months since the last PR was merged.

sgtcoder:
@hlqviet But keep in mind this is under my forked copy of it. Probably good for a temp solution until main Caprine gets updated.

hlqviet:
> @hlqviet But keep in mind this is under my forked copy of it. Probably good for a temp solution until main Caprine gets updated.

There still seems to be some CPU usage while the app is idle but it's much better than before.

sgtcoder:
I do agree that there is some usage (quite low, 33 percent on 1 thread). Not as bad as Mailspring, and I wonder if the usage has always been there.

Either way, if the Caprine team can point me in the direction of looking at the resource usage of caprine components, I might be able to take a peak. My guess is maybe monitoring of certain things like notifications. Like because you would have to monitor that in the app. Like every second or 5 seconds or something?

sgtcoder:
So the other thing is the Marketplace notifications aren't working. Maybe we can pass a 2nd paramater to track this?

source/browser/selectors.ts
```diff
@@ -1,6 +1,6 @@
 export default {
 	leftSidebar: '[role="navigation"] > div > div', // ! Tray icon dependency
-	chatsIcon: '[class="x1i10hfl xjqpnuy xa49m3k xqeqjp1 x2hbi6w x13fuv20 xu3j5b3 x1q0q8m5 x26u7qi x972fbf xcfux6l x1qhh985 xm0m39n x9f619 x1ypdohk xdl72j9 x2lah0s xe8uvvx x2lwn1j xeuugli x4uap5 xkhd6sd x1n2onr6 x16tdsg8 x1hl2dhg xggy1nq x1ja2u2z x1t137rt x87ps6o x1lku1pv x1a2a7pz x6s0dn4 x1q0g3np xn3w4p2 x1nn3v0j x1120s5i x1av1boa x1lq5wgf xgqcy7u x30kzoy x9jhf4c xdj266r x11i5rnm xat24cr x1mh8g0r x78zum5"]', // ! Tray icon dependency
+	chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a', // ! Tray icon dependency
```
mquevill, line 3: This selector will actually work on all 5 of the categories (Chats, People, Marketplace, Requests, Archive), so I would suggest a more specific selector:
```diff
-	chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a', // ! Tray icon dependency
+ 	chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] > span:nth-of-type(1) > a', // ! Tray icon dependency
```

While it will work in its current state (since it will return the first hit on that selector), I'd rather it be more specific just in case back-end changes affect these categories in an unexpected way.

mquevill: I know this was previously discussed in this thread, but I'd rather it break completely instead of grab another (unexpected) element.

(Also, the previous change of `a:nth-child(1)` still picks up all 5 categories.)

mquevill:
> So the other thing is the Marketplace notifications aren't working. Maybe we can pass a 2nd paramater to track this?

The selector `[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a` will return all 5 categories and in theory we could loop over each one and add them together. However, since the `elementReady()` only applies to one selector, it could be fragile to assume the other elements would be fully ready.

A useful function would be `observeReadyElements()` https://github.com/sindresorhus/element-ready/pull/36, but it's only available in v6.1.0+ and v6.0.0 is when that package changed to an [ESM package](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

source/browser/selectors.ts
stkrknds: Hello @mquevill . The problem with this selector `[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] > span:nth-of-type(1) > a` is that it doesn't work when the sidebar is expanded. I agree that it would be better to find a selector that is specifically for the chats icon but I haven't found one that works regardless of the sidebar's state.

sgtcoder:
It sounds like we want to consider detecting if the sidebar is open or closed and running the specific classes based off of that. That should make it a little more effective.

sgtcoder:
```
chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a:nth-of-type(1)',
```

Hey guys, have we tried this, it's working on closed and open for me.

sgtcoder:
Also, you guys mentioned you only want the chat icon notifications, but I would like the Marketplace as well.

source/browser/selectors.ts
mquevill: Ah, I didn't realize the structure changed that much between the two sidebar configurations.

mquevill:
> ```
> chatsIcon: '[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"] a:nth-of-type(1)',
> ```
> 
> Hey guys, have we tried this, it's working on closed and open for me.

This change works for the expanded sidebar, but not for the collapsed one, since each `a` element is within a `span` element, so that selector can still select the first `a` element in each `span`. So yes, this selector will work, but with the implicit selection of the first match. Below is 

```
#collapsed
div
    span
        a
    span
        a
    span
        a
    span
        a
    span
        a

#expanded
div
    a
    a
    a
    a
    a
```

sgtcoder:
Mine is working while collapsed and opened. What are you doing to test while closed and it not working?

sgtcoder:
Plus this just grabs the first a tag whether it's in a span or not `a:nth-of-type(1)`

mquevill:
@sgtcoder It's not that the selector doesn't work, but it technically matches all 5 of the categories listed. The selector `a:nth-of-type(1)` grabs any matching `a` that is the first within its parent. Ideally, I'd like it to be an exact match.

image

sgtcoder:
What about this? It doesn't let me paste into Messenger Console because of XSS.

```
'[class="x9f619 x1n2onr6 x1ja2u2z x78zum5 xdt5ytf x2lah0s x193iq5w xdj266r"]:nth-of-type(1) a:nth-of-type(1)'
```

sgtcoder:
Can you just have it return after the first element selection? It's going to be difficult getting just the first element only and support closed and open states.

stkrknds:
> Can you just have it return after the first element selection? It's going to be difficult getting just the first element only and support closed and open states.

From https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector: 

> Note: The matching is done using depth-first pre-order traversal of the document's nodes starting with the first element in the document's markup and iterating through sequential nodes by order of the number of child nodes. 

This means that querySelector, as it is now implemented, will always return the chats icon, which is the first matching element(depth-wise).

sgtcoder:
Yes, which is exactly what you wanted?

Otherwise, we might need to add a MarketPlace Icon entry and listen in on that (or add an option), which is why I just have it listening to the entire sidebar for right now on my fork. My phone is on silent 24/7, so these are important notifications too.

stkrknds:
If you want to do something similar for marketplace messages, I suggest you take a look at how the tray icon works, and in particular the following snippets:

https://github.com/sindresorhus/caprine/blob/91689db078f11e04d786536fe357169e118ecdbc/source/browser/conversation-list.ts#L274

https://github.com/sindresorhus/caprine/blob/91689db078f11e04d786536fe357169e118ecdbc/source/browser/conversation-list.ts#L296-L305

https://github.com/sindresorhus/caprine/blob/91689db078f11e04d786536fe357169e118ecdbc/source/browser/conversation-list.ts#L262-L270

https://github.com/sindresorhus/caprine/blob/91689db078f11e04d786536fe357169e118ecdbc/source/index.ts#L418-L421

https://github.com/sindresorhus/caprine/blob/91689db078f11e04d786536fe357169e118ecdbc/source/index.ts#L102-L139

I've never used the marketplace, so I can't help any further. I think this feature can be covered in another PR.

source/browser/selectors.ts
mquevill: I'm fine with this selector change to handle both the expanded and collapsed sidebar cases, since I don't think there's a better way just with CSS selectors

dusansimic:
Hi everyone. Sorry for the radio silence, it's not easy at the moment to be present all the time so you'll have to excuse me for that.

Regarding the PR. I see that @mquevill has joined in to review it. Seems like most of the issues have been resolved. Is this ready for merging?

mquevill:
> Hi everyone. Sorry for the radio silence, it's not easy at the moment to be present all the time so you'll have to excuse me for that.
> 
> Regarding the PR. I see that @mquevill has joined in to review it. Seems like most of the issues have been resolved. Is this ready for merging?

This selector works for now for both expanded and collapsed left sidebar. There's some additional work on other notification numbers, but that will be in another PR for later.

dusansimic:
Thank you everyone for contributing to fixing this issue! Especially @stkrknds for doing the initial work!

sgtcoder:
> If you want to do something similar for marketplace messages, I suggest you take a look at how the tray icon works, and in particular the following snippets:

@stkrknds 

Thank you for sharing this. I will look into this and see what I can put together.

Issue #2107, opened 2024-01-09
High CPU usage on Ubuntu 22.04


haifisch88:
Hello, I noticed Caprine using around 15-25% when idle on Ubuntu 22.04. Please let me know if you need more details. Thanks!

hujberhunor:
Same here, but 91%
5.15.144-1-MANJARO
caprine 2.59.1
Top outpot `1940406 i3hunor   20   0 1133.2g 287120 103908 R 107.0   3.6 113:38.55 electron`

maetthew:
Same on Debian. Driving me nuts.

mstapaj:
Same on Fedora 39. Higher CPU usage by about 5-10%

Terrorzyw:
Kubuntu 23.10 flatpak same problem.

tgm-git:
I am seeing the same in Debian 12. Initially the app has an expected usage and then after several hours of uptime it starts using more CPU, usually taking up 90~% of one core.

Restarting the application seems to help the problem.

dusatvoj:
Same @ arch linux. Eating one core.

wkrska:
Same on Ubuntu 22.04.3 LTS, when application is in foreground it has one core pinned

sgtcoder:
I am having this issue too again. They fixed it in previous versions, but looks like the issue is back. Debian 12.

maetthew:
There is a pull request (#2110) ready which seems to fix the issue.

sgtcoder:
When running the below, the app opens and displays Messenger and then the screen turns white. I noticed the CPU usage being low, but can't confirm usage.

Also, how do I build for a .deb package? I couldn't see that in the docs.

```
git clone https://github.com/bnabholz/caprine
cd caprine
git checkout fixes/high-cpu-usage

npm install && npm start
```

phodamentals:
Same issue with CPU hitting  at least 70% at all times while app is open.
OS: Pop!_OS 22.04 LTS
Caprine: 2.59.1

sgtcoder:
So here is what I did for a workaround solution for myself (so I can have it auto-pull into my repository server as well).

I forked this repo `sindresorhus/caprine` then pulled in the PR Fix: `https://github.com/sindresorhus/caprine/pull/2110`

I updated `"electron": "^28.2.1"` since that also was needed for me.

I also updated all the packages that could be updated without breaking.

Resulting in this: `https://github.com/sgtcoder/caprine/releases`

[...]

PR #2110, opened 2024-01-12
Upgrade element-ready to 7.0.0

bnabholz:
This fixes high CPU usage (for me, at least) as mentioned in https://github.com/sindresorhus/caprine/issues/2107.  I was seeing the same thing, close to 100% of a CPU being used on my 4 core / 8 thread system, whenever the Caprine window is open.  The offending process is the renderer process for electron, specifically the one with the `--no-sandbox` argument.

I was able to use profiling in the devtools and determined `element-ready` to be the cause.  Upgrading `element-ready` to 6.x or 7.x (and fixing the import in `source/browser.ts`) resolves it for me.  CPU will still spike when interacting with the application, but at idle is back to 1-2%.

Reverting `element-ready` back to ^5.0.0 brings the issue back.  I did not dig into the changes in `element-ready` to understand why this happens, just my observation that this does resolve the CPU issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
 				"electron-store": "^8.1.0",
 				"electron-updater": "^5.2.1",
 				"electron-util": "^0.17.2",
-				"element-ready": "^5.0.0",
+				"element-ready": "^7.0.0",
 				"facebook-locales": "^1.0.916",
 				"is-online": "^9.0.1",
 				"json-schema-typed": "^8.0.1",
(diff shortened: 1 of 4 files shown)


sgtcoder:
When running this, the app opens and displays Messenger and then the screen turns white (this isn't an issue with "electron": "^28.2.1".

maetthew:
Seems to fix the high CPU usage but dark theme doesn't seem to work at least on Debian.

sgtcoder:
This does not actually fix anything.

The reason why you don't get a high load is because it breaks notifications all together. When doing that the high CPU load drops.

When using the updated patch from Tray Icon PR, you can see notifications are broke. Reverting back to elementor-ready 5 with the patch, drops CPU usage and the notifications work again. But elementor 7 and the notification fix does not fix notifications.

bnabholz:
@sgtcoder You are right, this was a simple fix for a more difficult problem I did not fully understand.  I've tested and experienced that with these commits cherry-picked on top of your 2.59.3 tag, notifications do not work.  I believe you are on to the more correct solution.  I think we can close this PR.

PR #2124, opened 2024-02-09
Update Messenger for Mac/Windows selectors

mquevill:
These selectors work on both Mac and Windows for both the chat list and settings list.


mquevill:
In fact, Linux clients are even showing a "Try Messenger for Desktop" banner (despite no official client). Fortunately these changes also apply to Linux builds.

Issue #2121, opened 2024-02-03
Updating package.json and requesting to help


sgtcoder:
I was wondering how I can help with updating the package.json file to newer packages. The below are the latest package versions I was able to install while keeping the app working. Some of them could not be updated.

What would I need to do to get this tested fully and so I can create a PR if that's fine with the team?

I have been using it from my fork and has been working for a few days now: https://github.com/sgtcoder/caprine

```
"dependencies": {
	"@electron/remote": "^2.1.2",
	"@sindresorhus/do-not-disturb": "^1.1.0",
	"electron-better-ipc": "^2.0.1",
	"electron-context-menu": "^3.6.1",
	"electron-debug": "^3.2.0",
	"electron-dl": "^3.5.1",
	"electron-localshortcut": "^3.2.1",
	"electron-store": "^8.1.0",
	"electron-updater": "^6.1.7",
	"electron-util": "^0.17.2",
	"element-ready": "^5.0.0",
	"facebook-locales": "^1.0.916",
	"is-online": "^9.0.1",
	"json-schema-typed": "^8.0.1",
	"lodash": "^4.17.21",
	"npm-check-updates": "^16.14.14",
	"p-wait-for": "^3.2.0"
},
"devDependencies": {
	"@sindresorhus/tsconfig": "^0.7.0",
	"@types/electron-localshortcut": "^3.1.3",
	"@types/facebook-locales": "^1.0.2",
	"@types/lodash": "^4.14.202",
	"del-cli": "^5.1.0",
	"electron": "^28.2.1",
	"electron-builder": "^24.9.1",
	"husky": "^9.0.10",
	"np": "^9.2.0",
	"patch-package": "^8.0.0",
	"stylelint": "^16.2.1",
	"stylelint-config-xo": "^0.22.0",
	"typescript": "^5.3.3",
	"xo": "^0.56.0"
},
```

sgtcoder:
Actually, the newer element-ready breaks the notifications

dusansimic:
Hi @sgtcoder, thanks for reaching out and offering to help with the project!

The current main issue with the project is that there is a lot of technical debt, probably even before the project was rewritten from JS to TS. Currently everything is done in a relatively old fashioned way compared to how Electron has changed over time. There is also a lot of dead code that could be trimmed.

[...]

PR #2136, opened 2024-02-25
version bumps and lint updates

sgtcoder:
These are all the packages that could be updated before ES6 is required. I also couldn't update some lint packages since it becomes more strict.

This is tested and working on Debian 12.


dusansimic:
Looks fantastic! Thank you for this great contribution and all the effort you've put into it.

A great next step besides refactoring discussed in #2121 would be trimming down all those linter rules being turned of by actually fixing the lint issues. It's not something pressing at the moment but it would be a good thing to work on.

sgtcoder:
@dusansimic You are welcome and thank you for accepting my PR.

I agree with the linting to be updated to not exclude so many rules. I excluded the harmless ones and ones that would actually fail on the prior releases as well. There would be a few lint rules we would keep disabled because they are more of preferences vs actual linting errors. Like using full words for variable names vs short names. `Ex: var version vs var ver`

Monday is fine, we should test the Main branch before releasing so it doesn't cause potential issues for others.

`electron-vite` sounds interesting, but how do we want to handle the conversion of that? Like create a branch update it and merge it back? We can discuss on Discord.

Thanks again for your contribution on Caprine to get us back to a working state.

PR #2181, opened 2024-05-28
Use aria-label for Preferences selector

### LINKED ISSUE #2167: "Messages are missing. Sync now." does nothing
For end-to-end encrypted chats, which require the new Facebook Messenger pin number to be entered to view the chats, the "Sync now" feature does not work--nothing happens when I click it. As a result, all encrypted chats are missing up to the point at which Caprine was installed.

Caprine 2.60.1
Pop!_OS 22.04 LTS
Linux 6.8.0-76060800daily20240311-generic

mquevill:
The selector that was previously used is also used for Security Alerts (including entering in your message sync PIN) and other windows, so it would be difficult to use that style of selector. Using `aria-label=Preferences` seems to be the best way to address this.

The current method used `elementReady()` on the first element that matched, which was usually the Sync PIN window, which caused some other conflicts with how the typical code flow would work.

The extra "style" code is no longer necessary with the backend design changes. (It's probably been outdated for some time...)

Fixes #2167.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -243,11 +243,6 @@ async function openHiddenPreferences(): Promise<boolean> {
 	if (!isPreferencesOpen()) {
 		document.documentElement.classList.add('hide-preferences-window');
 
-		const style = document.createElement('style');
-		// Hide both the backdrop and the preferences dialog
-		style.textContent = `${selectors.preferencesSelector} ._3ixn, ${selectors.preferencesSelector} ._59s7 { opacity: 0 !important }`;
-		document.body.append(style);
-
 		await openPreferences();
 
 		return true;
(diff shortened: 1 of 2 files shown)


logiclrd:
In #2167, there is a comment indicating that the reason the security dialog doesn't show up is that the root element has class `hide-preferences-window`. Empirically, I can confirm that removing this class makes the security dialog show up. This PR is marked as fixing #2167, but won't `hide-preferences-window` still be on the root element??

mquevill:
> In #2167, there is a comment indicating that the reason the security dialog doesn't show up is that the root element has class `hide-preferences-window`. Empirically, I can confirm that removing this class makes the security dialog show up. This PR is marked as fixing #2167, but won't `hide-preferences-window` still be on the root element??

The `hide-preferences-window` class affects all pop-ups (preferences window, sync messages window, etc.) and should be removed after the hidden preferences window is closed. The issue has been that the function `isPreferencesOpen()` would return true even if the "Sync now" dialog was open. By switching to the `aria-label`, it's more specific to the preferences window only. This means that the code will correctly detect when that window closes and the `hide-preferences-window` class can be removed.

vibl:
Same for 2.60.3 here.

PR #2184, opened 2024-05-29
Update narrow and hidden sidebar selectors

mquevill:
Visually, the forced narrow sidebar looks the same as a narrow window (<900 px).
The hidden sidebar does not hide the far left bar, but does hide the conversation list.

[The good news is that the message previews are still there when the window is narrow, but just hidden. So it may be able to be reconstructed, unlike in #1652. I'll leave that as a separate pull request though.]


mquevill:
This is broken by recent backend changes, so I've converted this PR to a draft.

PR #2198, opened 2024-06-28
Fix Conversation Controls and Search Function

### LINKED ISSUE #2185: search shortcut does not work on caprine
Ctrl f does not seem to work in messages, so sadly you cannot search anything through caprine : / i asked for a friend and he has the same issue too

mquevill:
This fixes most of the selected conversation controls as well as the search functionality within a conversation.

The archive and delete functions need to be negative, since there are slightly different menus if a conversation is encrypted or not (usually if there's a separate encrypted conversation), but the end of the menus are the same. (Not sure why this distinction still exists...)

Already encrypted:
image
Can start encrypted chat:
image

Fixes #2185.

### DIFF
--- a/css/browser.css
+++ b/css/browser.css
@@ -102,7 +102,7 @@ body::-webkit-scrollbar {
 }
 
 /* A utility class for temporarily hiding all dropdown menus */
-html.hide-dropdowns [role='menu'].l9j0dhe7.swg4t2nn {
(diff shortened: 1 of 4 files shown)


mquevill:
I've updated the changes from this pull request for the current Messenger interface. In my testing, some of the archive/delete behavior is messed up by the `hide-preferences-window` bug (#2167), but work if the `hide-preferences-window` class is removed.

PR #2228, opened 2024-09-13
Update website

tbertels:
Since https://sindresorhus.com/caprine isn't available anymore, update it to point to the GitHub project
(https://github.com/sindresorhus/caprine).


raylu:
sindresorhus.github.io/caprine redirects to there. it seems like just configuring GH pages to not redirect to sindresorhus.com could be enough to get the old site up and running

PR #2298, opened 2025-03-19
Upgrade Messenger Icons to 2025 version

FixeQyt:
Recently Meta released new icons for Messenger so I'm just upgrading them ;)


tranducminh21208:
Can you update the message sync feature, when I click on the message sync feature it doesn't respond at all

FixeQyt:
> Can you update the message sync feature, when I click on the message sync feature it doesn't respond at all

Sorry, but this pr is only to upgrade the icons
Maybe in the future I will do this...

PR #2353, opened 2025-12-24
Disable hidden preferences functions

mquevill:
Removes all calls to functions that call for the `'hide-preferences-window'` class, which has caused many issues with its volatility. Most of the functions that were using this functionality were not working correctly anyway, so no functionality is lost. These functions should eventually be cleaned up or properly implemented, so I left relevant `TODO` comments.

Should "fix" many outstanding issues, but will go through them separately.

### DIFF
--- a/source/browser.ts
+++ b/source/browser.ts
@@ -424,13 +424,14 @@ async function updateSidebar(): Promise<void> {
 	}
 }
 
-// TODO: Implement this function
 async function updateDoNotDisturb(): Promise<void> {
+	/* TODO: Implement this function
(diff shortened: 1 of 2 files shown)


mquevill:
The actions that run for pull request do not do a full build. You would need to check out the repository, check out the [pull request locally](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally), and do a local build using the basic [dev instructions](https://github.com/sindresorhus/caprine?tab=readme-ov-file#dev).

